PR #12, opened 2021-06-18
add stylua config search argument

sandersantema:
This adds the `-s` argument to stylua which causes stylua to automatically
search for a config file until the root is found and if no config is found it
uses config files in `$XDG_CONFIG_HOME` as a fallback. This is still a bit janky
as stylua doesn't seem to be informed of the root directory of the language
server and therefore seems to search for config files in the current `:pwd` of
vim itself.


jose-elias-alvarez:
Thanks! I'll see how I can improve behavior but this is working well on my end after updating StyLua.

PR #18, opened 2021-06-25
feat: added support for lspconfig & builtin for selene [DRAFT]

folke:
Hi!

First of all, thank you for this amazing plugin.

This PR makes `null-ls` configurable with `lspconfig`.

With this change, autocommands or managing the null-ls client is now taken care of by lspconfig instead.

I haven't removed any code and haven't also change the way it works using `null-ls.setup`.

The lspconfig way is currently accessible by doing:

```lua
local nls = require("null-ls")

nls.setup_lspconfig({
	sources = {
		nls.builtins.formatting.prettierd,
		nls.builtins.formatting.stylua,
		nls.builtins.formatting.eslint_d,
		nls.builtins.diagnostics.shellcheck,
		nls.builtins.diagnostics.markdownlint,
		nls.builtins.diagnostics.selene,
	},
})

local lspconfig = require("lspconfig")
lspconfig["null-ls"].setup({ on_attach = function(client) end })
```

I also added a new builtin for [selene](https://github.com/Kampfkarren/selene), a lua linter.

## Why?

* all exisiting lspconfig functionality can now be used with null-ls
* root_dir detection works, and I made some changes so that the root_dir is used to spawn generators
* when restoring a session with multiple open splits, the different buffers will now correctly be checked for diagnostics
* LspInfo now correctly displays all info for null-ls
* simplifies the null-ls code since the client managing code is no longer needed
* other plugins can now properly inject custom things in null-ls too (I have an unreleased plugin for managing vim stuff with json schemas)

Let me know what you think.

**LspInfo:**
image

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -187,4 +187,46 @@ M.shellcheck = h.make_builtin({
     factory = h.generator_factory,
 })
 
+M.selene = h.make_builtin({
+    method = DIAGNOSTICS,
+    filetypes = { "lua" },
+    generator_opts = {
+        command = "selene",
+        args = { "--display-style", "json", "-" },
+        to_stdin = true,
+        format = "line",
+        check_exit_code = function(code)
+            return code <= 1
+        end,
+        on_output = function(line, params)
+            local function get_pos(byte)
+                return unpack(vim.api.nvim_buf_call(params.bufnr, function()
+                    local lnum = vim.fn.byte2line(byte)
+                    local lbyte = vim.fn.line2byte(lnum)
+                    return { lnum, byte - lbyte + 1 }
+                end))
+            end
+            local ok, diagnostic = pcall(vim.fn.json_decode, line)
+            if not ok then
+                return
+            end
+            local span = diagnostic.primary_label.span
+            local row, col = get_pos(span.start)
+            local end_row, end_col = get_pos(span["end"])
+            local ret = {
+                row = row,
+                col = col,
+                end_row = end_row,
+                end_col = end_col,
+                message = diagnostic.message,
+                code = diagnostic.code,
+                source = "selene",
+                severity = (diagnostic.severity == "Error" and 1) or (diagnostic.severity == "Warning" and 2) or 4,
+            }
+            return ret
+        end,
+    },
+    factory = h.generator_factory,
+})
+
 return M
--- a/lua/null-ls/client.lua
+++ b/lua/null-ls/client.lua
@@ -10,6 +10,7 @@ local api = vim.api
 local schedule = vim.schedule_wrap
 
 local function on_init(client)
+    s.set({ client_id = client.id })
     handlers.setup_client(client)
     s.initialize(client)
 end
@@ -18,10 +19,8 @@ local on_exit = function()
     s.reset()
 end
 
-local start_client = function()
-    s.reset()
-
-    local client_id = lsp.start_client({
+local function get_client_config()
+    return {
         cmd = {
             c.get().nvim_executable,
             "--headless",
@@ -38,8 +37,12 @@ local start_client = function()
         on_attach = c.get().on_attach,
         name = "null-ls",
         flags = { debounce_text_changes = c.get().debounce },
-    })
+    }
+end
 
+local start_client = function()
+    s.reset()
+    local client_id = lsp.start_client(get_client_config())
     -- this completes before the client is initialized
     -- and signals that start_client should not be called again
     s.set({ client_id = client_id })
@@ -69,6 +72,8 @@ local M = {}
 M.start = start_client
 
 M.try_attach = try_attach
+
+M.get_client_config = get_client_config
 
 -- triggered after dynamically registering sources
 M.attach_or_refresh = schedule(function()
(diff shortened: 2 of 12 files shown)


jose-elias-alvarez:
Wow, thanks so much for working on this – I definitely see the advantages, and I'd love to make this the preferred way to set up null-ls and deprecate the existing client functionality wherever possible.

A few questions from playing around with this: 

1. Depending on the user's configuration, it seems that `setup_lspconfig` can cause `get_rtp` to throw an error because it can't yet resolve the plugin path from `vim.o.rtp` (this is the case for my config, at least). Adding `vim.cmd("packadd null-ls.nvim")` to `get_rtp` seems to solve the issue by making sure the plugin is available before attempting to resolve its path. I don't know as much about `rtp` as I should, but what do you think? 
2. null-ls servers automatically shut down if they haven't received a notification from the client within a specified period, to prevent orphans caused by crashes / quitting Neovim with `noautocmd`. One of the current autocommands is responsible for checking that the server is still alive and restarting it if it's not. Is this something nvim-lspconfig can handle?
3. What do you think about renaming `setup_lspconfig` to `config` (or something similar)? I want to clarify the distinction between the methods, and I think it'll look a bit cleaner if we move entirely to nvim-lspconfig in the future.

folke:
HI!

1. I just pushed a change that gets the rtp from the source file instead. This is more reliable than using the runtime path
2. that autocmd is probably still needed then
3. done

When running null-ls through lsp config, multiple clients may exists (which is desirable in monorepos for example).

So I assume the state stuff needs to be indexed on client_id to make this work properly? Although I believe at least some of the code in state is also no longer needed.

folke:
Would also make sense to set configuration as you're doing right now in the setup method, as part of lsp server settings.

So:

```lua
require("null-ls").setup({sources = ...})

-- new way
lspconfig["null-ls"].setup(settings = {sources = ...})
```

`on_new_config` is then called with the resolved settings and root_dir for the new server that should start.

You can also reference the client settings from anywhere on the `client.settings` object.

If `null-ls` would get added to the lspconfig repo, then no additional requires for `null-ls` would be needed and it could be setup as a regular lsp server

jose-elias-alvarez:
Wow, interesting idea to solve the `rtp` issue. If we can do this, I imagine we could move it out of state entirely and just put it directly into the config. 

For the autocommand: I see that nvim-lspconfig has an `:LspRestart` method. Could we change the autocommand to simply check if the server is still active and restart it if it's not? We could then extend the server's default timeout and have the server refresh itself when it receives anything from the client (not just a specific notification). This + the `rtp` change would let us simplify the state down to cache management, which would work well with `client_id`-based indexing (which I agree is necessary if one Neovim instance could start multiple servers). 

I also like the idea of moving configuration into `lspconfig["null-ls"].setup`. Could we use a small Vimscript file to set up the lspconfig integration regardless of whether the user has called `require("null-ls").setup()`? We could then switch over to lspconfig entirely but keep the original `setup` method as a wrapper to avoid breaking existing configs (technically, null-ls doesn't currently depend on nvim-lspconfig, but I imagine nearly 100% of users are already using it). 

Let me know if you think these sound feasible and I can work on ripping out the unnecessary bits.

jose-elias-alvarez (approved):

jose-elias-alvarez:
No worries! The change to `setup` is exactly what I wanted. I've just merged the PR into a new branch on this repo and will finalize it there.

folke:
@jose-elias-alvarez great!

folke:
Once you push this to main, we can see to make a PR to lspconfig to include it there. I already checked and they're fine with the fact that we should then do require calls to `null-ls` code.

Another change then would be to implement `on_new_config` to setup `null-ls`.

On a sidenote, what part of null-ls actually requires to spawn nvim as an RCP server?

Is there anything we can do on the lspconfig side or neovim side to prevent the need for that?

jose-elias-alvarez:
Sounds good! Thank you for all of the help and support.

To answer your question: As far as I could tell, there's no way to get `lsp.start_client` to register and initialize a new client unless it spawns a server that [returns a valid RPC response](https://github.com/neovim/neovim/blob/af263711a368bcbdea66d086537316495ed64576/runtime/lua/vim/lsp.lua#L828). After that, you have access to the client in `on_init`, so you can override its methods to avoid server communication. Essentially, the server is only necessary to initialize the client, and after that it should (ideally) do nothing until Neovim tells it to shut down (though even that could be avoided by overriding `client.stop`). 

If there was a way to access the RPC request callback and directly register a client, that would open the door for anyone to write full-on in-memory language servers. The role of null-ls would then be to abstract away the process of properly structuring responses while providing a platform for writing and registering sources. From my perspective, that would be fantastic, and on a technical level it doesn't seem impossible, but I wonder how open other Neovim maintainers would be to the idea.

folke:
ok, thanks! I'll see if we can make a change that would make this possible.

folke:
What if we would allow a custom `rpc.request` to be passed as config?
Then you wouldn't even need to override client.request etc right?

folke:
I have an idea :)

We can actually add a `before_init` handler to the client and there we can use `debuginfo.setlocal` to set the `rpc` value to a new variable that has a custom `request` method.

I'll run some tests to see how that could work.

bpinto:
Sorry for commenting on an already merged PR, but just in case someone reads this PR in the future,  `setup_lspconfig` has been renamed to `config`, so the code is actually: `nsl.config({ sources = {....} })`

PR #25, opened 2021-07-05
Support textDocument/rangeFormatting

### LINKED ISSUE #22: [Feature Request] support for `textDocument/rangeFormatting`
I tried to implement it myself. But got stuck because the expected type of `args` is `table`:

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/40d6cb2f6e0609e6341ac89f448e7d43f35506a9/lua/null-ls/helpers.lua#L84

But for rangeFormatting, the formatter needs to know about the range. e.g. for `prettier`, we need to pass [two args](https://prettier.io/docs/en/options.html#range) `--range-start` and `--range-end`.

If you point me to how you want to handle this, I can create a PR with the changes.

MunifTanjim:
Adds support for `textDocument/rangeFormatting`.

Example plugin that uses this feature:

- https://github.com/MunifTanjim/prettier.nvim

---

resolves #22

### DIFF
--- a/lua/null-ls/builtins/formatting.lua
+++ b/lua/null-ls/builtins/formatting.lua
@@ -2,6 +2,7 @@ local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 
 local FORMATTING = methods.internal.FORMATTING
+local RANGE_FORMATTING = methods.internal.RANGE_FORMATTING
 
 local M = {}
 
@@ -54,8 +55,48 @@ M.isort = h.make_builtin({
     factory = h.formatter_factory,
 })
 
+local get_prettier_generator_args = function(common_args)
+    return function(params)
+        local args = vim.deepcopy(common_args)
+
+        if params.method == FORMATTING then
+            return args
+        end
+
+        local range = params.range
+        local start_row, start_col = range.row, range.col
+        local end_row, end_col = range.end_row, range.end_col
+
+        local range_start = 0
+        local range_end = 0
+
+        for i, line in pairs(params.content) do
+            if i <= end_row then
+                if i == start_row then
+                    range_start = range_start + string.len(string.sub(line, 1, start_col))
+                    if i == end_row then
+                        range_end = range_end + string.len(string.sub(line, 1, end_col))
+                    end
+                elseif i == end_row then
+                    range_end = range_end + string.len(line, 1, end_col)
+                else
+                    range_start = range_start + string.len(line)
+                    range_end = range_end + string.len(line)
+                end
+            end
+        end
+
+        table.insert(args, "--range-start")
+        table.insert(args, range_start)
+        table.insert(args, "--range-end")
+        table.insert(args, range_end)
+
+        return args
+    end
+end
+
 M.prettier = h.make_builtin({
-    method = FORMATTING,
+    method = { FORMATTING, RANGE_FORMATTING },
     filetypes = {
         "javascript",
         "javascriptreact",
@@ -69,7 +110,7 @@ M.prettier = h.make_builtin({
     },
     generator_opts = {
         command = "prettier",
-        args = { "--stdin-filepath", "$FILENAME" },
+        args = get_prettier_generator_args({ "--stdin-filepath", "$FILENAME" }),
         to_stdin = true,
     },
     factory = h.formatter_factory,
@@ -90,14 +131,14 @@ M.prettierd = h.make_builtin({
     },
     generator_opts = {
         command = "prettierd",
-        args = { "$FILENAME" },
+        args = get_prettier_generator_args({ "$FILENAME" }),
         to_stdin = true,
     },
     factory = h.formatter_factory,
 })
 
 M.prettier_d_slim = h.make_builtin({
-    method = FORMATTING,
+    method = { FORMATTING, RANGE_FORMATTING },
     filetypes = {
         "javascript",
         "javascriptreact",
@@ -106,7 +147,7 @@ M.prettier_d_slim = h.make_builtin({
     },
     generator_opts = {
         command = "prettier_d_slim",
-        args = { "--stdin", "--stdin-filepath", "$FILENAME" },
+        args = get_prettier_generator_args({ "--stdin", "--stdin-filepath", "$FILENAME" }),
         to_stdin = true,
     },
     factory = h.formatter_factory,
--- a/lua/null-ls/builtins/formatting.lua
+++ b/lua/null-ls/builtins/formatting.lua
@@ -63,28 +63,15 @@ local get_prettier_generator_args = function(common_args)
             return args
         end
 
-        local range = params.range
-        local start_row, start_col = range.row, range.col
+        local content, range = params.content, params.range
+
+        local row, col = range.row, range.col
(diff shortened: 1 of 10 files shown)


lua/null-ls/utils.lua
```diff
@@ -60,6 +61,7 @@ M.make_params = function(original_params, method)
         method = method,
         row = pos[1],
         col = pos[2],
+        range = range,
```
jose-elias-alvarez, line 64: One thing about LSP ranges is that they're annoying to work with in Lua, mostly because `end` is a reserved keyword and has to be wrapped in brackets and quotes (`["end"]`) to reference the key. Could we convert the range to the following format, which is the same one that sources return to null-ls?

```lua
local null_ls_range = {
    row,
    col,
    end_row,
    end_col,
}
```

There is already a [method](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/eed60b350100cd4897a60b6de2c984a7b1f979c1/lua/null-ls/formatting.lua#L41) to do the opposite conversion, so maybe we could consolidate both of those here.

jose-elias-alvarez (commented):
Looks great! This is everything we need to support the method at the source level. In addition to modifying the provided LSP range, I'd like to add a couple of more features to make it easier to write built-ins for the new method.

1. Could we make it possible for one source to register its availability for multiple methods?
2. Could we update [formatter_factory](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/eed60b350100cd4897a60b6de2c984a7b1f979c1/lua/null-ls/helpers.lua#L185) to check the range in the params instead of always updating the whole document? 

It would also be great if we could update the [Prettier built-in](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/eed60b350100cd4897a60b6de2c984a7b1f979c1/lua/null-ls/builtins/formatting.lua#L57) to support the method and provide an example for other integrations.

MunifTanjim:
> 2. Could we update [formatter_factory](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/eed60b350100cd4897a60b6de2c984a7b1f979c1/lua/null-ls/helpers.lua#L185) to check the range in the params instead of always updating the whole document?

I'm not sure about other formatters, but with range formatting, Prettier still takes in the whole document and outputs the whole document (but only the part inside `--range-start` and `--range-end` is formatted, the rest of the document is left unchanged). It needs the whole document because otherwise it would lose context (e.g. indentation level etc.). So we need to update the whole document with the output from Prettier.

I guess it would be the same with other formatters? So maybe we can skip this change right now?

jose-elias-alvarez:
I honestly have no idea about other formatters, but if that's how Prettier handles it, that sounds fine to me.

lua/null-ls/utils.lua
MunifTanjim: Can you please check the 3rd commit to see if that reflects what you wanted?

MunifTanjim:
> 1. Could we make it possible for one source to register its availability for multiple methods?

Done.

> It would also be great if we could update the [Prettier built-in](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/eed60b350100cd4897a60b6de2c984a7b1f979c1/lua/null-ls/builtins/formatting.lua#L57) to support the method and provide an example for other integrations.

Done.

lua/null-ls/builtins/formatting.lua
```diff
@@ -54,8 +55,48 @@ M.isort = h.make_builtin({
...
+        table.insert(args, "--range-start")
+        table.insert(args, range_start)
+        table.insert(args, "--range-end")
+        table.insert(args, range_end)
+
+        return args
+    end
+end
+
 M.prettier = h.make_builtin({
```
jose-elias-alvarez, line 98: It looks like Prettier doesn't support range formatting for some filetypes, including [Markdown](https://github.com/prettier/prettier/issues/5008). There's obviously nothing we can do about this here, but I think it's worth documenting in [BUILTINS.md](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md).

lua/null-ls/builtins/formatting.lua
```diff
@@ -90,14 +131,14 @@ M.prettierd = h.make_builtin({
     },
     generator_opts = {
         command = "prettierd",
-        args = { "$FILENAME" },
+        args = get_prettier_generator_args({ "$FILENAME" }),
```
jose-elias-alvarez, line 134: Are we sure that `prettierd` and `prettier_d_slim` support range formatting (and that it actually works)? When working with [eslint_d](https://github.com/mantoni/eslint_d.js/), I ran into situations where it didn't support the same CLI arguments as ESLint (or where they didn't work), so I just want to confirm.

lua/null-ls/builtins/formatting.lua
```diff
@@ -54,8 +55,48 @@ M.isort = h.make_builtin({
...
+        end
+
+        local range = params.range
+        local start_row, start_col = range.row, range.col
+        local end_row, end_col = range.end_row, range.end_col
+
+        local range_start = 0
+        local range_end = 0
+
+        for i, line in pairs(params.content) do
```
jose-elias-alvarez, line 73: This is behaving a little strangely for me, and I'm also not sure that it'll work for multibyte characters. I recently dealt with a similar problem, and so far the best solution I've found to convert document positions to character positions is something like this:

```lua
local range = params.range

local view = vim.fn.winsaveview()
vim.api.nvim_win_set_cursor(0, { range.row, range.col })
local range_start = vim.fn.wordcount().cursor_chars
vim.api.nvim_win_set_cursor(0, { range.end_row, range.end_col })
local range_end = vim.fn.wordcount().cursor_chars
vim.fn.winrestview(view)

table.insert(args, "--range-start")
table.insert(args, range_start)
table.insert(args, "--range-end")
table.insert(args, range_end)
```

This improves the behavior on my end, so let me know what you think.

MunifTanjim: Yes, you're right. It was buggy with multibyte chars.

However I think moving the cursor around and relying on `winsaveview` will cause issue with folds (as per `:h winsaveview`) and have other side-effects.

I've updated the character counting logic that results in same output as your suggested method, but doesn't involve `winsaveview` or moving the cursor.

Let me know if that works for you 😃

lua/null-ls/builtins/formatting.lua
MunifTanjim: Turns out `prettier_d_slim` does, `prettierd` doesn't. Updated accordingly.

lua/null-ls/builtins/formatting.lua
MunifTanjim: Done.

lua/null-ls/builtins/formatting.lua
jose-elias-alvarez: Works great! Much better than my solution. One small optimization: `params.content` is a table list containing the buffer's content, so can we refactor this to get the lines from there instead of `vim.fn.getline`?

MunifTanjim: Nice. Done!

jose-elias-alvarez (approved):

jose-elias-alvarez:
Thank you for your work on this!

PR #30, opened 2021-07-12
feat: builtin vale

carrascomj:
### Description

[vale](https://docs.errata.ai/vale/about) is an open-source syntax-aware linter for prose built with speed and extensibility in mind (repo [here](https://github.com/errata-ai/vale)).

It supports adding syntax files (from, for instance, Linode, Gitlab or write-good) and per project syntax styles and it is quite fast.

### Implementation
* Added Diagnostics built-in `vale`.
* Parsed as JSON (all the buffer at once).
* The user is expected to have a `.vale.ini` or `_vale.ini` somewhere at `$HOME` or at the current working directory (you can take one from https://docs.errata.ai/vale/about#open-source-configurations, put the `.vale.init` at home and modify it to point to the `StylesPath`).
* Added some documentation.

TODO:
* [ ] Some tests.
* [x] Will this be accepted? I don't know lua so it would be nice to know before I try to start learning how to do functional testing here.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -88,6 +88,40 @@ M.markdownlint = h.make_builtin({
     factory = h.generator_factory,
 })
 
+M.vale = h.make_builtin({
+    method = DIAGNOSTICS,
+    filetypes = { "markdown", "tex" },
+    generator_opts = {
+        command = "vale",
+        format = "json",
+        args = { "--no-exit", "--output=JSON", "$FILENAME" },
+        on_output = function(params, done)
+            local diagnostics = {}
+            if not params.output or params.output == "" then
+                return done()
+            end
+            for _, diagnostic in ipairs(params.output[params.bufname]) do
+                if diagnostic then
+                    local severity = 1
+                    if diagnostic.Severity == "warning" then
+                        severity = 2
+                    end
+                    table.insert(diagnostics, {
+                        row = diagnostic.Line,
+                        col = diagnostic.Span[1],
+                        end_col = diagnostic.Span[2],
+                        source = diagnostic.Check .. " (vale)",
+                        message = diagnostic.Message,
+                        severity = severity,
+                    })
+                end
+            end
+            return diagnostics
+        end,
+    },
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics.lua
```diff
@@ -88,6 +88,40 @@ M.markdownlint = h.make_builtin({
...
+            end
+            for _, diagnostic in ipairs(params.output[params.bufname]) do
+                if diagnostic then
+                    local severity = 1
+                    if diagnostic.Severity == "warning" then
+                        severity = 2
+                    end
+                    table.insert(diagnostics, {
+                        row = diagnostic.Line,
+                        col = diagnostic.Span[1],
```
jose-elias-alvarez, line 111: On my end, I need to subtract 1 from `col` in order for diagnostics to line up correctly. Is that the same for you?

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -88,6 +88,40 @@ M.markdownlint = h.make_builtin({
...
+                        end_col = diagnostic.Span[2],
+                        source = diagnostic.Check .. " (vale)",
```
jose-elias-alvarez, line 113: I think that a more standard approach would be to leave `vale` as the source, then assign `diagnostic.Check` to the key `code`. This makes the information available to plugins like [trouble.nvim](https://github.com/folke/trouble.nvim).

jose-elias-alvarez:
Thank you for the PR! No need to worry about tests for now, since the other built-ins don't have them.

lua/null-ls/builtins/diagnostics.lua
carrascomj: This looks alright to me with col in markdown
image
 
Or in latex
image

Maybe it is something within my config `init.lua`, if someone can confirm, we can leave it at `col - 1`.

carrascomj: OK, I confused col with row for a moment, is `end_col` also wrong for you? I cannot see the col with my actual config, I am working on that.

jose-elias-alvarez:
Thanks again for your work on this!

PR #32, opened 2021-07-16
make `null-ls` in-process and manage it with `lspconfig`

folke:
Hi!

I went ahead and did all the remaining changes to make `null-ls` fully **in-process** and make it manageable with `lspconfig`.

A lot of the code could be removed and it is no longer needed to wrap any **client** methods. All the magic now happens inside the `rpc.lua` file.

This PR is a rebase of the `lspconfig` branch, so it's ready to be merged in **main** once you reviewed it.

I'm aware this is a pretty huge PR, so feel free to ask any questions on any of the changes.

There's two ways to use `null-ls`:

```lua
-- method 1 (preferred)
require("null-ls").config(...)
require("lspconfig")["null-ls"].setup({on_attach = ...})

-- method 2 (does the above and automatically calls `lspconfig.setup`
require("null-ls").setup(...)
```

## Notes

* I removed the [handle_all_factory](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/56d4b76203d2b442353e25c247f49fa5ca70f42e/lua/null-ls/handlers.lua#L40) and replaced it with a `debounce` handler that batches results as before, without needing to wrap `buf_request`. It wraps the `textDocument/codeAction` handlers instead. Slow servers will possible show code actions after the first batch, but that's better imo so a slow server can't hold back on doing actions. Will see to upstream this to Neovim for code actions.
* added sorting to code actions to make sure they always have the same number to execute them
* added **selene** to builtins
* changed `save_win_data` to make it work with the default handlers. (they pass bufnr now too)

## Todo

* [x] all tests return OK


jose-elias-alvarez:
Amazing! Thank you for all your hard work on this! 

All of the changes make sense to me – I especially like how code actions are handled – and the PR is working almost perfectly on my end. I'm happy to take over from here and update the documentation so I can coordinate merging into `main` with any necessary changes to nvim-lsp-ts-utils. 

A couple of things I noticed while testing this out locally:

1. When a diagnostics source is dynamically registered (e.g. via nvim-lsp-ts-utils), diagnostics aren't shown until the buffer's content is modified / updated. The current implementation [triggers an autocommand](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/56d4b76203d2b442353e25c247f49fa5ca70f42e/lua/null-ls/config.lua#L101) on registration to make sure diagnostics are updated, but do you think there's a more elegant way to handle this? 
2. I noticed a delay when exiting Neovim, and I think I traced it to the fact that [dispatch.on_exit](https://github.com/neovim/neovim/blob/682247b52e76180e4c49b794c6cc2548bfd32d4c/runtime/lua/vim/lsp.lua#L728) is never called. I fixed the delay by passing the `dispatch` object to `handler` and calling `on_exit(0)` on an `exit` notification, but does that make sense to you?

folke:
I just fixed both issues.

1. fixed this in `lspconfig.lua`. Will now works as expected when adding a new source with a new filetype (doing `try_add` ) and also when adding a new source with an existing filetype (we send a `didOpen`).
2. you're right. Hadn't implemented the `notify('exit')`. It now correctly calls `dispatchers.on_exit` in that case

One thing that's still open, is that after formatting, and then undoing, the cursor jumps to the bottom.

Was that issue already there before?

Either way, I'll see if I can fix this too.

folke:
I think I just fixed that format/undo/jump to end issue as well

jose-elias-alvarez:
Great! I had noticed the undo issue before and planned on investigating, so thanks for taking care of it.

I like the idea of sending a `didOpen` notification, but I think there's an issue with the params we're sending. With the current implementation, `tsserver` doesn't work after the notification is sent, and I see this message in my logs:

```
[ ERROR ] 2021-07-17T18:38:10+0900 ] /usr/local/share/nvim/runtime/lua/vim/lsp/handlers.lua:413 ]	"Notification handler 'textDocument/didOpen' failed with message: Cannot read property 'length' of undefined"
```

Changing the notification's parameters to include `text` fixes the issue, and I wonder if it's a good idea to make sure the params match [what Neovim sends](https://github.com/neovim/neovim/blob/d8f5f4d09078e7d52d1abd13418133da10b2cc27/runtime/lua/vim/lsp.lua#L443).

folke:
Right :)

Maybe it's better to send a custom null-ls notification?

jose-elias-alvarez:
Empty `didChange` is what [the current implementation does](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/bd835497b41874ca78d0d0d0808d8a834dc03167/lua/null-ls/client.lua#L78), so I think it's the easiest solution!

folke:
fixed :)

folke:
One of the e2e tests keeps failing right now. The one where we shouldn't get the `cached` version of code actions after a text edit. I'm pretty sure it worked last night 😅

It seems as if the lsp client is not sending a `didChange` for some reason. I'm further looking into this, but if you have any ideas, let me know

folke:
But when using `null-ls`, everything seems to work as expected

folke:
image

That first failure is to be expected, since that moved to the handler, but that second one should work.

jose-elias-alvarez:
Wow, no idea about this failure. I've been hammering away at it, but the only thing I've discovered is that commenting out the first code action `describe` block in the e2e test file solves the issue (and that it is indeed related to whether the notification is sent or not). You're right that it seems to work fine, though, so I wonder what's going on. 

Another question: I have some code in nvim-lsp-ts-utils that iterates over responses to extract actions:

```
for _, response in ipairs(responses) do
    if response.result then
        for _, action in pairs(response.result) do
            if should_handle_action(action) then
                parse_action(action)
            end
        end
    end
end

```

With the new implementation, I have to check whether `response.result == vim.NIL` (otherwise it'll throw an error saying I tried to iterate over userdata). Is there any harm to calling `send` with a `nil` result? I'm not sure whether this is an issue with my code, specifically, but I wouldn't want to break other integrations.

folke:
Yep, you're right. I took that over from the old code. There it was needed becuase of the actual RPC messages. Since it's all in-process now, sending `nil` is good enough.

Still looking into that failing test.

I did notice that the `is_registered` could never work. So I fixed that as well. When registering the same source twice, it actually added multiple generators and after a while things started failing due to using the same plenary spawn twice or something like that.

I also added config files for `selene` (added a builtin to nullls for that too). It's a pretty great linter for lua.

There's also `.nvim.setttings.json`. Just ignore that for now. It's used by an upcoming plugin to manage workspace settings etc like in vscode. This will correctly setup sumneko for `null-ls` by adding the libraries for `plenary` and `lspconfig`.

You'll likely get tons of lua warnings rights now, especially inside the tests. The workspace stuff fixes all that.

I'll further check on that failure. I think it has something to do with change tracking.

jose-elias-alvarez:
Thank you for the fixes! I wanted to give `selene` a shot, so this is a good opportunity. 

For the test: I realized that null-ls isn't receiving `didOpen` notifications for anything but the first document Neovim opens, which means it's only subscribed to changes for that first document (this is also why I think the diagnostic test is failing). I'm not sure how to fix it (I imagine it's related to lspconfig?) but I think it's both the cause of the failure as well as an issue on its own.

folke:
In my case I do get the `didOpen` notifications. Are you sure about that?

jose-elias-alvarez:
I think so. I added `print(method)` to `notify` in `rpc.lua`, and when I open a null-ls-enabled file, I see `textDocument/didOpen` for the first file, but not subsequent files. Likewise, when I run the e2e tests, I see `textDocument/didOpen` right after the tests start, but never again. Is that not the case for you? If so, that's even stranger.

folke:
image

I get didOpen for every new file.

Opening an existing file won't send a didOpen, maybe that's what you have?

folke:
In the 2e2 tests, we're opening the same file over and over, but there's a `budo! bdelete!`, so you will only see a `didOpen` twice, for each of the code actions describve block

folke:
The failing test is definitely because of the change tracking. So it all works, it's just the test that is not working properly, becuase vim batches the edits and doesn't send the update.

I'm looking into disabling change tracking for the tests

folke:
doesn't fix it yet unfortunately, so probably still something else...

I think I know why you don't see the `didOpen` message. It's because of the `FileType` autocmd, that silences any output.

You can either do:

```lua
-- defer to main and print it there
vim.defer_fn(function() print("foo")end, 0)

-- or set this option to unsilence those autocmds (remove the `F` flag)
vim.o.shortmess = "IToOlxfitn"
```

I have to go here, but will see if I can further look into that failing test later :)

folke:
I just pushed a change that adds 'unsilent' to the autocmd. This probably also fixes the reason you don;t see that `didOpen` message :)

folke:
fyi: with the `gh` command line tool, you can easily switch to this pr and also push commits in it.

jose-elias-alvarez:
That did it! That was driving me crazy, so thank you! I'm also out of time for today, but as long as the actual behavior is fine, I'm happy, so I'll continue working on the PR the next time I have a chance.

jose-elias-alvarez:
I had no idea `gh` let you push to PRs – fantastic. That fix takes care of the other issue I was seeing where diagnostics weren't loaded after opening a file. 

I'll stress test the PR at work this week and hopefully finalize it, too. Thanks again for all your work on this!

folke:
I only discovered that you could push commits to a PR recently myself. Super useful.

I can see how things could fail without your fix. Makes sense!

folke:
fyi. I made an upstream PR to aggregate code-actions https://github.com/neovim/neovim/pull/15121

jose-elias-alvarez:
Great! I think it's a big improvement on the default behavior, which doesn't make a lot of sense (and I think the reason it hasn't been a bigger issue is because few people run more than one language server that provides code actions).

jose-elias-alvarez:
@folke Since your PR was merged, is it safe to take out the corresponding code here? Other than that, I think this is ready to merge – I used it pretty heavily at work the past couple of days and ran into exactly zero issues.

folke:
@jose-elias-alvarez I'm not entirely sure what you mean :)

With **here**, do you mean the branch you got with `gh pr checkout`? All your commits are on this PR too.

So yep, you should be able to safely merge this PR. (not sure what you mean with `take out the corresponding code here`)

jose-elias-alvarez:
Sorry, I was specifically referring to your [code action aggregation PR](https://github.com/neovim/neovim/pull/15121) (and I realize now that my message was spectacularly unclear).

folke:
Oh right 😅

Yep, would make sense to remove the handler again in null-ls. Unless you want to keep supporting Neovim 0.5 for now. Your call :)

jason0x43:
Just randomly chiming in -- I would vote for supporting Neovim 0.5 if it's not too messy. We just got this nice feature-filled stable release. 🙂

jose-elias-alvarez:
Sure, I can't guarantee that we'll remain fully compatible, but in this case putting in a version check is easy.

jose-elias-alvarez:
Just merged, thanks again for all the hard work on this!

folke:
Thank you for merging!!

PR #34, opened 2021-07-18
feat: add flake8 diagnostics

yujinyuz:
[asciicast](https://asciinema.org/a/ASx3KjSZVoKoOBMMjttCk0XJo)

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -366,7 +366,6 @@ M.hadolint = h.make_builtin({
     factory = h.generator_factory,
 })
 
-
 M.flake8 = h.make_builtin({
     method = DIAGNOSTICS,
     filetypes = { "python" },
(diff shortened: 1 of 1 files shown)


yujinyuz:
I was checking other diagnostics and I'm not quite sure about the purpose of what this was doing

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/ec58d64d49fcea19447d1e17ce6be33908eb219d/lua/null-ls/builtins/diagnostics.lua#L25-L33

Doesn't it already return the row and column?

```
"was stolen" may be passive voice on line 4 at column 11
```

jose-elias-alvarez:
Looks good! I don't use Python much, but the first element is an error code, right? If so, we should probably extract it from the string and insert it as `diagnostic.code`. 

No idea about the `write-good` thing (I wrote those mostly as proofs-of-concept, though they work on my end) but if you can find a way to improve an existing built-in, feel free to put in a PR for that, too.

yujinyuz:
@jose-elias-alvarez Here's what it returns

```
users/models.py:10:24: E202 whitespace before ')'
```

Only `10:24 and E202 whitespace before ')'` is extracted.

I'm not sure what you mean by inserting it to `diagnostic.code`.

Do you have samples I can use for reference?

jose-elias-alvarez:
If you look at how [ESLint](https://eslint.org/docs/user-guide/formatters/#table) formats errors, I think the `Rule ID` column of the table corresponds to `E202` in your example. [Here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/56d4b76203d2b442353e25c247f49fa5ca70f42e/lua/null-ls/builtins/diagnostics.lua#L248), you can see we insert the rule ID as `diagnostic.code` (the example is a bit different because ESLint outputs to JSON, but the idea is the same). 

For consistency, I think we should extract `E202` from the message and assign it to `code`, which will make it display nicely with integrations like [trouble.nvim](https://github.com/folke/trouble.nvim).

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -302,4 +302,47 @@ M.hadolint = h.make_builtin({
...
+        command = "flake8",
+        to_stdin = true,
+        to_stderr = true,
+        args = { "--stdin-display-name", "$FILENAME", "-" },
+        format = "line",
+        check_exit_code = function(code)
+            return code == 0 or code == 255
+        end,
+        on_output = function(line, params)
+            local pos = vim.split(string.match(line, "%d+:%d+:.+"), ":")
```
stephan-hesselmann-by, line 319: You can just do
```
local row, col, msg = line:match(":(%d+):(%d+): (.*)")
```

yujinyuz:
It works great for me

Screen Shot 2021-07-20 at 13 04 52

PR #36, opened 2021-07-20
Add shellcheck code to diagnostics

mks1508:
A small change is to show shellcheck's problematic code and diagnostic message together.
```
echo $HOME     ■ [SC2086] Double quote to prevent globbing and word splitting.
```
It's realy helpful and quicker to get more information about the problem from shellcheck's wiki
https://github.com/koalaman/shellcheck/wiki/SC2086

It's also easier to ignore some specific messages
```
# shellcheck disable=SC2086
echo $HOME
# the problem is now ignored
```

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -206,7 +206,7 @@ M.shellcheck = h.make_builtin({
                     end_row = diagnostic.endLine,
                     end_col = diagnostic.endColumn == diagnostic.column and diagnostic.endColumn
                         or diagnostic.endColumn - 1,
-                    message = diagnostic.message,
+                    message = "[SC" .. diagnostic.code .. "] " .. diagnostic.message,
                     source = "shellcheck",
                     severity = diagnostic.level == "error" and 1
                         or diagnostic.level == "warning" and 2


jose-elias-alvarez:
Thank you! However, for consistency, we should return the code as part of the inserted diagnostic table and not as part of the message. This will make the status code show up properly with integrations like [trouble.nvim](https://github.com/folke/trouble.nvim), or when the user modifies their diagnostics handler, as in [this example](https://github.com/neovim/nvim-lspconfig/wiki/UI-customization#show-source-in-diagnostics). 

I'm also considering adding a config option to null-ls itself to automatically include the source and / or code in the diagnostic message, so separating the code would be helpful there, too.

mks1508:
Thank for your explanation.
I really waiting for that config option. Writing shellscript is my main day job and generally I pay more attention to problematic code than its short description.
I haven't figured out the conflict that people can face if we combine problematic code in message.
In the other hand, the builtin flake8 source also had problematic code in its message
Eg:
```
import     ■ E999 SyntaxError: invalid syntax
```
message: E999 SyntaxError: invalid syntax
code: E999

It's something like:
```lua
return {
  ...
  code = diagnostic.code,
  message = diagnostic.code .. diagnostic.message
  ...
}
```

jose-elias-alvarez:
In the flake8 built-in, we ended up [splitting the message and the code](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/0a09a95bb4b45746dfc3731b82e314d8b3064bec/lua/null-ls/builtins/diagnostics.lua#L396). I think it's a matter of consistency – actual LSP sources don't include error codes or sources in the message, so we want to do the same. 

I'll have a little time to work on the configuration option tomorrow. I plan on implementing it as an option for built-ins, meaning you'll be able to do something like this:

```lua
require("null-ls").builtins.shellcheck.with({ show_code = true })
```

PR #37, opened 2021-07-20
enhance: show vale suggestions as style LSP diagnostics

carrascomj:
### Description

I forgot `vale`'s **suggestion** severity. This is a tiny PR that simply adds them.

### Implementation

`vale`'s **suggestion** severities are turned into LSP **hint (4)** severities (errors are 1 and warnings as 2, as before).


jose-elias-alvarez:
Thank you!

Issue #33, opened 2021-07-17
Linter name in diagnostics


stephan-hesselmann-by:
It would be cool to have the option to prefix diagnostics with the linter name. That way diagnostics from multiple linting sources can be distinguished from each other. Efm added a similar feature recently: https://github.com/mattn/efm-langserver/pull/148

gegoune:
It already exists in null-ls, you need to update your diagnostic handler on nvim side. null-ls provides `source` field, have a look here https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics.lua

stephan-hesselmann-by:
Thanks I wasn't aware of that. Could you give me a hint on what I have to do on neovim side to override the handler?

gegoune:
https://github.com/neovim/nvim-lspconfig/wiki/UI-customization#show-source-in-diagnostics

stephan-hesselmann-by:
Perfect thanks a lot :)

PR #39, opened 2021-07-21
feat: add customizable diagnostics format

jose-elias-alvarez:
Several users have requested the ability to show more information in null-ls diagnostics, e.g. #33 (source name) and #36 (code). It's already possible to customize how diagnostics are displayed by [overriding the default handler](https://github.com/neovim/nvim-lspconfig/wiki/UI-customization#show-source-in-diagnostics), and strictly speaking, I think that's the right solution, but it's also pretty hard to customize (beyond copy-pasting) if you're not familiar with Lua / LSP structure. For example, users may want to show codes / source names for certain sources only, and that's 100% possible to do by modifying the handler, but it's also not trivial. 

This tentative PR adds a simple tmux-like template format for diagnostics, which can be set globally or via a built-in's `with` method (the latter overrides the former). I've implemented 3 options, `#{c}` (code), `#{m}` (message), and `#{s}` (source).

For example, using the following settings for shellcheck (which is also fixed in this PR to include the code): 
```lua
require("null-ls").builtins.diagnostics.shellcheck.with({ diagnostics_format = "[#{c}] #{m} (#{s})" })
``` 
Results in the following:
Screen Shot 2021-07-21 at 12 47 45 PM

Alternatives:

- Do nothing and continue to encourage users to override the handler. Not friendly, but easy and strictly "correct". 
- Add hardcoded config options for `show_code` and `show_source` and append / prepend them to the message. If users want further customization, encourage them to override the handler. This would make config even easier and is (maybe?) slightly better in terms of performance but not that different in terms of technical complexity. 
- Add both a customizable format and hardcoded config options, which internally will just set the format to a predefined option. The best of both worlds?
- ???

I'm not 100% sure that this is necessary, so I'd appreciate feedback from anyone who happens to see this!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,59 @@
+name: Bug Report
+description: Report a problem in null-ls
+labels: [bug]
+body:
+
+  - type: checkboxes
+    id: faq-prerequisite
+    attributes:
+      label: FAQ
+      options:
+        - label: I have checked [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.
+          required: true
+
+  - type: checkboxes
+    id: issue-prerequisite
+    attributes:
+      label: Issues
+      options:
+        - label: I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no issues already open with the same problem.
+          required: true
+
+  - type: input
+    attributes:
+      label: "Neovim Version"
+      description: "`nvim --version`:"
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: "Steps to reproduce"
+
+  - type: textarea
+    attributes:
+      label: "Expected behavior"
+
+  - type: textarea
+    attributes:
+      label: "Actual behavior"
+      description: "A description of the behavior you expected. May optionally include logs, images, or videos."
+
+  - type: dropdown
+    id: help
+    attributes:
+      label: "Help"
+      description: "Would you be able to resolve this issue by submitting a pull request?"
+      options:
+        - "Yes"
+        - "Yes, but I don't know how to start. I would need guidance"
+        - "No"
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: "Implementation help"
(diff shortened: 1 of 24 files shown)


gegoune:
I am not strictly against that change but I am leaning toward option 1, do nothing, as there is already good system to change that behaviour right in nvim. I don't think any other either lsps or generic purpose runners do such thing.

jose-elias-alvarez:
I'm also leaning towards doing nothing, since as you say Neovim already provides a very good solution. If we could help users understand how to override the handler to get the results they want, that would be idea. 

On the other side of the argument, I think diagnostic-languageserver actually has this as an option (maybe as a default?), as in [this screenshot](https://user-images.githubusercontent.com/5492542/54487533-15590b80-48d2-11e9-8cba-7e58c0edcf6f.png), so there's some precedent, and reducing the need to copy-paste was one of my primary motivations to make the plugin in the first place. I think I'll sit on this for a couple of days and give it some more thought.

gegoune:
Ah, did not realise they provided it, used that ages ago, but with coc rather than built in lsp.

mks1508:
diagnostic_format seem more flexible and easy to configure depend on people's flavor.... I really like it!

MunifTanjim:
> Do nothing and continue to encourage users to override the handler. Not friendly, but easy and strictly "correct".

I guess we can try to make it easier to configure in Neovim core? Maybe by adding a callback function here (or somewhere around it)?

https://github.com/neovim/neovim/blob/46009499afbb0232124072d775caa9552d0f71de/runtime/lua/vim/lsp/diagnostic.lua#L1099-L1105

jose-elias-alvarez:
That's a good point. If the hurdle is that configuring it via Neovim isn't beginner-friendly, maybe we can improve the situation on the Neovim side. I can imagine adding an optional callback that would be called for each diagnostic, and we could potentially hook into that through null-ls to make things like per-source configuration a bit easier. I'll give it some more thought and maybe put together a PR for Neovim when I can.

jose-elias-alvarez:
So since null-ls already handles some things that are (relatively) straightforward to configure for experienced users, e.g. async formatting, I figured it made sense to add this, too, especially since the implementation is simple and adds very little complexity. I still want to contribute to improving diagnostics customization in Neovim itself, but for now I'm happy with this. Thanks to everyone for the feedback!

PR #44, opened 2021-07-23
Add vint as a built in diagnostics source

ranebrown:


jose-elias-alvarez:
LGTM! I'm happy to merge as-is, but we have a `to_temp_file` option that (if I'm understanding the problem correctly) would solve the `stdin` issue. Would that be worth trying?

ranebrown:
I switched to using the `to_temp_file` option. I didn't see that was available. Thanks!

jose-elias-alvarez:
Great! That should provide a better experience if there's issues with `stdin`.

PR #48, opened 2021-07-23
[WIP] diagnotics: support 'errorformat' style diagnostics

bfredl:
Note: severity not tested yet, as `zig ast-check` only has errors. I will try to support lualint and clint of the neovim repo for more coverage.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -409,7 +409,7 @@ M.misspell = h.make_builtin({
     method = DIAGNOSTICS,
     filetypes = { "*" },
     generator_opts = {
-        command = "misspell",
+        command = "zig",
         to_stdin = true,
         args = {},
         format = "line",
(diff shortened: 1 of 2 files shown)


folke:
That's awesome! Efm in null-ls 🙂

jose-elias-alvarez:
Wow, great idea! Let me know if I can provide any help.

jandamm:
I've done a similar thing but I used a function to parse the raw output with `vim.fn.getqflist()`:

```lua
local from_errorformat = function(efm, source)
    return function(params, done)
        local output = params.output
        if not output then
            return done()
        end

        local diagnostics = {}
        local lines = vim.split(output, "\n")

        local qflist = vim.fn.getqflist({ efm = efm, lines = lines })
        local severities = {e=1, w=2, i=3, n=4}

        for _, item in pairs(qflist.items) do
            if item.valid == 1 then
                table.insert(diagnostics, {
                    row = item.lnum,
                    col = item.col,
                    source = source,
                    message = item.text,
                    severity = severities[item.type],
                })
            end
        end
        return done(diagnostics)
    end
end


local jsonlint = helpers.make_builtin({
    filetypes = { "json" },
    generator_opts = {
        command = "jsonlint",
        args = { "--compact", },
        to_stdin = true,
        from_stderr = true,
        on_output = from_errorformat("%ELine %l:%c,%Z\\s%#Reason: %m,%C%.%#,%Eline %l\\, col %c\\, %m,%E%f line %l\\, col %c\\, %m,%-G%.%#", "jsonlint")
    },
})
```

This does work with severities so yours should work as well :)

jose-elias-alvarez:
This feature was implemented in #304.

PR #55, opened 2021-07-25
Expose reusable function

MunifTanjim:
Exposes two functions:

- `require("null-ls.formatting").apply_edits`
- `require("null-ls.generators").run`

With this change, plugins built on top of `null-ls` will be able to use these as utility functions even if they don't want to use the LSP setup. For example:

- Provide command/keymaps to run formatters bypassing the LSP stuffs.


jose-elias-alvarez:
Looks good to me! I really like the idea of running generators directly, and the implementation makes sense. Thank you for your contributions!

Issue #51, opened 2021-07-24
Support eslint as a formatter


ChristianChiarulli:
seems like eslint_d is supported why not eslint?

jose-elias-alvarez:
Vanilla ESLint can only format either by directly writing the file (which isn't desirable when editing it) or by outputting to JSON with the [`--fix-dry-run` flag](https://eslint.org/docs/user-guide/command-line-interface#--fix-dry-run). `eslint_d` can format directly to `stdout`, but `eslint` requires an additional step to parse JSON output and return the formatted content. It wouldn't be too hard to implement, though, so a PR would be welcome.

jose-elias-alvarez:
Closed in 9cc156e82b7ad3e2a783668fd634df31391396a6 (though as I mention in the documentation, I really don't recommend using it if `eslint_d` is an option).

PR #57, opened 2021-07-26
[Feature] provide diagnostic helpers

Tastyep:
Hello :wave: 

I implemented helper functions to implement new diagnostic providers with more ease.
I also refactored the existing ones to showcase the added value.
<del>
I was unfortunately not able to test all linters, so if you could give me a hand on this that would be super helpful.
I tested:
- shellcheck
- selene
- eslint
- flake8
- misspell
- vint
- teal (unitary)
- write-good (unitary)
</del>

All linters are now unit-tested :tada: 

I didn't refactor the implementation for vale as I never managed to make it report something.
I hope you like it, tell me what you think.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -35,7 +35,7 @@ local make_attr_adapters = function(severities)
             end
 
             _, end_col = line:find(quote, 1, true)
-            return end_col
+            return end_col > tonumber(entries["col"]) and end_col or nil
         end,
     }
 end
@@ -69,6 +69,7 @@ local from_pattern = function(pattern, groups, severities, defaults)
     return function(line, params)
         local results = { line:match(pattern) }
         local entries = {}
+
         for i, match in ipairs(results) do
             entries[groups[i]] = match
         end
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -5,6 +5,80 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
 local M = {}
 
+local default_severities = {
+    ["error"] = 1,
+    ["warning"] = 2,
+    ["information"] = 3,
+    ["hint"] = 4,
+}
+
+--- Parse a linter's output using a regex pattern
+-- @param pattern The regex pattern
+-- @param groups The groups defined by the pattern: {"line", "message", "col", ["end_col"], ["code"], ["severity"]}
+-- @param severities An optional table mapping the severity values to their codes
+-- @param defaults An optional table of diagnostic default values
+local from_pattern = function(pattern, groups, severities, defaults)
+    local group_handlers = {
+        row = function(entries)
+            return tonumber(entries["row"])
+        end,
+        col = function(entries)
+            return tonumber(entries["col"])
+        end,
+        end_col = function(entries)
+            return tonumber(entries["end_col"]) or tonumber(entries["col"])
+        end,
+        code = function(entries)
+            return entries["code"]
+        end,
+        severity = function(entries)
+            local severity = entries["severity"]
+            return severities[severity] or defaults["severity"] or default_severities["error"]
+        end,
+        message = function(entries)
+            return entries["message"]
+        end,
+    }
+
+    severities = severities or {}
+    defaults = defaults or {}
+    return function(line, params)
+        local results = { line:match(pattern) }
+        local entries = {}
+
+        for i, match in ipairs(results) do
+            entries[groups[i]] = match
+        end
+        if not (entries["row"] and entries["message"]) then
+            return nil
+        end
+
+        local diagnostic = {}
+        for key, handler in pairs(group_handlers) do
+            diagnostic[key] = defaults[key] or handler(entries)
+        end
+
+        return diagnostic
+    end
+end
+
+--- Parse a linter's output using multiple regex patterns until one matches
+-- @param patterns The regex pattern list
+-- @param groups The groups list defined by the patterns
+-- @param severities An optional table mapping the severity values to their codes
+-- @param defaults An optional table of diagnostic default values
+local from_patterns = function(patterns, groups, severities, defaults)
+    return function()
+        for i, pattern in ipairs(patterns) do
+            local diagnostic = from_pattern(pattern, groups[i], severities, defaults)
+            if diagnostic then
+                return diagnostic
+            end
+        end
+        return nil
+    end
+end
+
 M.write_good = h.make_builtin({
     method = DIAGNOSTICS,
     filetypes = { "markdown" },
@@ -15,32 +89,12 @@ M.write_good = h.make_builtin({
         check_exit_code = function(code)
             return code == 0 or code == 255
         end,
-        on_output = function(line, params)
-            local pos = vim.split(string.match(line, "%d+:%d+"), ":")
-            local row = pos[1]
-
-            local message = string.match(line, ":([^:]+)$")
-
-            local col, end_col
-            local issue = string.match(line, '%b""')
-            local issue_line = params.content[tonumber(row)]
-            if issue and issue_line then
-                local issue_start, issue_end = string.find(issue_line, string.match(issue, '([^"]+)'))
-                if issue_start and issue_end then
-                    col = tonumber(issue_start) - 1
-                    end_col = issue_end
-                end
-            end
-
-            return {
-                row = row,
-                col = col,
-                end_col = end_col,
-                message = message,
-                severity = 1,
-                source = "write-good",
-            }
-        end,
+        on_output = from_pattern(
+            [[(%d+):(%d+):(.*)]], --
+            { "row", "col", "message" },
+            nil,
+            { source = "write-good" }
+        ),
     },
     factory = h.generator_factory,
 })
@@ -57,33 +111,12 @@ M.markdownlint = h.make_builtin({
         check_exit_code = function(code)
             return code <= 1
         end,
-        on_output = function(line, params)
-            local split = vim.split(line, " ")
-            local rule = split[2]
-            local _, rule_end = string.find(line, rule, nil, true)
-            local message = string.sub(line, rule_end + 2)
-
-            local pos = vim.split(split[1], ":")
-            local row = pos[2]
-            local col = pos[3]
-
-            local end_col
-            local issue = string.match(line, "%b[]")
-            if issue then
-                issue = string.sub(issue, 2, #issue - 1)
-                local issue_line = params.content[tonumber(row)]
-                _, end_col = string.find(issue_line, issue, nil, true)
-            end
-
-            return {
-                row = row,
-                col = col and col - 1,
-                end_col = end_col,
-                message = message,
-                severity = 1,
-                source = "markdownlint",
-            }
-        end,
+        on_output = from_patterns(
+            { [[:(%d+):(%d+) [%w-/]+ (.*)]], [[:(%d+) [%w-/]+ (.*)]] },
+            { { "row", "col", "message" }, { "row", "message" } },
+            nil,
+            { source = "markdownlint" }
+        ),
     },
     factory = h.generator_factory,
 })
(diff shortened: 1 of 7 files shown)


lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,115 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
...
 local M = {}
 
+local default_severities = {
+    ["error"] = 1,
+    ["warning"] = 2,
+    ["information"] = 3,
+    ["hint"] = 4,
+}
+
+local make_attribute_handlers = function(severities, defaults)
```
jose-elias-alvarez, line 15: I think this code might be superfluous, since we have [a similar function](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/f93e0d9c03e67010591cd15bc0d7500886dcebf7/lua/null-ls/diagnostics.lua#L8) in the diagnostics handler.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,115 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
...
+        end,
+    }
+end
+
+--- Parse a linter's output using a regex pattern
+-- @param pattern The regex pattern
+-- @param groups The groups defined by the pattern: {"line", "message", "col", ["end_col"], ["code"], ["severity"]}
+-- @param severities An optional table mapping the severity values to their codes
+-- @param defaults An optional table of diagnostic default values
+local from_pattern = function(pattern, groups, severities, defaults)
```
jose-elias-alvarez, line 47: What do you think about changing the order of `severities` and `defaults`, since it seems more likely that the former will be `nil`?

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -15,32 +124,12 @@ M.write_good = h.make_builtin({
...
-                message = message,
-                severity = 1,
-                source = "write-good",
-            }
-        end,
+        on_output = from_pattern(
+            [[(%d+):(%d+):(.*)]], --
+            { "row", "col", "message" },
+            nil,
+            { source = "write-good" }
```
jose-elias-alvarez, line 131: I wonder if we could refactor `make_builtin` itself to accept a `source` key that gets passed to the diagnostics handler, since it's something we always want to include.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -57,33 +146,12 @@ M.markdownlint = h.make_builtin({
...
-            return {
-                row = row,
-                col = col and col - 1,
-                end_col = end_col,
-                message = message,
-                severity = 1,
-                source = "markdownlint",
-            }
-        end,
+        on_output = from_patterns(
```
jose-elias-alvarez, line 149: One thing about this refactor is that we lose the potential to find an `end_col`, which is useful for themes that include diagnostic highlighting (e.g. [Sonokai](https://github.com/sainnhe/sonokai)) or for integrations that show it (e.g. [trouble.nvim](https://github.com/folke/trouble.nvim)).

I think there's definitely room to improve the code quality of the built-in (it's one of the first ones I wrote, and I'm terrible at string matching), but if we can't think of a way to include similar functionality in the new helpers, I'd rather leave this one alone for now.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -121,61 +189,17 @@ M.teal = h.make_builtin({
...
-                                source = "tl check",
-                            })
-                        end
-                    end
-                end
-            end
-
-            return done(diagnostics)
-        end,
+        on_output = from_pattern(
```
jose-elias-alvarez, line 198: We're also losing a potential `end_col` value here, though I'm amazed at how many lines of code we were able to cut out otherwise.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -191,25 +215,21 @@ M.shellcheck = h.make_builtin({
...
-                    severity = diagnostic.level == "error" and 1
-                        or diagnostic.level == "warning" and 2
-                        or diagnostic.level == "info" and 3
-                        or diagnostic.level == "style" and 4,
-                })
-            end
-            return diagnostics
-        end,
+        on_output = from_json({
+            row = "line",
```
jose-elias-alvarez, line 219: Do you think it makes sense to have the position keys (`line`, `column`, `endLine`, and `endColumn`) as defaults, since so many sources use them?

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -219,61 +239,20 @@ M.selene = h.make_builtin({
...
-                            severity = (diagnostic.severity == "Error" and 1)
-                                or (diagnostic.severity == "Warning" and 2)
-                                or 4,
-                        })
-                    end
-                end
-            end
-            done(ret)
-        end,
+        on_output = from_pattern(
```
jose-elias-alvarez, line 250: I didn't write this one, but I believe we are also losing `end_col` here.

jose-elias-alvarez:
This looks great! With recent contributions, I noted that there was a lot of repetition in how sources were defined, so this really helps cut down on the amount of boilerplate needed. Ideally, I want to go through these when I have time and verify that everything is working. In the meantime, I've left some comments.

lua/null-ls/builtins/diagnostics.lua
Tastyep: We could probably reuse `generator_opts.command`

lua/null-ls/builtins/diagnostics.lua
Tastyep: Matching end_col shouldn;t be a problem, I notice I simply forgot to specify it in the group parameter --'

lua/null-ls/builtins/diagnostics.lua
Tastyep: Why not, instead of providing the full mapping, we could provide a default and the user would just provide overrides.
I thought about doing this for severity levels too, for example here we could just provide `style = 4`

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: The issue with (at least some of) these is that `end_col` isn't actually available – instead, we parsed the relevant line from the file to find the information. Lots of sources don't specify a final column, unfortunately, so I wonder if there's something we can do here.

jose-elias-alvarez:
Also, I thought of this in relation to #51, but I think it would be great to extract the JSON functions beyond diagnostics, since I think they could also be helpful for formatters / code actions.

lua/null-ls/builtins/diagnostics.lua
Tastyep: I would keep this order now that the source is defaulted to the command, if that's fine with you.

Tastyep:
> Also, I thought of this in relation to #51, but I think it would be great to extract the JSON functions beyond diagnostics, since I think they could also be helpful for formatters / code actions.

I would prefer to do this in a separate PR if you don't mind :sweat_smile:

lua/null-ls/builtins/diagnostics.lua
Tastyep: I added a new `quote` attribute that is used for parsing  `end_col` in attr_adapters
https://github.com/jose-elias-alvarez/null-ls.nvim/pull/57/commits/4dc6b1f548f2de96720507d29e8faae1607a8fa3

Tastyep:
I think I'm done

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: Yep, makes sense to me.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,120 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
...
+local from_patterns = function(patterns, groups, severities, defaults)
+    return function()
+        for i, pattern in ipairs(patterns) do
+            local diagnostic = from_pattern(pattern, groups[i], severities, defaults)
+            if diagnostic then
+                return diagnostic
+            end
+        end
+        return nil
+    end
```
jose-elias-alvarez, line 91: I think this callback is missing the arguments it receives from `on_output`, right? I had to change it a bit to get it to work on my end:

```suggestion
    return function(line, params)
        for i, pattern in ipairs(patterns) do
            local diagnostic = from_pattern(pattern, groups[i], severities, defaults)(line, params)
            if diagnostic then
                return diagnostic
            end
        end
        return nil
    end
```

lua/null-ls/builtins/diagnostics.lua
Tastyep: Actually, I did that for write-good and forgot markdownlint, I'll update it now

lua/null-ls/generators.lua
```diff
@@ -35,7 +35,7 @@ M.run = function(generators, params, postprocess, callback)
                         elseif type(results) == "table" then
                             for _, result in ipairs(results) do
                                 if postprocess then
-                                    postprocess(result)
+                                    postprocess(result, generator)
```
jose-elias-alvarez, line 38: I think it would be more flexible if we assign the command to `params` and pass that as the second argument to `postprocess` instead, since other than this I can't think of a situation where we would want to give handlers access to the generators directly.

jose-elias-alvarez:
Thank you for your hard work on this! The changes look good to me overall, and we can definitely wait to expand JSON parsing to other methods.

I think that for now, I'd prefer to leave existing diagnostic sources untouched and refactor them using the new helpers as we go. I've run into a couple of small issues while testing out a couple of sources, and since there's so much variation in CLI output, I'd prefer to have contributors test them thoroughly and make sure that everything is working as expected (I'll work on refactoring some of the sources I use, so we can have examples). Does that make sense to you?

Tastyep:
Thank you for the review :)

Adapting the existing linters represent most of the work, so I would feel a bit bad reverting the change.
However, I completely understand your position, since I wasn;t even able to test everything.
I was thinking that we could wait a bit more before merging this so that I can implement unit tests for the linters.

Would you be more ok with merging this if tests were provided ?

jose-elias-alvarez:
Right, I understand. I think either way adding coverage for the core helpers is a good idea, and then I like the idea of writing unit tests for sources that parse line output from linters, e.g. for `markdownlint` we could feed it some of the lines from [the rules documentation](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md). For JSON linters, once we've covered the helpers, I don't think we need to unit test them specifically, since their behavior is more structured. How does that sound?

Tastyep:
Good to me :+1:

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -191,25 +214,9 @@ M.shellcheck = h.make_builtin({
...
-                    severity = diagnostic.level == "error" and 1
-                        or diagnostic.level == "warning" and 2
-                        or diagnostic.level == "info" and 3
-                        or diagnostic.level == "style" and 4,
-                })
-            end
-            return diagnostics
-        end,
+        on_output = from_json({}, {
+            style = 4,
```
jose-elias-alvarez, line 218: Small bug here: `shellcheck` uses `info`, not `information`.

Tastyep:
I ended up testing the json linters too, I guess that you don't mind, right ?

lua/null-ls/generators.lua
Tastyep: I'm not sure how to do that, from what I've seen, I still need to set it on generator (that's when I have access to the command). Do you want to do smth like `params["command"] = generator.command` and then pass params instead ?

jose-elias-alvarez: Yep, sorry, that's what I meant. We have access to both in [`generator_factory`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/9b8458bd1648e84169a7e8638091ba15c2f20fc0/lua/null-ls/helpers.lua#L72), so I think that would be the place to set it, and then we can pick it up in the handler.

jose-elias-alvarez:
Fantastic! Everything looks great, and all of the linters I've tested seem to be working perfectly. 

One final thing before we get this merged – can we make sure that all of the other tests are passing? I'm getting failures from the handlers for `diagnostics`, `code_actions`, and `formatting`, and there's an additional small failure in the E2E tests because the Teal source name has changed. (There's another failure in a cache test that you don't need to worry about.)

Tastyep:
All these tests are already failing on `main`, I think there is only one test that I need to update on my side

jose-elias-alvarez:
My bad! You are right. It's just the `tl check` test, then.

test/spec/e2e_spec.lua
```diff
@@ -230,10 +230,10 @@ describe("e2e", function()
 
             local tl_check_diagnostic = buf_diagnostics[1]
             assert.equals(tl_check_diagnostic.message, "in return value: got string, expected number")
-            assert.equals(tl_check_diagnostic.source, "tl check")
+            assert.equals(tl_check_diagnostic.source, "tl")
             assert.same(tl_check_diagnostic.range, {
-                start = { character = 52, line = 0 },
-                ["end"] = { character = 54, line = 0 },
+                start = { character = 53, line = 0 },
+                ["end"] = { character = -1, line = 0 },
```
Tastyep, line 236: The diagnostic output is
`/tmp/lua_arziS1:1:53: in return value: got string, expected number`
So end_col is not available

jose-elias-alvarez: Right, I see the issue. The original source basically kept going forwards from `start_col` until it reached a space, so we won't have access to it for now. It's fine to change this to fit the new format, I think.

jose-elias-alvarez: On my end, this diff doesn't fix the test – `end_col` is set to `22`, for some reason. Is that just me?

jose-elias-alvarez: Whoops, didn't pull the latest changes. Looks good then!

Tastyep: Did you pull ?
https://github.com/jose-elias-alvarez/null-ls.nvim/pull/57/commits/383c7ca09be175e3272b39babd2df5283579974c#diff-52a4752613765e712de54d7e39fb1bd2891390845c197a4ecc93310b54ad1572R38

jose-elias-alvarez:
Thanks again for all of your work on this!

PR #58, opened 2021-07-26
[Feature] Add vue filetype support to builtins

PasiBergman:
Vue.js is typically handled with the same linter (diagnostics) and formatter as other JavaScript family files in the project.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -88,7 +88,7 @@ Supports both `textDocument/formatting` and `textDocument/rangeFormatting`
 local sources = {null_ls.builtins.formatting.prettier}
 ```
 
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
LGTM! I don't work with Vue, but could we quickly verify that `prettierd` and `prettier_d_slim` work as expected with Vue files? In the past I've found that their behavior can differ from vanilla `prettier`, so I'd just like to verify before we document it.

PasiBergman:
@jose-elias-alvarez 
Good point. I've tested formatting .vue file(s) with both `prettierd` and `prettier_d_slim`. In my tests I noticed, that both are able to format .vue files. I found some general issues with `prettier_d_slim`, but the issues where not related to the filetype.

gegoune:
Likewise, using prettierd with vue (overrides default with `:with` and it works just fine.

jose-elias-alvarez:
Great, thanks for confirming!

PR #59, opened 2021-07-26
feat: add chktex diagnostics

nawordar:


jose-elias-alvarez:
This looks good. At the moment, we have an open PR, #57, that will make parsing output from well-behaving linters much simpler. I'm happy to merge this as-is, but would you prefer to wait until that PR is merged?

nawordar:
That is a good idea. I will update it, after that PR gets merged

jose-elias-alvarez:
We've merged the PR! Take a look at the examples using the new helpers.

Tastyep:
Handling the length attribute might be a bit tricky, I'll add a way to provide user defined attr adapters.

This would look like that:

``` lua
adapter = {
end_col = function(entries, _) return tonumber(entries["col"]) + tonumber(entries["_length"]) end
}
```

nawordar:
@Tastyep I have a problem caused by `chktex` using 1-based column numbers. I tried to add two diagnostic adapters: `from_1_based` and `from_length` and add them to `adapters` array. Unfortunately, `from_length` receives the original column number, so `end_col` still ends up being off.

```lua
local diagnostic_adapters = {
    col = {
        from_1_based = {
            col = function(entries)
                local col = tonumber(entries["col"])
                return col - 1
            end,
        },
    },
    end_col = {
        -- ...
        from_length = {
            end_col = function(entries)
                local col = tonumber(entries["col"])
                local length = tonumber(entries["_length"])
                return col + length
            end,
        },
    },
}

-- ...

M.chktex = h.make_builtin({
    method = DIAGNOSTICS,
    filetypes = { "tex" },
    generator_opts = {
        -- ...
        on_output = from_pattern(
            [[(%d+):(%d+):(%d+):(%w+):(.+)]], --
            { "row", "col", "_length", "severity", "message" },
            {
                adapters = {
                    diagnostic_adapters.col.from_1_based,
                    diagnostic_adapters.end_col.from_length,
                },
                severities = {
                    Error = default_severities["error"],
                    Warning = default_severities["warning"],
                },
            }
        ),
    },
})
```

jose-elias-alvarez:
@nawordar There's a lot of inconsistency in how linters report positions, but in #68, null-ls will now expect all positions passed to it to be 1-indexed. In that PR I also added an `overrides.offsets` options to more easily handle cases like this. If I understand correctly, I think the changes should let `chktex` work without modifying offsets, so I'll let you know once we've merged it.

nawordar:
There is something weird going on. I had thought that `chktex` reports different position than other linters, but that is actually not the case.

I wrote the following test:

```lua
describe("chktex", function()
    local linter = diagnostics.chktex
    local parser = linter._opts.on_output
    local file = {
        [[\documentclass{article}]],
        [[]],
        [[\begin{document}]],
        [[Lorem ipsum dolor \sit amet]],
        [[\end{document}]],
    }

    it("should create a diagnostic", function()
        local output = [[3:23:1:Warning:Command terminated with space.]]
        local diagnostic = parser(output, { content = file })
        assert.are.same({
            row = "3",
            col = 23,
            end_col = 24,
            severity = 1,
            message = "Command terminated with space.",
        }, diagnostic)
    end)
end)
```
When I copied the test "file" string content to a real file and opened it in Neovim, it reports the diagnostic on a column 24

A screenshot of Neovim reporting wrong diagnostic position

jose-elias-alvarez:
I just tried your code on the PR I linked, and if I remove `diagnostic_adapters.col.from_1_based,`, I get the correct position, so I think that'll solve it!

nawordar:
The wrong severity is a mistake in the test, I will change that :)

Tastyep:
Alright, that's because lua tables don't keep the insertion order (at least with the ones with non numeric keys), so I wouldn;t be surprised if the second adapter gets called before the first one, I'll see what I can do about that.

Otherwise Like @jose-elias-alvarez said `from_1_based` can be removed.
Sorry for changing so many things ^^

nawordar:
Generator test shows that there is one error:

```
Testing:        /home/nawordar/projects/null-ls.nvim/test/spec/generators_spec.lua
Success ||      generators run should immediately return when method has no registered generators

Success:        1
Failed :        0
Errors :        1
```

Does it have anything to do with my PR?

jose-elias-alvarez:
Nope, that's on my end (and nice catch – it's really hard to see Plenary errors in a giant list). I'll look into it.

nawordar:
OK, I've updated the PR. I also noticed, that when `chktex` reports a diagnostic with severity "Error", it exits with error code 1, so I added a `check_exit_code` function

jose-elias-alvarez:
Awesome, looks perfect! Thank you for the PR (and for your patience dealing with our changes)!

PR #65, opened 2021-07-28
[FIX]: add terraform ft, remove hcl

ChristianChiarulli:
Context: https://github.com/neovim/nvim-lspconfig/pull/1009


jose-elias-alvarez:
Not sure I understand the situation completely, but if this is the right call, I'm happy to merge it.

PR #66, opened 2021-07-29
Allow user defined diagnostic attribute adapters.

Tastyep:
This will allow the user to define its own diagnostic attribute adapters like [here](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/59#issuecomment-888865823).
This should make the parser flexible enough to support any kind of edge case.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -22,22 +22,35 @@ local default_json_attributes = {
     message = "message",
 }
 
-local make_attr_adapters = function(severities)
-    return {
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Great! As you say this should let us handle pretty much anything via string matching, and I like consolidating everything into`overrides`, too.

Tastyep:
Oops, I should not have opened the PR this early since I had other improvements in mind.
I'll open another PR :wink:

jose-elias-alvarez:
FYI I'm also working on a PR to consolidate how range conversion is handled that will also affect some of the new diagnostic handlers.

PR #67, opened 2021-07-29
Feature/diagnostic attribute adapters multiple patterns

Tastyep:
Make `from_patterns` accept overrides for each pattern.
Adds a fallback pattern for `teal` which has a very inconsistent output.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -245,14 +245,21 @@ M.teal = h.make_builtin({
         end,
         to_stderr = true,
         to_temp_file = true,
-        on_output = from_pattern(
-            [[:(%d+):(%d+): (.* ['"]*([%w%.%-]+)['"]*)]], --
-            { "row", "col", "message", "_quote" },
+        on_output = from_patterns({
             {
-                adapters = { diagnostic_adapters.end_col.from_quote },
-                diagnostic = { source = "tl check" },
-            }
-        ),
+                pattern = [[:(%d+):(%d+): (.* ['"]?([%w%.%-]+)['"]?)$]], --
+                groups = { "row", "col", "message", "_quote" },
+                overrides = {
+                    adapters = { diagnostic_adapters.end_col.from_quote },
+                    diagnostic = { source = "tl check" },
+                },
+            },
+            {
+                pattern = [[:(%d+):(%d+): (.*)]], --
+                groups = { "row", "col", "message" },
+                overrides = { diagnostic = { source = "tl check" } },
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Could you review #68 when you get a chance? I imagine we'll need to resolve conflicts in this one once that one is merged.

jose-elias-alvarez:
Other than that, this looks good to me. I wonder if it makes sense to do a type check so that we can also keep the original behavior of simply matching against multiple patterns (or perhaps splitting it up into two separate functions).

Tastyep:
> Other than that, this looks good to me. I wonder if it makes sense to do a type check so that we can also keep the original behavior of simply matching against multiple patterns (or perhaps we could split it up into two separate functions, if you think the original still handles cases we are likely to see).

I'm not sure if I'm understanding you, can you put a comment below the line of code where you would put a type check ?

jose-elias-alvarez:
Something like this is what I had in mind:

```lua
for i, matcher in ipairs(matchers) do
    local diagnostic
    if type(matcher) == "string" then
        diagnostic = from_pattern(pattern, groups[i], overrides)(line, params)
    else
        diagnostic = from_pattern(matcher.pattern, matcher.groups, matcher.overrides)(line, params)
    end
end
```

The idea was to keep the original functionality for cases like `markdownlint`, where we don't have to specify an override for one pattern, but maybe that's not useful (or maybe the whole idea doesn't make sense).

Tastyep:
That still wouldn't be possible since `from_patterns` accepts a list of matchers (which contains the groups).

IMO this is better
``` lua
from_patterns(
{
            {
                pattern = [[:(%d+):(%d+) [%w-/]+ (.*)]],
                groups = { "row", "col", "message" },
            },
            {
                pattern = [[:(%d+) [%w-/]+ (.*)]],
                groups = { "row", "message" },
            },
})
```
than
``` lua
from_patterns(
    { [[:(%d+):(%d+) [%w-/]+ (.*)]], [[:(%d+) [%w-/]+ (.*)]]},
    { { "row", "col", "message" },  { "row", "message" }}
)
```

But tell me if you think otherwise

jose-elias-alvarez:
Not at all – totally up to you. If you think it's better let's go for it.

jose-elias-alvarez:
Everything looks good to me, just give me a heads-up when it's ready to be merged.

Tastyep:
I think it's good :+1:

jose-elias-alvarez:
Great! Thanks again! I think these changes will make it much easier for contributors to add new diagnostic sources, which will make a big difference.

PR #68, opened 2021-07-29
Consolidate range conversion

jose-elias-alvarez:
The main goal of this PR is to consolidate range conversion by clarifying that diagnostics provided to null-ls should be 1-indexed and sources are therefore responsible for offsetting their ranges if that's not the case.

As part of that, I've updated the new diagnostic helpers from #57 to allow sources to define optional offsets, which if present will be applied to each diagnostic before returning to the null-ls diagnostics handler. The handler is now only responsible for converting strings to numbers, setting defaults if necessary, and converting Lua ranges to the correct LSP format. 

So far, I've found that `write-good` uses 0-indexed columns. Other line-based built-ins probably need to be offset, too, but I think the best course of action is to rely on users to find and report issues. I don't believe we had any sources that didn't use 1-indexed rows, and incorrect columns is annoying but not the end of the world, so we can fix it as we go. 

I've also bundled in a fix for an issue involving default command names. I forgot that `params` is generated only once per LSP request, so if the user has defined multiple diagnostic generators, `command` will be overwritten. I've removed `command` from `params` completely and re-written the diagnostic helpers to require `command` as their first argument, which is more verbose but avoids unnecessary complexity. 

@Tastyep Let me know what you think so I can merge this and you can fix any conflicts with your new PR.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -90,7 +90,7 @@ local from_pattern = function(pattern, groups, overrides)
     overrides = overrides or {}
     local severities = vim.tbl_extend("force", default_severities, overrides.severities or {})
     local defaults = overrides.diagnostic or {}
-    defaults.source = source
+    local offsets = overrides.offsets or {}
     local attr_adapters = make_attr_adapters(severities, overrides.adapters or {})
 
     return function(line, params)
@@ -101,7 +101,7 @@ local from_pattern = function(pattern, groups, overrides)
             entries[groups[i]] = match
         end
 
-        return make_diagnostic(entries, defaults, attr_adapters, params, overrides.offsets or {})
+        return make_diagnostic(entries, defaults, attr_adapters, params, offsets)
     end
 end
 
@@ -137,7 +137,7 @@ local from_json = function(overrides)
     local attributes = vim.tbl_extend("force", default_json_attributes, overrides.attributes or {})
     local severities = vim.tbl_extend("force", default_severities, overrides.severities or {})
     local defaults = overrides.diagnostic or {}
-    defaults.source = source
+    local offsets = overrides.offsets or {}
     local attr_adapters = make_attr_adapters(severities, overrides.adapters or {})
 
     return function(params)
@@ -148,7 +148,7 @@ local from_json = function(overrides)
                 entries[attr] = json_diagnostic[json_key]
             end
 
-            local diagnostic = make_diagnostic(entries, defaults, attr_adapters, params, overrides.offsets or {})
+            local diagnostic = make_diagnostic(entries, defaults, attr_adapters, params, offsets)
             if diagnostic then
                 table.insert(diagnostics, diagnostic)
             end
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -90,6 +90,7 @@ local from_pattern = function(pattern, groups, overrides)
     overrides = overrides or {}
     local severities = vim.tbl_extend("force", default_severities, overrides.severities or {})
     local defaults = overrides.diagnostic or {}
+    defaults.source = source
     local attr_adapters = make_attr_adapters(severities, overrides.adapters or {})
 
     return function(line, params)
@@ -136,6 +137,7 @@ local from_json = function(overrides)
     local attributes = vim.tbl_extend("force", default_json_attributes, overrides.attributes or {})
     local severities = vim.tbl_extend("force", default_severities, overrides.severities or {})
     local defaults = overrides.diagnostic or {}
+    defaults.source = source
     local attr_adapters = make_attr_adapters(severities, overrides.adapters or {})
 
     return function(params)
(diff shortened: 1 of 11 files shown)


lua/null-ls/builtins/diagnostics.lua
```diff
@@ -34,7 +34,7 @@ local diagnostic_adapters = {
                 end
 
                 _, end_col = line:find(quote, 1, true)
-                return end_col > tonumber(entries["col"]) and end_col or nil
+                return end_col and end_col > tonumber(entries["col"]) and end_col or nil
```
Tastyep, line 37: good catch :+1:

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -71,20 +71,27 @@ local make_diagnostic = function(entries, defaults, attr_adapters, params)
...
+        diagnostic[k] = diagnostic[k] and diagnostic[k] + offset
+    end
+    return diagnostic
 end
 
 --- Parse a linter's output using a regex pattern
+-- @param source The name of the source
 -- @param pattern The regex pattern
 -- @param groups The groups defined by the pattern: {"line", "message", "col", ["end_col"], ["code"], ["severity"]}
 -- @param overrides A table providing overrides for {adapters, diagnostic, severities}
```
Tastyep, line 85: Missing offsets in the list

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -95,21 +102,23 @@ local from_pattern = function(pattern, groups, overrides)
             entries[groups[i]] = match
         end
 
-        return make_diagnostic(entries, defaults, attr_adapters, params)
+        return make_diagnostic(entries, defaults, attr_adapters, params, overrides.offsets or {})
```
Tastyep, line 105: Maybe move `overrides.offsets or {}` as `local offsets = overrides.offsets or {}` close to `local defaults = overrides.diagnostic or {}` and the others

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -119,16 +128,19 @@ local from_patterns = function(patterns, groups, overrides)
 end
 
 --- Parse a linter's output in JSON format
+-- @param source The name of the source
 -- @param overrides A table providing overrides for {adapters, attributes, diagnostic, severities}
```
Tastyep, line 132: List needs update

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -139,7 +151,7 @@ local from_json = function(overrides)
                 entries[attr] = json_diagnostic[json_key]
             end
 
-            local diagnostic = make_diagnostic(entries, defaults, attr_adapters, params)
+            local diagnostic = make_diagnostic(entries, defaults, attr_adapters, params, overrides.offsets or {})
```
Tastyep, line 154: Like for pattern, I would make a local variable on top to set it to {} if unspecified

lua/null-ls/diagnostics.lua
```diff
@@ -5,22 +5,21 @@ local generators = require("null-ls.generators")
...
-
-    return {
-        start = { line = start_line, character = start_char },
-        ["end"] = { line = end_line, character = end_char },
-    }
+    local row = u.string.to_number_safe(diagnostic.row, 1)
+    local col = u.string.to_number_safe(diagnostic.col, 1)
+    -- default end_row to row
+    local end_row = u.string.to_number_safe(diagnostic.end_row, row)
+    -- default end_col to -1, which wraps to the end of the line (after range.to_lsp conversion)
```
Tastyep, line 14: A bit tricky, but ok

Tastyep (changes requested):
LGTM :+1:  I added some comments addressing minor issues.

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: Maybe I wrote it incorrectly (I'm not familiar with the syntax) but I think I listed it at the bottom, under `overrides.offset`. Is that the wrong way to do it?

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: Agreed, moved.

jose-elias-alvarez:
Thanks for the feedback! I just force pushed a couple of additional changes, including one that re-institutes setting the default source name from `command` (I realized that passing the generator's opts directly would be useful for other features like #39). Sorry for the back-and-forth on that – for personal projects, I'm used to iterating quickly over the internals while I get it figured out, but from the perspective of these helpers, it should work just as before. 

Let me know how everything looks (I'm not sure if my description of `overrides.offsets` was somehow badly structured) and I'll merge.

lua/null-ls/builtins/diagnostics.lua
Tastyep: That was good ! I was talking about this comment `A table providing overrides for {adapters, diagnostic, severities}` <- the list is missing offsets, but all things considered, it's probably better to just rephrase it like `@param overrides A table providing overrides`

Tastyep: Feel free to merge it once this is done :+1:

jose-elias-alvarez: Ah, got it, thank you!

PR #69, opened 2021-07-30
Feature/parse more linters

### LINKED ISSUE #62: [Feature request] Add support for golangci-lint
Allow using golangci-lint as either formatter or linter.
This might turn out complicated, as golangci-lint is sometimes slow. This could be alleviated increasing concurrent threads via command line options.
It is also quite extensible, and it would need to be able to source a config file that is project specific.
I can also help, if this feature is deemed desirable.

Tastyep:
Add linters
- clang-tidy
- golangci-lint (which aggregates many more)

Sort linters in alphabetical order.
Update documentation to add missing linters.

I didn't test the added linters in an integrated setup, we should probably do that before considering merging.

Fix #62


lua/null-ls/builtins/diagnostics.lua
```diff
@@ -201,128 +201,157 @@ M.chktex = h.make_builtin({
...
-M.markdownlint = h.make_builtin({
+M.golangci_lint = h.make_builtin({
     method = DIAGNOSTICS,
-    filetypes = { "markdown" },
+    filetypes = { "go" },
     generator_opts = {
-        command = "markdownlint",
-        args = { "--stdin" },
+        command = "golangci-lint",
+        args = { "run" },
```
tiannaru, line 287:
```suggestion
        command = "golangci-lint",
        args = { "run", "$FILENAME"},
```
Without `$FILENAME` golangci-lint reports errors for the entire projec. But they will show up in the diagnostics for the current file, creating a lot of false positives.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -201,128 +201,157 @@ M.chktex = h.make_builtin({
...
         to_stdin = true,
```
tiannaru, line 288:
```suggestion
        to_stderr = true,
```
I understand that we need to use `to_stderr` for linters that return non-zero return codes. In my testing linting failed unless `to_stderr` was used. 
`to_stdin` hadn't had much effect, as `golangci-lint` doesn't support receiving data via stdin, it just wants filanames/package names.

tiannaru:
Tested the `golangci-lint` diagnostics on lunarvim using this project I put together for testing:
https://github.com/tiannaru/go-is-crazy
Looked great! Had a couple of issues, and submitted change suggestions for these.
There is one issue that was left unresolved: it seems that using the `.golangci.yml` config with `fix: true` messes with the opened file, as the diagnostic wasn't meant to make changes to the project. IMO, this is just nice to have and could be more easily fixed in the `golangci-ci` config or the pipeline setup, and should only be documented (I can also help with a documentation suggestion if you guys agree on this)
Finally, this PR solves #62, so thank you!

lua/null-ls/builtins/diagnostics.lua
rliebz: Running `golangci-lint run $FILENAME` will produce diagnostics for only that file, but unfortunately without intelligence from other files.

As an easy to reproduce example:

```sh
mkdir example && cd example
go mod init example
echo -e "package main\n\nfunc main() {}" > main_1.go
echo -e "package main\n\nfunc main() {}" > main_2.go
```

Running `golangci-lint run` produces:
```
main_2.go:3:6: `main` redeclared in this block (typecheck)
func main() {}
     ^
main_1.go:3:6: other declaration of main (typecheck)
func main() {}
     ^
```

But running `golangci-lint run main_1.go` does not find the error, because it is not aware of the other files in the project.

The ideal solution would be being able to run it at the package level, then filter down to only report errors from the open file.

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: I'll just mention that strictly speaking, `to_stderr` is independent from the exit code – it depends on the behavior of the linter itself (e.g. some linters will only output to `stderr` if linting itself failed but use exit codes to indicate results).

lua/null-ls/builtins/diagnostics.lua
jose-elias-alvarez: Filtering by filename should be simple. [The original feature request](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/62) mentions speed concerns, so I wonder about performance. null-ls updates diagnostics according to LSP behavior, but LSP behavior assumes we're sending updates for one file at a time to a language server, so slow-running, project-wide linters could cause issues.

jose-elias-alvarez:
If I understand correctly, having `fix: true` in a config file will cause `golangci-lint run` to automatically fix files, right? I see in [the documentation](https://golangci-lint.run/usage/configuration/) that

> The config file has lower priority than command-line options

Does that mean we could add `--fix=false` to the args to make sure we never write the file, regardless of the user's config?

jose-elias-alvarez:
After going through the documentation a bit, I tried this PR out locally using the repo @tiannaru linked, and it seems that that it has some less-than-ideal interactions with LSP behavior. For example, since it doesn't support stdin, changing a file doesn't actually update diagnostics, and saving a file won't trigger a diagnostics refresh since it's not a relevant event for actual LSP servers. Normally, we could get around this by using a temp file, but since it seems that proper error reporting depends on checking the whole project, it doesn't seem to be a good option in this case.

I'm not at all familiar with the linter (or Go in general) so I'll defer to you guys re: including this (and there may be workarounds to these issues), but in its current state, I don't know if it'll provide a good experience. Other than that, everything looks good.

tiannaru:
> Does that mean we could add `--fix=false` to the args to make sure we never write the file, regardless of the user's config?
Silly me, I thought you can only pass `--fix` with no values. 😅 
 Indeed, that fixes the issue.

Tastyep:
Thanks a lot for testing my work ! That repo you provided will be handy. Seeing that some information is not provided by all linters, I'll switch from pattern to JSON matching.

Let me test this and come back to your comments :)

lua/null-ls/builtins/diagnostics.lua
Tastyep: There are some things I don't understand, and that maybe you could explain.
if I run `cat cmd/main.go | golangci-lint run` it does work, doesn't it mean that it supports stdin ?

lua/null-ls/builtins/diagnostics.lua
Tastyep: If I run `cat cmd/main.go | golangci-lint run 2> /dev/null` I still get diagnostics from stdin, am I missing smth ?

tiannaru: Or I'm missing something 😄 
I suggested this null-ls because it fails without this option:
```
Error executing vim.schedule lua callback: ...e/pack/packer/start/null-ls.nvim/lua/nul
l-ls/helpers.lua:135: error in generator output: cmd/main.go:12:1: commentFormatting:
put a space between `//` and comment text (gocritic)
//bad comment formatting
...
```
Happens after I reinstall neovim and lunarvim with different versions, as well. So maybe there's a bug somewhere else and this is just the workaround? 😄 
Does it work for you without `to_stderr = true`?

jose-elias-alvarez:
The current commit is working for me, but diagnostics still aren't updated on change, which in my opinion is a pretty serious problem, since it breaks the user expectation of LSP-like behavior. Other plugins like [nvim-lint](https://github.com/mfussenegger/nvim-lint) support it, but it relies on a `BufWritePost` autocommand. What do you guys think?

lua/null-ls/builtins/diagnostics.lua
Tastyep: No, I need to make the change but I don't understand why, any idea @jose-elias-alvarez

Tastyep:
So, unfortunately, clang-tidy doesn't work either ^^, I'm getting the following error
```
error output: Error while trying to load a compilation database:
Could not auto-detect compilation database for file "/tmp/lua_Tk1s5M"                                                                                                                                                          
No compilation database found in /tmp or any parent directory
fixed-compilation-database: Error while opening fixed database: No such file or directory
json-compilation-database: Error while opening JSON database: No such file or directory
Running without flags.
Error while processing /tmp/lua_Tk1s5M.
error output: nil
output: error: unable to handle compilation, expected exactly one compiler job in '' [clang-diagnostic-error]
```

This is obtained by forcing the debug log of the wrapper declared in `helpers:L124`

Tastyep:
Regarding golang-ci I tried writing to a temp file but I don't get any diagnostic, I think we should try to make this work and also figure out why we need to specify `to_stderr=true` while it writes on `stdout` first.

tiannaru:
> Regarding golang-ci I tried writing to a temp file but I don't get any diagnostic, I think we should try to make this work and also figure out why we need to specify `to_stderr=true` while it writes on `stdout`

Indeed, `golangci` is quite the picky client 😄 . It wants the temp file to be in the project folder. I imagine it actually needs to match the base path exactly, or you'll get false positives; this would be quite hard to fix (ar at least, ugly). After working around this, `null-ls` started getting becoming much smoother, but also very buggy (false positives appearing, diagnostics disappearing after unrelated changes and so on)
In conclusion, @jose-elias-alvarez when I opened #62 I didn't expect `golangci-lint` to be so incompatible with LSP. I'll try fiddling with this over the weekend and maybe some time after, but don't have much hope for it, as it's too much trouble for what it's worth. 
I've closed the issue, since I can't see a way to implement `golangci-lint` with LSP behavior.

Tastyep:
Tested clang-tidy and it now works
It was not possible to parse a linter's output if it also wrote on stderr, I slightly changed the condition so that it's possible when `ignore_errors = true`

image

jose-elias-alvarez:
Glad it's working! I agree with both the change and renaming the variable, since I agree that it's confusing when paired with `to_stdin`. 

For `golangci`: I imagine there's other linters with the same requirements, i.e. they need to lint a file from disk and also need the file to be in a project folder for config / context.

If we're going to support this, we basically can't avoid refreshing on `BufWritePost`, since writing a temp file to the project directory is way too messy and could cause bad interactions with file watchers. I think the cleanest solution would be to add a config flag for built-ins (something like `on_buf_write_post`), have the diagnostics handler check if any generators have the flag, and define a buffer-local `BufWritePost` autocommand to send a `textDocument/didChange` notification.

My personal opinion is that it's not worth it, since the user experience of linting on save is not great once you're used to LSP behavior, and we'll get regular questions about why diagnostics aren't updating on change. If you want to include it, that's the solution I would be most comfortable with (it's also similar to the behavior of the VS Code extension, which updates diagnostics on save).

jose-elias-alvarez:
One additional note: I forgot that this existed, but while digging around Neovim's LSP code, I remembered that there is a `textDocument/onSave` notification as well, which would make handling this case much simpler (no autocommands).

jose-elias-alvarez:
A recent commit sorted all built-ins alphabetically and made sure they're all properly documented. #318 adds support for `golangci-lint` and is ready to go once we have `textDocument/didSave` support, so the one unmerged feature – support for `clang-tidy` – would be better in a more focused PR (anyone who works on C++ and wants to test it and put in a PR is very welcome to do so). Thank you again for the hard work on this!

PR #75, opened 2021-08-02
[WIP] feat: add conditional source registration

### LINKED ISSUE #60: Feature Request: conditional formatting
Would it be possible to select a formatter based on some conditions?

For example with `eslint` and `prettier`, a user could write a condition to check if there is `.eslintrc` present in the project, if yes, use `eslint`, if not use `prettier`. 

Awesome work on the plugin btw, its a breeze to set up compared to `efm`.

jose-elias-alvarez:
Will close #60 when done.

This is a basic implementation that handles the conditional formatting case described in #60 as follows:

```lua
local sources = {
    require("null-ls.helpers").conditional(function(utils)
        return utils.root_has_file(".eslintrc.js") and b.formatting.eslint_d or b.formatting.prettier
    end),
}
```

The above example shows a simple utility to check if a file exists in the current working directory. I mentioned in the linked issue that checking whether a command runs successfully is another option, and I'm open to adding more to handle common use cases. However, when it comes to handling more complicated cases (e.g. monorepos) or language-specific checks, I'd prefer to have users / integrations be responsible for writing the conditional logic. 

If registered unconditionally, the above example will run once whenever the user opens Neovim. A simple file existence check takes less than 1 millisecond, so I don't think it's a big deal, but for heavier conditional logic (or for users really obsessed with their startup time, an obsession I don't personally understand) it would be simple to call it conditionally on a `FileType` event:

```lua
_G.register_formatter = function()
    require("null-ls").register({ require("null-ls.helpers").conditional(function(utils)
        return utils.root_has_file(".eslintrc.js") and b.formatting.eslint_d or b.formatting.prettier
    end) })
end

vim.cmd("autocmd FileType javascript,typescript lua register_formatter()")
```

This is just a first pass, so I'm happy to hear feedback about the implementation and if there's any other common use cases we can handle for now.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -65,6 +65,46 @@ See [CONFIG](CONFIG.md) to learn about `diagnostics_format`. Note that
 specifying `diagnostics_format` for a built-in will override your global
 `diagnostics_format` for that source.
 
+## Conditional registration
+
+null-ls supports dynamic registration, meaning that you can register sources
+whenever you want. To simplify this, built-ins have access to the `conditional`
+option, which should be a function that returns a boolean or `nil` indicating
+whether null-ls should register the source. null-ls will pass a single argument
+to the function, which is a table of utilities to handle common conditional
+checks (though you can use whatever you want, as long as the return value
+matches).
+
+For example, to conditionally register `stylua` by checking if the root
+directory has a `stylua.toml` file:
+
+```lua
+local sources = {
+    null_ls.builtins.formatting.stylua.with({
+        condition = function(utils)
+            return utils.root_has_file("stylua.toml")
+        end,
+    }),
+}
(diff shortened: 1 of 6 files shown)


jose-elias-alvarez:
I also forgot to mention that the functionality in this PR is already possible to implement via dynamic registration, so fundamentally, I want this to add a little convenience for users, not handle every potential case. 

I also plan on adding a similar condition check into `with` to make it convenient to register individual sources conditionally, so the `conditional` helper is meant for cases where you want to register one of several possible options.

jose-elias-alvarez:
I'm pretty happy with this for now (I'm using the example I added to the documentation to avoid running `stylua` on random Lua files), so I'm going to merge it and see if we can continue to improve it as we go.

PR #77, opened 2021-08-02
Fix: set end column in vale properly

carrascomj:
### Description

It seems like the way of handling the column position in null-ls changed, so this needs to be updated in the [vale](https://github.com/errata-ai/vale) diagnostics plug-in.

### Implementation

It only required to sum one to the end column position to be correct.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -255,15 +255,18 @@ M.vale = h.make_builtin({
     generator_opts = {
         command = "vale",
         format = "json",
-        args = { "--no-exit", "--output=JSON", "$FILENAME" },
+        to_stdin = true,
+        args = function(params)
+			      return { "--no-exit", "--output", "JSON", "--ext", "." .. vim.fn.fnamemodify(params.bufname, ":e") }
+        end,
         on_output = function(params)
             local diagnostics = {}
             local severities = { error = 1, warning = 2, suggestion = 4 }
-            for _, diagnostic in ipairs(params.output[params.bufname]) do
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks! As you noted we had some changes to how diagnostic ranges were set, so I expected a few fixes like this would be necessary.

While we have this open: I was testing out Vale with [this sample repo](https://github.com/errata-ai/vale-boilerplate), and I noted that diagnostics don't seem to update on change. As far as I can tell, Vale reads the file from the disk if the filename is specified in the arguments, so the diagnostics won't update until the file is saved and then changed again. Is this also a problem on your end? 

I played around with it and was able to get it working from `stdin` with the following changes:

```diff
     generator_opts = {
         command = "vale",
+        to_stdin = true,
         format = "json",
-        args = { "--no-exit", "--output=JSON", "$FILENAME" },
+        args = function(params)
+            return { "--no-exit", "--output", "JSON", "--ext", "." .. vim.fn.fnamemodify(params.bufname, ":e") }
+        end,
         on_output = function(params)
             local diagnostics = {}
             local severities = { error = 1, warning = 2, suggestion = 4 }
-            for _, diagnostic in ipairs(params.output[params.bufname]) do
+            for _, diagnostic in ipairs(params.output["stdin." .. vim.fn.fnamemodify(params.bufname, ":e")]) do
                 table.insert(diagnostics, {
                     row = diagnostic.Line,
                     col = diagnostic.Span[1],
```

But if it works as expected for you, there's no need to change it.

carrascomj:
Yes, that works beautifully. Vale also accepts passing it text so it could be done line by line, but this way looks performant enough I think.

jose-elias-alvarez:
Got it – let's see if this is okay for now and modify if necessary later.

PR #79, opened 2021-08-03
Add support formatting for `fennel` lang

babygau:
It will format `*.fnl` file using [`fnlfmt`](https://git.sr.ht/~technomancy/fnlfmt)

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -254,6 +254,18 @@ local sources = {null_ls.builtins.formatting.fish_indent}
 - Command: `fish_indent`
 - Arguments: `{}`
 
+#### [fnlfmt](https://git.sr.ht/~technomancy/fnlfmt)
+
+```lua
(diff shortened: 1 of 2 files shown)


vuki656:
It would be nice if you could also update the docs [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md).

jose-elias-alvarez:
Thanks for this! And yep, I'd appreciate it if you could add a section to the documentation so other users are able to know we've added support.

babygau:
> Thanks for this! And yep, I'd appreciate it if you could add a section to the documentation so other users are able to know we've added support. (Edit: also a good reminder that we should probably put together a quick contribution guide.) 

Sure, I'll update the doc later on tonight 😀👍

babygau:
@jose-elias-alvarez, I updated the docs as your suggestion. Thank you!

jose-elias-alvarez:
Thank you!

Issue #74, opened 2021-08-01
Add luacheck


ChristianChiarulli:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Add luacheck as a linter

### Help

No

### Implementation help

If I find some time I will add this, but if someone wants to beat me to it, I would appreciate it.

vuki656:
On the off note, isn't luacheck pretty much dead? The last commit was in 2018.

abzcoding:
> On the off note, isn't luacheck pretty much dead? The last commit was in 2018.

yeah people use `selene` nowadays but it is a good linter in its own right :sweat_smile:

vuki656:
Oh cool, will look into adding it to null-ls.

abzcoding:
> Oh cool, will look into adding it to null-ls.

it is already supported by null-ls :cold_sweat:

vuki656:
True, but nothing in the BUILTINS.md, so I guess I should add that instead :smile:

jose-elias-alvarez:
Good point – the `selene` built-in was part of a much larger commit, so we overlooked it.

PR #80, opened 2021-08-03
add stylua on PRs, clean up BUILTINS.md

### LINKED ISSUE #72: [DOCS] Unclear setup instructions in BUILTINS.md
Hey, in the docs for builtins, for every source the doc states that config is set up like so 

- Filetypes: `{ "asm" }`
- Command: `asmfmt`
- Arguments: `{}`

While configuring mine I assumed the param names would be `filetypes`, `command`, and `arguments`. While this is true for `filetypes`, and `command`, arguments actually take `args` param instead of `arguments`. Lost a few hours trying to figure it out a few days ago. 

More to the point, I doesn't state anywhere in `BUILTINS.md` that you should use `args` and not `arguments`.

I also found that in #35, there was the same problem. I assume I'm not the only person that is going to make that mistake. 

What do you think about the examples being written in the following format: 

```lua 
{
    command = 'prettier',
    filetypes = { 'css', 'html' },
    args = { '--no-semi', '--tab-width=4' }
}
```

That way people can just copy-paste and not have to figure out parameter names.

What do you think? 

I'd be happy to open a PR.

vuki656:
As discussed in #72 I cleaned up docs a bit.

- Added `selene` as it was [missing](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/74#issuecomment-891037232).
- Fixed some missing usage fields
- Fixed formatting in accordance to `markdown-lint`

May I also recommend integration of `markdown-lint` into the pipeline to run on PRs since people are often contributing to the docs as well and that would ensure consistency. I'd be happy to open a PR if you agree.   

closes #72

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -1,0 +1,14 @@
+name: default
+
+on: [pull_request]
+
+jobs:
+  stylua:
+    name: Check codestyle
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: JohnnyMorganz/stylua-action@1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --color always --check .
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Awesome, thank you for this! Using the correct keys for the defaults should reduce confusion.

PR #82, opened 2021-08-04
feat: add svelte in builtins format

zack9433:
support svelte in prettier and ESLint

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -320,7 +320,7 @@ local sources = { null_ls.builtins.formatting.eslint }
 
 ##### Defaults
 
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
Seems fine to me, could you also update the [built-ins documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md) with the new filetype? Also, if Svelte works with ESLint, we can maybe add it as a default for the diagnostics source, too.

zack9433:
updated

jose-elias-alvarez:
Thanks!

Issue #85, opened 2021-08-05
Builtin clang-format formatter not working because of missing filename argument


PasiBergman:
### FAQ

- [X] I have checked [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

0.5

### Steps to reproduce

Open a file with filetype "c", "cpp", "cs" or "java" and try to use builtin clang_format (clang-format) to format the current buffer. Buffer does not get formatted, because the `-assume-filename` option value does not get set.

### Expected behavior

Buffer content gets formatted using `clang-format`

### Actual behavior

Buffer content does not get formatted.

### Debug log

error: empty filenames are not allowed
Error: clang-format exited with exit code 1.

### Help

Yes

jose-elias-alvarez:
Closed by #86.

PR #86, opened 2021-08-05
fix: clang-format filename argument

PasiBergman:
set value to clang_format (clang-format) filename argument

Refs #85


jose-elias-alvarez:
Thank you!

PR #92, opened 2021-08-08
Add stylua range formatting.

mks1508:
NOTE: keymap setting
```vim
" It doesn't work.
<cmd>lua vim.lsp.buf.range_formatting()<CR>

" It works
:lua vim.lsp.buf.range_formatting()<CR>
```

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -769,7 +769,8 @@ local sources = { null_ls.builtins.formatting.shfmt }
 
 ##### About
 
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Works well, thank you! Could you also add a note to the [built-ins documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md) that `stylua` now supports range formatting, similar to what we wrote for Prettier?

jose-elias-alvarez:
Thanks!

PR #101, opened 2021-08-11
fix typo

beauwilliams:
quick one just because I noticed it


jose-elias-alvarez:
Thank you! Should make it easier to search through the document.

PR #106, opened 2021-08-13
Extend buitin args with user provided ones

### LINKED ISSUE #105: Builtin args are not extendable
Example:

``` lua
local luacheck = null_ls.builtins.diagnostics.luacheck
luacheck = luacheck.with({args = { myargs }})
```

Expected:

``` lua
luacheck.args = { myargs + null_ls defined args }
```

Actual:

``` lua
luacheck.args = { myargs }
```

That's unfortunately the behavior of `vim.tbl_deep_extend`, for a list, it will overwrite it instead of extending it.

So then I thought about doing

``` lua
local luacheck = null_ls.builtins.diagnostics.luacheck
vim.list_extend(myargs, luacheck._opts.args)
luacheck = luacheck.with({args = { myargs }})
```

But then it didn't work for [other builtins](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting.lua#L372)
Some of them have a table as args, others have a function, this would need to be streamlined.

Tastyep:
Fixes #105

### DIFF
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -260,13 +260,13 @@ end
 
 M.make_builtin = function(opts)
     local method, filetypes, factory, generator_opts, generator =
-        opts.method, opts.filetypes, opts.factory, opts.generator_opts, opts.generator
+        opts.method, opts.filetypes, opts.factory, opts.generator_opts or {}, opts.generator
 
     local builtin = {
         method = method,
         filetypes = filetypes,
         generator = generator,
-        _opts = generator_opts or {},
+        _opts = generator_opts,
         name = opts.name,
     }
 
@@ -282,14 +282,11 @@ M.make_builtin = function(opts)
         -- Extend args manually as vim.tbl_deep_extend overwrites the list
         if user_opts.extra_args and vim.tbl_count(user_opts.extra_args) > 0 then
             local builtin_args = builtin._opts.args
-            local user_args = user_opts.extra_args
+            local extra_args = user_opts.extra_args
 
             builtin._opts.args = function(params)
-                local user_args_cpy = user_args
-                return vim.list_extend(
-                    user_args_cpy,
-                    type(builtin_args) == "function" and builtin_args(params) or builtin_args
-                )
+                local builtin_args_cpy = type(builtin_args) == "function" and builtin_args(params) or builtin_args
+                return vim.list_extend(builtin_args_cpy, extra_args)
             end
             user_opts.extra_args = nil
         end
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -266,7 +266,7 @@ M.make_builtin = function(opts)
         method = method,
         filetypes = filetypes,
         generator = generator,
-        _opts = generator_opts,
+        _opts = vim.deepcopy(generator_opts),
         name = opts.name,
     }
 
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -280,17 +280,18 @@ M.make_builtin = function(opts)
         builtin.filetypes = user_opts.filetypes or builtin.filetypes
 
         -- Extend args manually as vim.tbl_deep_extend overwrites the list
-        if user_opts.args then
+        if user_opts.extra_args and vim.tbl_count(user_opts.extra_args) > 0 then
             local builtin_args = builtin._opts.args
-            local user_args = user_opts.args
+            local user_args = user_opts.extra_args
 
-            user_opts.args = nil
             builtin._opts.args = function(params)
+                local user_args_cpy = user_args
                 return vim.list_extend(
-                    user_args,
+                    user_args_cpy,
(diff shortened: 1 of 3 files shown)


lua/null-ls/helpers.lua
```diff
@@ -278,6 +278,21 @@ M.make_builtin = function(opts)
 
     builtin.with = function(user_opts)
         builtin.filetypes = user_opts.filetypes or builtin.filetypes
+
+        -- Extend args manually as vim.tbl_deep_extend overwrites the list
+        if user_opts.args then
```
jose-elias-alvarez, line 283: Could we also make this depend on `vim.tbl_count(user_opts.args) > 0`? I think it would be convenient for integrations that want to set arguments conditionally.

jose-elias-alvarez (changes requested):

jose-elias-alvarez:
I left a couple of comments, and as I mentioned earlier, I also think we should make this a separate key, e.g. `extra_args`. I don't want to break existing configurations, and I also think it's inconsistent with expected behavior. For example, with the following:

```lua
null_ls.builtins.formatting.prettier.with({
    filetypes = { "html", "json", "yaml", "markdown" },
})
```

I'd expect `filetypes` to override the defaults, not extend them. 

In the documentation, we should describe `extra_args` as the preferred method to modify arguments, since (as you were saying) the goal is that the user should not want to override the defaults, but I'd like to keep the current `args` behavior for the reasons above. (If you want this to be the default behavior in LunarVim, it seems pretty simple to assign user-defined `args` to the `extra_args` key.)

Tastyep:
So, I realized that there is an issue pushing the extended args front, for instance with linters using subcommand, ex: `tl check`.

So I tried pushing them back, and almost succeeded, but it requires making a copy of arguments because they are then modified afterward, and then I must check for the type of `args` to see if it's a table or a function.
I feel like I end up with a clumsy solution and hardly maintainable code.

I would suggest defining builtin as a class (table) in a separate file and provide methods to update the attributes (or even make some of them public like the filetype).
To avoid breaking the API, a with function could be provided, which would dispatch to the appropriate setters for each entry.

Tell me what you think.

jose-elias-alvarez:
I'm not opposed to re-writing built-ins as a class, but I also think it's a big change given that the functionality is (as far as I know) handling the use cases we have seen so far. What do you think of something like this? It allows setting where to insert the arguments but keeps the code a bit more readable and avoids the mutation issue. (`extra_args_at` is just a sketch and could definitely be improved.)

```diff
diff --git a/lua/null-ls/helpers.lua b/lua/null-ls/helpers.lua
index e275d7e..e3978ab 100644
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -282,14 +282,19 @@ M.make_builtin = function(opts)
         -- Extend args manually as vim.tbl_deep_extend overwrites the list
         if user_opts.extra_args and vim.tbl_count(user_opts.extra_args) > 0 then
             local builtin_args = builtin._opts.args
-            local user_args = user_opts.extra_args
+            local extra_args = user_opts.extra_args
+            local extra_args_at = user_opts.extra_args_at or "start"

             builtin._opts.args = function(params)
-                local user_args_cpy = user_args
-                return vim.list_extend(
-                    user_args_cpy,
-                    type(builtin_args) == "function" and builtin_args(params) or builtin_args
-                )
+                local original_args = type(builtin_args) == "function" and builtin_args(params) or builtin_args
+                local combined_args = {}
+
+                local first = extra_args_at == "start" and extra_args or original_args
+                local second = extra_args_at == "start" and original_args or extra_args
+
+                vim.list_extend(combined_args, first)
+                vim.list_extend(combined_args, second)
+                return combined_args
             end
             user_opts.extra_args = nil
         end
```

If the user / an integration needs more customization than this, I think the right solution is to override the arguments or to just create a new source. I don't think built-ins need to handle every conceivable case.

Tastyep:
I think it's a source of new errors to introduce this `extra_args_at` entry, the user will never know how to use it, as the usage is not transparent (if that means anything).

Also, I forgot to mention it in my message, but what made pushing extra_args back difficult is that we need to handle the `source` argument when it's set to `"-"`. I checked, it needs to be last.

I'll give the class idea a try and if we see we don't like it, we will find another way.

jose-elias-alvarez:
In that case, why not just always insert `extra_args` at the front and rely on overriding the arguments for other cases? If that handles 95% of cases and users know how to handle the other 5%, that seems good enough.

I just don't know if I see the justification to add that much complexity to something that is fundamentally simple. `with` is basically just a wrapper for manually overriding `builtin._opts`, with a little special handling for filetypes and conditions.

Tastyep:
So, I ended up always pushing back, I didn't `with` to only work 95% of the cases.
If `"-"` is detected, then it's pushed back to extra_args, this way it stays last.
I also deepcopied some tables so that user-provided parameters are not modified by the function since table assignments work like pointer assignment in C.

jose-elias-alvarez:
Perfect! I think this is ideal. Thank you for working on this and for your flexibility about the implementation!

PR #109, opened 2021-08-14
[Bugfix] Add missing check for nil args + tests

Tastyep:
I overlooked this possibility :)


jose-elias-alvarez:
Thanks for catching this!

PR #113, opened 2021-08-15
Remove opinionated builtin parameters

Tastyep:
Follow up on #105 
I just kept "--search-parent-directories"

Search parent directories for stylua.toml, if not found in current directory.
Ignored if config_path is provided. Keeps searching recursively up the parent
directory tree, until the root directory is reached. If not found, looks in
$XDG_CONFIG_HOME or $XDG_CONFIG_HOME/stylua

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -14,8 +14,6 @@ M.chktex = h.make_builtin({
         args = {
             -- Disable printing version information to stderr
             "-q",
-            -- Disable executing \input statements
-            "-I0",
             -- Format output
             "-f%l:%c:%d:%k:%m\n",
         },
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Thank you for this! I added a note to the contribution guidelines too and will try to be on the lookout for this when PRs get submitted.

Tastyep:
Did you fix the formatting issue ?

jose-elias-alvarez:
Nope! I'm still not used to having the check and ignored it completely. I'll format and commit right now.

kylo252:
I have an idea but it might be too much work to do it retroactively. It should be that those tools are reading the project and/or user configurations by default. If that's not the case, then the required flag should be set.

I'm not sure how often this can happen, and I'm hoping it never does, but a semi-related example is `clangd` that needed an explicit `--enable-config`, at least until recently.

Tastyep:
Correct me if my vision is wrong, but null-ls goal is to run builtins, not configure them.
However, smth that could be done on our side is detecting the builtin configuration and run it conditionally based on whether it is present or not

PR #115, opened 2021-08-15
chore(stylua): add --stdin-filepath $FILENAME

zeertzjq:
Since `--search-parent-directories` is added, maybe `--stdin-filepath` `$FILENAME` should be added to search the correct parent directories.


jose-elias-alvarez:
Makes sense, thank you!

PR #130, opened 2021-08-24
Fix eslint formatter argument

polyzen:
Fixes `Error: Cannot find module '/usr/lib/node_modules/eslint/lib/cli-engine/formatters/JSON'`.


jose-elias-alvarez:
Makes sense, thank you!

PR #133, opened 2021-08-25
Add fixjson and codespell to builtins

### LINKED ISSUE #123: Codespell integration
### Issues

### Feature description

Would it be possible to implement [codespell](https://github.com/codespell-project/codespell)?

### Help

### Implementation help

What would codespell need to provide in order to be compatible?

Metallum19:
Fixes #123
Codespell may need more work.


jose-elias-alvarez:
Thanks for working on this! I wonder if for `codespell`, it would be better to use `stdin`, since that'll allow the user to get updated diagnostics without writing the file:

```diff
diff --git a/lua/null-ls/builtins/diagnostics.lua b/lua/null-ls/builtins/diagnostics.lua
index a43517b..d63c7e0 100644
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -400,7 +400,8 @@ M.codespell = h.make_builtin({
     filetypes = { "*" },
     generator_opts = {
         command = "codespell",
-        args = { "$FILENAME" },
+        args = { "-" },
+        to_stdin = true,
         to_stderr = true,
         format = "line",
         on_output = h.diagnostics.from_pattern([[:(%d+): [%w-/]+ (.*)]], { "row", "message" }, {
```

The issue is that doing this seems to cause `codespell` to split up each diagnostic into two lines. There is an [old PR](https://github.com/codespell-project/codespell/pull/726) to improve its `stdin` behavior, but it seems to have stalled. I see that [nvim-lint](https://github.com/mfussenegger/nvim-lint/blob/f9446c0cfef5e199f8b61c99bff235da35c1b665/lua/lint/linters/codespell.lua) doesn't use `stdin` here, either. 

Unless there's a way to force it to output to one line only, the solution may be to write a custom `on_output` handler to combine multiple lines into a single diagnostic.

Metallum19:
Thank you for the review.
Yes, that's why I used `stderr` to at least show the error message and then work from there.
My question is, if you disable `stdin` and `stderr`, if there is a way to get the result of codespell execution and then format this output.
I read about the `params` object but need to see it in more detail.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,62 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
 local M = {}
 
+local function mysplit(inputstr, sep)
```
jose-elias-alvarez, line 8: The Neovim standard library has a function, `vim.split`, that does the same thing.

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,62 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
...
+    if sep == nil then
+        sep = "%s"
+    end
+    local t = {}
+    for str in string.gmatch(inputstr, "([^" .. sep .. "]+)") do
+        table.insert(t, str)
+    end
+    return t
+end
+local handle_output = function(params)
```
jose-elias-alvarez, line 18: If we aren't going to re-use this for multiple sources, could we inline this into the source itself?

lua/null-ls/builtins/diagnostics.lua
```diff
@@ -5,6 +5,62 @@ local DIAGNOSTICS = methods.internal.DIAGNOSTICS
...
+        -- for _, _ in ipairs(table_output) do
+        --     length = length + 1
+        -- end
+        -- print(length)
+
+        -- for l = length, 1, -2 do
+        --     print(table_output[l - 1] .. ': ' .. table_output[l]:gsub("\t", ""))
+        -- end
+
+        return {
```
jose-elias-alvarez, line 40: I made a mistake describing how this should work. Since built-in sources are asynchronous, `handle_output` receives a second argument, `done`, which needs to be called with the results:

```diff
diff --git a/lua/null-ls/builtins/diagnostics.lua b/lua/null-ls/builtins/diagnostics.lua
index 84da6ee..a6f8d16 100644
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -15,7 +15,7 @@ local function mysplit(inputstr, sep)
     end
     return t
 end
-local handle_output = function(params)
+local handle_output = function(params, done)
     local output_error = params.err
     if output_error ~= nil then
         local split_table = mysplit(output_error, '\n')
@@ -37,14 +37,14 @@ local handle_output = function(params)
         --     print(table_output[l - 1] .. ': ' .. table_output[l]:gsub("\t", ""))
         -- end

-        return {
+        return done({
             {
                 row = table_output[1],
                 source = 'codespell',
                 message = table_output[2],
                 severity = 2,
             }
-        }
+        })
     end
 end
```

jose-elias-alvarez (commented):
Sorry for the faulty description! Using `done`, it works so far though at the moment it seems like it'll only handle the first diagnostic.

Metallum19:
Thanks for the review.
I have already finished fixing the recommendations you made.
The function now handles multiple errors.

Metallum19:
I just notice one problem.
It seem that somewhere, the last value of the returned diagnostics table is stored, when the misspell warning is fixed it just keeps showing the warning.
I tried starting the local variables with `nil` but has no effect.
After fixing the last warning I got this, I guess it has to do with `output: []` not being `output: nil`

2021-08-29-185337_666x732_scrot

jose-elias-alvarez:
Right, we need to explicitly call `done` if `codespell` output is empty. I got it to work with this diff:

```diff
diff --git a/lua/null-ls/builtins/diagnostics.lua b/lua/null-ls/builtins/diagnostics.lua
index ca78741..3ad2477 100644
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -402,31 +402,34 @@ M.codespell = h.make_builtin({
         command = "codespell",
         args = { "-" },
         to_stdin = true,
-        format = "raw",
+        to_stderr = true,
         on_output = function(params, done)
-            if params.err then
-                local diagnostics = {}
-                local final_table = {}
-                local raw_table = vim.split(params.err, "\n")
+            local output = params.output
+            if not output then
+                return done()
+            end

-                for _, str in ipairs(raw_table) do
-                    local tmp_table = vim.split(str, ":")
-                    tmp_table[1] = string.gsub(tmp_table[1], "\t", "")
-                    if tmp_table[1] ~= "" then
-                        table.insert(final_table, tmp_table[1])
-                    end
-                end
+            local diagnostics = {}
+            local final_table = {}
+            local raw_table = vim.split(output, "\n")

-                for l = #final_table, 1, -2 do
-                    table.insert(diagnostics, {
-                        row = final_table[l - 1],
-                        source = "codespell",
-                        message = final_table[l],
-                        severity = 2,
-                    })
+            for _, str in ipairs(raw_table) do
+                local tmp_table = vim.split(str, ":")
+                tmp_table[1] = string.gsub(tmp_table[1], "\t", "")
+                if tmp_table[1] ~= "" then
+                    table.insert(final_table, tmp_table[1])
                 end
-                return done(diagnostics)
             end
+
+            for l = #final_table, 1, -2 do
+                table.insert(diagnostics, {
+                    row = final_table[l - 1],
+                    source = "codespell",
+                    message = final_table[l],
+                    severity = 2,
+                })
+            end
+            return done(diagnostics)
         end,
     },
     factory = h.generator_factory,
```

Metallum19:
Yep, it makes sense.
I didn't notice any other things to improve, I think it's finished.

jose-elias-alvarez:
Yep, looks good to me. Thanks for working on this!

PR #136, opened 2021-08-28
perf: don't update diagnostics in insert mode

### LINKED ISSUE #111: Insert mode lag?
I've seen a few issues on LunarVim's repo that seem to be related to insert mode lag from null-ls diagnostics, such as [this one](https://github.com/LunarVim/LunarVim/issues/1223), which seems to have been solved by disabling Lua linting. On my end (using an underpowered Mac Mini currently), I haven't experienced any lag, even with slow sources like ESLint, and we haven't had any issues reported here. 

I did notice that attempting to spawn a command that didn't exist / wasn't on the user's `$PATH` caused lag, so in 6acbccff4214988aa6855b0fd9c24ad4397f6c62, I added a simple check to make sure the command is executable and also made it so the generator wouldn't run again if it threw an error. I wonder if these issues are related, but from what I understand, diagnostics have to be explicitly enabled in LunarVim, right?

Since I can't replicate these issues, I'd like to know more and see if they're still occurring, and I'm very open to any suggestions / fixes to improve performance. We're currently using the `debounce_text_changes` flag, which makes a big difference, but one idea is to simply avoid running generators in insert mode and send a synthetic `didChange` even when the user exits insert mode. 

cc @Tastyep @kylo252 @ other LunarVim people

jose-elias-alvarez:
Closes #111. 

This is working as expected (at least on my end), but to be honest I'm not thrilled with it, mostly because I feel that we're straying too far from LSP behavior here. I wonder if this would be better as a change / extension to Neovim itself, since I think the current behavior of the `update_in_insert` flag doesn't really match user expectations. 

Also, from what I understand, this is fundamentally not a null-ls issue but rather with slow linters, and it seems that other solutions like efm also suffer from [the same issue](https://github.com/mattn/efm-langserver/issues/98). Having users of problematic linters raise the value of `debounce` in their configuration is another potential solution (and one that's built-in to Neovim already). 

Finally, as I mentioned in the issue, I haven't personally experienced lag with any source on my (modest) setup, so I have no idea if this actually solves anything. If all we're doing is delaying lag until after the user exits insert mode, I don't know if this change is worth the increase in complexity. 

cc @abzcoding (since it seems like you've dealt with related issues in LunarVim)

Edit: also, yet another reason I'm lukewarm about this is that diagnostics won't immediately update / show when exiting insert mode, since that's when generators will actually run. That's probably an acceptable trade-off if it does solve the issue, though.

### DIFF
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -4,6 +4,58 @@ local c = require("null-ls.config")
 local methods = require("null-ls.methods")
 local generators = require("null-ls.generators")
 
+local api = vim.api
+
+local get_augroup_name = function(bufnr)
+    return string.format("NullLsInsertLeave%s", bufnr)
+end
+
+local tracked_buffers = {}
+
+local init_tracking = function(uri)
+    local bufnr = vim.uri_to_bufnr(uri)
+    api.nvim_exec(
+        string.format(
+            [[
+        augroup %s
+            autocmd!
+            autocmd InsertLeave <buffer=%s> lua require("null-ls.diagnostics").run()
+        augroup END
+            ]],
+            get_augroup_name(bufnr),
+            bufnr
+        ),
+        false
+    )
+
+    tracked_buffers[uri] = {}
+end
+
+local clear_tracking = function(uri)
+    local bufnr = vim.uri_to_bufnr(uri)
+    api.nvim_exec(
+        string.format(
+            [[
+        augroup %s
+            autocmd!
+        augroup END
+            ]],
+            get_augroup_name(bufnr)
+        ),
+        false
+    )
+
+    tracked_buffers[uri] = nil
+end
+
+local handle_tracked_change = function(uri, params)
+    if not tracked_buffers[uri] then
+        return
+    end
+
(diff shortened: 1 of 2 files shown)


abzcoding:
thank you,
cc: @Tastyep @kylo252

jose-elias-alvarez:
Update: I'm not very happy with this and think that the implementation may actually cause performance issues by running generators multiple times if the user makes a change and immediately exits insert mode (a common situation). 

I'm happy to consider alternatives but I don't think I'll be merging this as-is. Raising the value of `debounce` from the default `250` to `1000` significantly reduces how often generators run in insert mode without any of the additional complexity added here, so that's something I'd recommend for users with slower setups.

abzcoding:
We can probably document that somewhere and let users know that. if you don't feel comfortable merging this, please don't :)

jose-elias-alvarez:
@abzcoding I appreciate the understanding! To be 100% clear, I'm definitely open to solutions to fix the issues users seem to be having. It just seems that the approach I tried in this PR (updating only in insert mode) is a dead end, so I'm going to close this but keep the issue open.

PR #142, opened 2021-08-31
Update 'codespell' docs

LongerHV:
Update 'codespell' docs to reflect actual default arguments.


jose-elias-alvarez:
Thank you for catching this!

PR #145, opened 2021-09-02
refactor: switch to plenary jobs

### LINKED ISSUE #140: Switch to Plenary jobs to spawn processes?
The current implementation of `loop.spawn` is working well, but it's caused mysterious issues in the past (e.g. #84), and offloading the task to Plenary would reduce the maintenance burden here. We'd also benefit from future improvements to Plenary, like async jobs. 

The `loop` module is pretty isolated, so this should be a pretty straightforward change. I want to investigate how other plugins are using Plenary jobs and make sure that they have all the features we need (timeouts are the only thing I'm unsure about).

jose-elias-alvarez:
Closes #140 (if merged).

This knocks out ~50 lines of code from `loop.lua`. The main advantages are simplified pipe handling and the ability to register callbacks on exit, which make timeouts cleaner.  In the future, this change would let us take advantage of future Plenary features (e.g. async jobs) and potentially handle more complex cases, such as piping the output of one command into another. 

The biggest disadvantage is that Plenary splits the output of each command into lines to pass them to `on_output`, but we'll end up combining them anyways. It's not a big deal, since it's just a single `table.concat`, but it's also not ideal. The current `on_output` implementation frankly confuses me, but in the future I can see if I can put together a PR to allow jobs to call `on_output` with a command's full output. 

Overall, I'm pretty happy with the change but want to think about it a bit more before merging. I also have to fix unit tests.

### DIFF
--- a/lua/null-ls/builtins/test.lua
+++ b/lua/null-ls/builtins/test.lua
@@ -61,31 +61,6 @@ M.mock_code_action = {
     filetypes = { "lua" },
 }
 
-M.slow_code_action = h.make_builtin({
-    method = methods.internal.CODE_ACTION,
-    filetypes = { "lua" },
-    generator_opts = {
-        command = "bash",
-        args = { "./test/scripts/sleep-and-echo.sh" },
-        timeout = 100,
-        on_output = function(params, done)
-            if not params.output then
-                return done()
-            end
-
-            return done({
-                {
-                    title = "Slow mock action",
-                    action = function()
-                        print("I took too long!")
-                    end,
-                },
-            })
-        end,
-    },
-    factory = h.generator_factory,
-})
-
 M.cached_code_action = h.make_builtin({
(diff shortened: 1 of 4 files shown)


jose-elias-alvarez:
Based on [this PR](https://github.com/nvim-lua/plenary.nvim/pull/101) (which became [this other PR](https://github.com/nvim-lua/plenary.nvim/pull/113) and has an unclear relationship to [this other other PR](https://github.com/nvim-lua/plenary.nvim/pull/183/files)), it seems that it may be better to wait until async jobs are merged, since it seems that that'll bring significant improvements and make handling output simpler.

jose-elias-alvarez:
While testing this out, I found some bugs that don't exist with the current implementation (e.g. diagnostics failing to update after changes). I'm going to close this PR and revisit once async jobs are merged.

Issue #149, opened 2021-09-03
pylint is giving columns with 0 offset instead of 1


jameshiew:
### FAQ

- [X] I have checked [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.5.0

### Steps to reproduce

Use the built in `pylint` functionality (`null_ls.builtins.diagnostics.pylint`) on a Python file

### Expected behavior

Diagnostic column values start from `1`

### Actual behavior

For `pylint`, diagnostics have column values starting from 0. Some other plugins that use LSP diagnostic messages may error as they are expecting columns to start from 1 (e.g. [trouble.nvim](https://github.com/folke/trouble.nvim))

### Debug log

N/A

### Help

Yes

jameshiew:
Example - pylint and pyright give different locations for the equivalent diagnostic (variable `x` is unused)

![20210903_15h14m34s_grim](https://user-images.githubusercontent.com/2990876/132020067-0a70cbcc-4522-4a79-a744-82af61644f16.png)

PR #150, opened 2021-09-03
fix(builtin): pylint column off by one

jameshiew:
https://github.com/jose-elias-alvarez/null-ls.nvim/issues/149

Similarly to https://github.com/dwb/neovim/commit/9794f6caef063ba1283bb728ac3befda477935f3


jose-elias-alvarez:
Thanks for putting this together!

PR #152, opened 2021-09-03
Add `$ROOT` substitution arg for LSP workspace root

### LINKED ISSUE #151: Support passing detected LSP project root to external command
### Issues

### Feature description

I am currently in the process of migrating my configuration from EFM to null-ls. 

Some formatting providers need to know the project root to work correctly. For example `isort` needs it do distinguish between local and external imports. EFM has the variable `$ROOT` for this:
https://github.com/disrupted/dotfiles/blob/3efca0ac29500410cc34f6e85cd98c2df1a940a5/.config/nvim/lua/efm/isortd.lua#L4

It would be useful to have the same `$ROOT` arg available in null-ls in addition to the existing ones (`$FILENAME`, `$FILEEXT`, `$TEXT`).

### Help

Yes

### Implementation help

_No response_

disrupted:
Closes #151

This adds a new substitution argument `$ROOT` for the LSP workspace root path.

### DIFF
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -37,6 +37,9 @@ local parse_args = function(args, params)
         if string.find(arg, "$FILEEXT") then
             arg = u.string.replace(arg, "$FILEEXT", vim.fn.fnamemodify(params.bufname, ":e"))
         end
+        if string.find(arg, "$ROOT") then
+            arg = u.string.replace(arg, "$ROOT", vim.lsp.get_client_by_id(params.client_id).config.root_dir)
+        end
 
         table.insert(parsed, arg)
     end
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -38,7 +38,9 @@ local parse_args = function(args, params)
             arg = u.string.replace(arg, "$FILEEXT", vim.fn.fnamemodify(params.bufname, ":e"))
         end
         if string.find(arg, "$ROOT") then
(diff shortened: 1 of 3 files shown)


lua/null-ls/helpers.lua
```diff
@@ -37,6 +37,9 @@ local parse_args = function(args, params)
         if string.find(arg, "$FILEEXT") then
             arg = u.string.replace(arg, "$FILEEXT", vim.fn.fnamemodify(params.bufname, ":e"))
         end
+        if string.find(arg, "$ROOT") then
+            arg = u.string.replace(arg, "$ROOT", vim.lsp.get_client_by_id(params.client_id).config.root_dir)
```
jose-elias-alvarez, line 41: I think we should guard against the possibility that `client` is `nil` and use `vim.fn.getcwd` as a fallback (similar to what we are doing on line 210 of this file).

jose-elias-alvarez (changes requested):
Thank you for this! Could you also add test coverage for the `$ROOT` case? If you stub `vim.lsp.get_client_by_id`, it should be pretty simple, but let me know if you want help.

disrupted:
> Thank you for this! Could you also add test coverage for the `$ROOT` case? If you stub `vim.lsp.get_client_by_id`, it should be pretty simple, but let me know if you want help.

I added the tests now. let me know if this is how you intended it.

jose-elias-alvarez (approved):

jose-elias-alvarez:
Perfect, thank you!

PR #154, opened 2021-09-04
Support rubocop formatting for ruby

luong-komorebi:
reference pull request that works in nvim formatter https://github.com/mhartington/formatter.nvim/pull/65

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -801,6 +801,26 @@ local sources = { null_ls.builtins.formatting.format_r }
 - `command = "R"`
 - `args = { "--slave", "--no-restore", "--no-save", "-e", 'formatR::tidy_source(source="stdin")' }`
 
+#### [rubocop](https://github.com/ruby-formatter/rubocop)
+
+##### About
+
+Ruby static code analyzer and formatter, based on the community Ruby style guide.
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Thank you! The arguments are pretty strange but I see that Neoformat [does something similar](https://github.com/sbdchd/neoformat/blob/a75d96054618c47fbafef964d4d705525e8e37b9/autoload/neoformat/formatters/ruby.vim). 

Could you please squash your commits and also add a note to the documentation to indicate that `awk` is required for the formatter to work? I don't know too much about Windows but the arguments seem very Unix-specific.

luong-komorebi:
@jose-elias-alvarez you are totally true that the script needs awk to work. 
The method that you quote on neoformat is pretty popular. However, you can notice that someone already pointed out an edge case in my pull request at formatter.nvim and it also requires another tool, `sed`.

A much more complete solution would be parsing the text with lua like what Ale did with vimscript https://github.com/dense-analysis/ale/blob/master/autoload/ale/fixers/rubocop.vim#L8-L20, but I am not fluent enough in Lua to achieve that

jose-elias-alvarez:
Thanks! We can merge this as-is and use a Lua parser in the future if it becomes an issue.

Issue #139, opened 2021-08-29
Add support for mypy


Funk66:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Mypy is conspicuously absent from the list of supported diagnostic tools. I was wondering whether there is a reason for that. It is supported by [nvim-lint](https://github.com/mfussenegger/nvim-lint/blob/c086009aaa6e73c0cf39432997790b317a875fca/lua/lint/linters/mypy.lua) and its implementation doesn't seem to be particularly troublesome.
Mypy has open issues to add support for [json formatting](https://github.com/python/mypy/issues/10816) and a full-fledged [language server](https://github.com/python/mypy/issues/10463), although I doubt that these are going to be implemented anytime soon.

### Help

Yes

### Implementation help

I can try to implement it myself, but I may need some help to make it work properly.

jose-elias-alvarez:
I'm not familiar with `mypy`, but based on [this issue](https://github.com/python/mypy/issues/2119) it seems like it'll need some work to support linting a buffer's content (versus linting a file on disk). At the moment, all of the built-in diagnostic sources support linting on change, so I'd be hesitant to introduce a source that doesn't support it.

It seems that the [ALE source](https://github.com/dense-analysis/ale/blob/7d8fb2ba1716a744446b811fc278ecf30d4eb771/ale_linters/python/mypy.vim) has solved this issue, so we should probably use that as the base for the implementation (plus the patterns from nvim-lint). I'll have time to work on this myself in the second half of September but can review a PR before then.

Funk66:
Oh, bugger. I thought I had an almost-working prototype, but you're right - this isn't as straightforward as I initially thought. I'll see if I can still open a PR in the coming weeks.
You may already know this, but for completeness I'll mention that mypy can run as a [daemon](https://mypy.readthedocs.io/en/stable/mypy_daemon.html). I'm not familiar with the internal workings of null-ls, but I assume that running a background process and interacting with it is going to be more complex.
Also, to provide additional context, mypy is the "official" static type checker for Python and pyright is - in my experience - not quite mature enough to replace it.

jose-elias-alvarez:
From the documentation, it seems that the daemon might work well, since (if I understand correctly) it seems that we can interact with it as if it were a normal CLI program. I'd definitely give that a shot.

PR #155, opened 2021-09-04
Add mypy diagnostic builtin for Python.

dsully:


jose-elias-alvarez:
Can you verify whether this updates diagnostics when the buffer's content changes (before saving the file to the disk)? #139 has some discussion about how `mypy` doesn't seem to have proper `stdin` support and may require additional work (changing the arguments or using a temp file).

dsully:
You're right - but it looks like we can use mypy's `--shadow-file` option if `null-ls` provides access to the tempfile.

I see you already referenced https://github.com/python/mypy/issues/2119 which talks about this.

jose-elias-alvarez:
null-ls built-ins have access to a `to_temp_file` flag, which will write a file to a temporary directory (`/tmp` on Unix). We can try that out with `--shadow-file`, though from what I understand, `mypy` needs to scan the whole directory for accurate type checking, so we'd need to make sure that the arguments make `mypy` aware of where the actual file is.

WhyNotHugo:
`to_temp_file` seems to fall a bit short -- `mypy` needs to be passed two paths:

- The location of the _actual_ file, to resolve relative imports, etc.
- The location of the `--shadow-file`, which is, in this case, the temporary file.

However, when using `to_temp_file`, the original path is no longer available.

jose-elias-alvarez:
I made a small change in fce6e29c589ed4e1d617d4fdd80af8e49bf5281a so that `args` functions have access to both paths:

```lua
args = function(params)
    print(params.bufname) -- original
    print(params.temp_path) -- temp path
    return { "actual", "args", "here" }
end
```

Let me know if that makes it possible to handle this, or if further changes are needed.

nawordar:
I wanted to try to finish this feature, but I stumbled across a different problem: `mypy` exits with error code 2 when there is almost any kind of error — including invalid command arguments and syntax error.

If I make `check_exit_code` return `true` on error code 2, then some important errors can get ignored and I also get duplicated "invalid syntax" error from pyflakes (from python-language-server):

```
def convert_quantity(quantity: int) - float:     ■ invalid syntax     ■ invalid syntax
```

When I make it return `false`, then neovim spams error messages before I finish writing the code. What should I do in this situation?

jose-elias-alvarez:
@nawordar One option might be to have `check_exit_code` always return `true`, then set `format = raw` so you can check the output, then either feed it into one of the parsers or parse it manually.

jose-elias-alvarez:
Closed by #325.

PR #161, opened 2021-09-06
Append filname to temp_file

### LINKED ISSUE #159: Maintain file endings on `to_temp_file`
### Issues

### Feature description

I'm adding [this linter](https://github.com/priv-kweihmann/oelint-adv) to my editing setup. However it only takes files not stdin as input. This would work fine with `to_temp_file` however the linter uses the file-endings to categorize input files, since `to_temp_file` uses the format `/tmp/lua_XXXX` this brakes things.

One possible solution would be to fetch the `bufname` and add to the temp file name e.g. `/tmp/lua_XXXoriginalfilename.orgend`  

### Help

Yes

### Implementation help

Not to familiar with lua, but i could give it a go, but it might be an easy fix for someone else...

meck:
Should fix #159

### DIFF
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -145,9 +145,14 @@ M.timer = function(timeout, interval, should_start, callback)
     }
 end
 
-M.temp_file = function(content)
+M.temp_file = function(content, extension)
     local tmp_path = os.tmpname()
-    local fd = uv.fs_open(tmp_path, "w", 0)
+    if extension then
+        uv.fs_unlink(tmp_path)
+        tmp_path = tmp_path .. "." .. extension
+    end
+    -- open with (0700) permissions
+    local fd = uv.fs_open(tmp_path, "w", 384)
     uv.fs_write(fd, content, -1)
     uv.fs_close(fd)
 
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -220,7 +220,8 @@ M.generator_factory = function(opts)
             }
 
             if to_temp_file then
-                local temp_path, cleanup = loop.temp_file(get_content(params))
+                local filename = vim.fn.fnamemodify(params.bufname, ":e")
+                local temp_path, cleanup = loop.temp_file(get_content(params), filename)
 
                 spawn_args = u.table.replace(spawn_args, "$FILENAME", temp_path)
                 spawn_opts.on_stdout_end = cleanup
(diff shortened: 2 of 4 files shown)


lua/null-ls/loop.lua
```diff
@@ -145,9 +145,14 @@ M.timer = function(timeout, interval, should_start, callback)
     }
 end
 
-M.temp_file = function(content)
+M.temp_file = function(content, filename)
     local tmp_path = os.tmpname()
-    local fd = uv.fs_open(tmp_path, "w", 0)
+    if filename then
+       uv.fs_unlink(tmp_path)
+       tmp_path = tmp_path .. "_" .. filename:match("^.+/(.+)$")
```
jose-elias-alvarez, line 152: I think it would be more consistent to use `vim.fn.fnamemodify` to get the extension. Also, it seems more logical to have this method accept the prepared extension.

jose-elias-alvarez:
I tried this solution too, and I'm not sure I love it, mostly because I don't like the requirement to immediately unlink the temp file created by `os.tmpname`.

While looking into this, I realized that `os.tmpname` doesn't work on Windows, since it's hard-coded to generate a temp file with the template `/tmp/lua-XXXXXX`. That makes me think that we could just generate a path by ourselves from a filename, so that `my-file.lua` would generate something like `/tmp/null-ls_my-file.lua`. What do you think?

meck:
Seems reasonable, i don't know the lua ecosystem we need robust way to get a `/tmp` equivalent on windows.  A random UUID should be added to the filename if multiple files with the same name are open simultaneously?

jose-elias-alvarez:
Right, if we go down that route, we should add a random element to avoid collisions in that case and to (slightly) improve security, but it seems that random number generation isn't that robust in Lua, either. 

Either way, I think the temp file functionality needs attention in the future (I've read that `os.getenv("TMP")` will get the path to a temp dir in Windows, but we'd need someone to test it), and I think this solution is good enough for now, so I'm happy to merge it as-is if it solves your issue.

meck:
You're right that the unlink thing is bit ugly and should be changed, but for my purposes its fine. So please go head if it looks ok for now.

meck:
Or hold on, might try to rename the existing file instead of creating a new one... checking

jose-elias-alvarez:
I went ahead and fixed / added the test coverage. Thanks for working on this!

meck:
Great! i noticed now that renaming works as well. Saves the extra file dance.
but i guess that can wait.

``` lua
local tmp_path_w_ext = tmp_path .. "." .. extension
uv.fs_rename(tmp_path, tmp_path_w_ext)
tmp_path = tmp_path_w_ext
```

PR #162, opened 2021-09-06
refactor(loop): prefer using fs_mkstemp and fs_rename

jose-elias-alvarez:
This PR modifies `loop.temp_file` to use `uv.fs_mkstemp` when available, which should theoretically work on Windows (though we need a Windows user to actually test it). Also, when given an extension, the method now uses `uv.fs_rename` to rename the original temp file (instead of unlinking it). 

@meck Could you let me know if this PR works for you? And thank you for the suggestion to use `fs_rename`!

### DIFF
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -146,13 +146,29 @@ M.timer = function(timeout, interval, should_start, callback)
 end
 
 M.temp_file = function(content, extension)
-    local tmp_path = os.tmpname()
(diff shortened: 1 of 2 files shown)


meck:
Seems to work fine over here :+1:

jose-elias-alvarez:
Thanks for confirming! Going to merge this and hopefully confirm that it actually does work on Windows later this month when I have access to a PC.

PR #163, opened 2021-09-07
feat: introduce psalm php static analyser

gbprod:
This introduce builtin diagnostic for [psalm](https://psalm.dev), a php static analyser.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1412,3 +1412,25 @@ local sources = { null_ls.builtins.diagnostics.phpstan }
 ##### Requirements
 
 A valid `phpstan.neon` at root.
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Looks good! One small change (unless there's a special requirement I'm not seeing) might be to inline `from_json`, the same way we do [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/5b0e21625a920cf4a2d1ce1c516dd503b7a32ca9/lua/null-ls/builtins/diagnostics.lua#L334).

gbprod:
Done!

jose-elias-alvarez:
Thanks!

PR #176, opened 2021-09-09
Missing filetype in formatting.trim_whitespace

lbiaggi:
Hi, I tracked down a problem when I tried to load with `formatting.trim_whitespace` and I noticed that `filetype` was missing, so I inserted one empty not sure if this is the right answer. At least this addressed the problem and since at [lines](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/config.lua#L89-L93) you validate they, I think it's necessary.

Regards,
Lucas

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1036,16 +1036,21 @@ A simple wrapper around `awk` to remove trailing whitespace.
 ##### Usage
 
 ```lua
-local sources = { null_ls.builtins.formatting.trim_whitespace.with({
-    filetypes = { ... }
-}) }
+local sources = { null_ls.builtins.formatting.trim_whitespace }
 ```
 
 ##### Usage
-
-- `filetypes = one (must specify in with(), as above)`
+- `filetypes = { }`
 - `command = "awk"`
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
So my idea here (and I think I mentioned it in the documentation) was to specifically require the user to define filetypes and not provide a default, since it made sense for this formatter to be a sort of fallback when other formatters aren’t available. However, that was before we had wildcard sources, so I wonder if it make sense to make this change (we also have sequential formatting, so there shouldn’t be any weird interactions).

I want to think about this a bit more, but either way, could you please update the documentation to remove the bit about needing to set filetypes?

jose-elias-alvarez:
@lbiaggi I thought more about this and I think your change makes sense. If you update [the documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#trim_whitespace), I'll be happy to merge this.

lbiaggi:
@jose-elias-alvarez Okay, done

jose-elias-alvarez:
Thanks!

PR #180, opened 2021-09-10
Allow users to set the neovim binary path

### LINKED ISSUE #173: Option to set neovim path
### Issues

### Feature description

I have installed neovim on remote servers by downloading the appimage and extracting it somewhere. The installation directory is not in my path, rather I have an alias set to it. With null-ls I get the following error: 
```
cmd["nvim"] is not executable.
```
Which is expected as nvim is not in my path. When I add the neovim directory to `PATH` it works fine. Would it be possible to have a config option to set the path to nvim binary?

### Help

### Implementation help

_No response_

oponkork:
A new config option `cmd` allows users to set the path to the neovim binary. Fixes #173 

Sample config

```lua
require("null-ls").config({
      cmd = "/path/to/neovim/directory/nvim",
     ...
})
```

## TODO

- [X] Add option
- [x] Update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,13 @@ from a command to generate code actions.
 
 ## FAQ
 
+### How do I set the path to neovim binary?
+
+Set it while calling lsp setup for null-ls.
+```lua
+require("lspconfig")["null-ls"].setup({ cmd = { "/path/to/nvim" }, ... })
+```
+
 ### How do I format files?
 
 null-ls formatters run when you call `vim.lsp.buf.formatting()` or
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1102,6 +1102,24 @@ local sources = { null_ls.builtins.formatting.autopep8 }
 - `command = "autopep8"`
 - `args = { "-" }`
 
+#### [php-cs-fixer](https://github.com/FriendsOfPhp/PHP-CS-Fixer)
+
+##### About
+
+Formatter for `php` files
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.formatting.phpcsfixer }
+```
+
+##### Defaults
+
+- `filetypes = { "php" }`
+- `command = "php-cs-fixer"`
+- `args = { '--no-interaction', '--quiet', 'fix', "$FILENAME" }`
+
 ### Diagnostics
 
 #### [chktex](https://www.nongnu.org/chktex/)
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -47,18 +47,12 @@ The following code block shows the available options and their defaults.
 
 ```lua
 local defaults = {
-    cmd = "nvim",
     diagnostics_format = "#{m}",
     debounce = 250,
     default_timeout = 5000,
     sources = nil,
 }
 ```
-
-## cmd (string)
-
-Set the absolute path of the Neovim binary. Useful in case your Neovim installation
-directory is not in `PATH` environmental variable.
 
 ## diagnostics_format (string)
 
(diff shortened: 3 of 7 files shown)


jose-elias-alvarez:
@oponkork Sorry for the late review. The way you handled this works perfectly, but I think it makes sense to use lspconfig's `setup` for this case. What I had in mind was something like this:

```diff
diff --git a/lua/null-ls/rpc.lua b/lua/null-ls/rpc.lua
index c78be75..482cc91 100644
--- a/lua/null-ls/rpc.lua
+++ b/lua/null-ls/rpc.lua
@@ -21,7 +21,8 @@ function M.setup()

     local rpc_start = rpc.start
     rpc.start = function(cmd, cmd_args, dispatchers, ...)
-        if cmd == "nvim" then
+        local config = require("lspconfig")["null-ls"]
+        if config and cmd == config.cmd[1] then
             return M.start(dispatchers)
         end
         return rpc_start(cmd, cmd_args, dispatchers, ...)
```

This would then let users change the path to the Neovim binary like this:

```lua
require("lspconfig")["null-ls"].setup({ cmd = { "/path/to/nvim" } })
```

Could you make the change and update the documentation? This is kind of a special case, so I think we can move the explanation to another section.

oponkork:
Thanks for the review, I will update it. But it is not clear to me that why we should prefer the second approach over the first one. Can you please elaborate on that?

jose-elias-alvarez:
For now, it doesn't make much of a difference, but in the future the goal is to move all null-ls configuration into lspconfig, so I prefer to put new options there if possible.

oponkork:
@jose-elias-alvarez  Okay, cool. I have updated the code, however I am not sure where to document it now. Should I update the code blocks right above https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#options ?

oponkork:
Looks like `config` is not defined on when rpc setup function is called, and the if condition fails at https://github.com/oponkork/null-ls.nvim/blob/8b5428997c2448bac6b22f925e6420cb7a1ed7b1/lua/null-ls/rpc.lua#L29

Here's the modified function :
```lua
function M.setup()
    local rpc = require("vim.lsp.rpc")
    local config = require("lspconfig")["null-ls"]

    local rpc_start = rpc.start
    rpc.start = function(cmd, cmd_args, dispatchers, ...)
        if config and cmd == config.cmd[1] then
            return M.start(dispatchers)
        end
        return rpc_start(cmd, cmd_args, dispatchers, ...)
    end
end

```

jose-elias-alvarez:
This seems to work okay if we check for `config` inside `rpc.start`:

```diff
diff --git a/lua/null-ls/rpc.lua b/lua/null-ls/rpc.lua
index cf688a6..482cc91 100644
--- a/lua/null-ls/rpc.lua
+++ b/lua/null-ls/rpc.lua
@@ -18,10 +18,10 @@ local lastpid = 5000

 function M.setup()
     local rpc = require("vim.lsp.rpc")
-    local config = require("lspconfig")["null-ls"]

     local rpc_start = rpc.start
     rpc.start = function(cmd, cmd_args, dispatchers, ...)
+        local config = require("lspconfig")["null-ls"]
         if config and cmd == config.cmd[1] then
             return M.start(dispatchers)
         end

```

jose-elias-alvarez:
This looks good now. For documentation: I think it makes sense to add a section to the FAQ on the main README.md file about this case and how to handle it. Does that sound okay to you?

oponkork:
Yeah, that makes sense. I will do that and ping you.

oponkork:
@jose-elias-alvarez check now.

jose-elias-alvarez:
LGTM, thanks for working on this!

Issue #170, opened 2021-09-09
Adding phpcs diagnostics


marwan38:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Null-ls has existing feature support for php formatting (phpcbf, and the newly added php-cs-fixer).
php-cs is also a diagnostics tool and a linter for php. I'm new to neovim/vim land so it'd be quicker and more efficient for perhaps one of the more learned contributors here to add support for `phpcs`.

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

Would null-ls be able to support it without any added logic?

jose-elias-alvarez:
It really depends on the source itself. If it’s well-behaved (supports `stdin` and has well-formatted output), it should be simple, but otherwise it may require a little work.

I’m not at all familiar with PHP or the ecosystem, but maybe @gbprod (who has added some sources recently) could provide input?

marwan38:
It should be well-behaved according to your description. I use it with `diagnosticls` just fine, here is the config I use (found it floating on the internet):

```
   phpcs = {
      sourceName = "phpcs",
      command = "./vendor/bin/phpcs",
      debounce = 100,
      args = { "--report=emacs", "-s", "-" },
      offsetLine = 0,
      offsetColumn = 0,
      formatLines = 1,
      formatPattern = {
         [[^.*:(\d+):(\d+):\s+(.*)\s+-\s+(.*)(\r|\n)*$]],
         { line = 1, column = 2, security = 3, message = { "[phpcs] ", 4 } },
      },
      securities = { error = "error", warning = "warning" },
      rootPatterns = { ".git", "vendor", "composer.json" },
   },

```

gbprod:
I'll take a look.
IMHO, we should use phpcs json output that is more reliable that the emacs' output.
The output looks like the phpstan's one (https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#printing-a-json-report), the difference is that phpcs can use stdin. 

I try to make a PR before the end of the week ;)

marwan38:
@gbprod that's kind of you to do, thank you! :)

gbprod:
@marwan38 Can you try #178 and tell me if it works for you ?

Regarding to you're diagnosticls settings, I think you should set it up like this : 

```lua 
[...]

PR #178, opened 2021-09-10
introduce phpcs diagnostic

gbprod:
This commit adds [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) analyser.

Related to #170

PR #185, opened 2021-09-12
Allow phpcs builtin messages to always reach output

sirbrillig:
Support for phpcs as a diagnostic tool was added to the builtins in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/178, but (unless I'm doing something incorrectly) there were a few issues. 

Firstly, this adds some overrides to the JSON parsing of the phpcs output so that its properties are correctly mapped. In particular, the `severity` was probably not correct as phpcs's message `severity` property does not map directly to the ones used by `null-ls`.

Secondly, the output from phpcs is to STDOUT, not STDERR, and so this PR changes the builtin's `from_stderr` to `false` (this sometimes worked anyway because of a separate issue, see below).

Sometimes, additional status messages are included in the phpcs output even when reporting in JSON mode, creating invalid JSON. This PR also adds the `-q` option to the phpcs call to silence the status reporting output to prevent the possibility of creating unparsable JSON.

Finally, because phpcs returns a 1 or 2 error code depending on the severity of the messages it reports, it was possible for the builtin's `check_error_code` to consider a successful run with errors to be reported as a failed run, directing its output to stderr. This meant that the builtin would work or not work depending on the type of errors being reported. To resolve this we also pass config options to phpcs to cause it to only return an error code if there is an actual error with the command.

Before this PR, the phpcs diagnostic builtin would sometimes not report any warnings or errors. After this PR it should always report correctly.

Fixes https://github.com/jose-elias-alvarez/null-ls.nvim/issues/170

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -511,7 +511,7 @@ M.phpcs = h.make_builtin({
             "ignore_errors_on_exit",
             "1",
             -- process stdin
-            "-"
+            "-",
         },
         format = "json_raw",
         to_stdin = true,
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -497,7 +497,22 @@ M.phpcs = h.make_builtin({
     filetypes = { "php" },
     generator_opts = {
         command = "phpcs",
-        args = { "--report=json", "-q", "-s", "-" },
+        args = {
+            "--report=json",
+            -- silence status messages during processing as they are invalid JSON
+            "-q",
+            -- always report codes
+            "-s",
+            -- phpcs exits with a non-0 exit code when messages are reported but we only want to know if the command fails
+            "--runtime-set",
+            "ignore_warnings_on_exit",
+            "1",
+            "--runtime-set",
+            "ignore_errors_on_exit",
+            "1",
+            -- process stdin
+            "-"
+        },
         format = "json_raw",
         to_stdin = true,
         from_stderr = false,
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -497,7 +497,7 @@ M.phpcs = h.make_builtin({
     filetypes = { "php" },
     generator_opts = {
         command = "phpcs",
-        args = { "--report=json", "-s", "-" },
+        args = { "--report=json", "-q", "-s", "-" },
         format = "json_raw",
         to_stdin = true,
         from_stderr = false,
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -505,7 +505,16 @@ M.phpcs = h.make_builtin({
             return code <= 1
         end,
         on_output = function(params)
-            local parser = h.diagnostics.from_json({})
+            local parser = h.diagnostics.from_json({
+                attributes = {
+                    severity = "type",
+                    code = "source",
+                },
+                severities = {
+                    ERROR = h.diagnostics.severities["error"],
+                    WARNING = h.diagnostics.severities["warning"],
+                }
+            })
             params.messages = params.output
                     and params.output.files
                     and params.output.files["STDIN"]
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -500,7 +500,7 @@ M.phpcs = h.make_builtin({
         args = { "--report=json", "-s", "-" },
         format = "json_raw",
         to_stdin = true,
-        from_stderr = true,
+        from_stderr = false,
         check_exit_code = function(code)
             return code <= 1
         end,
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -513,7 +513,7 @@ M.phpcs = h.make_builtin({
                 severities = {
                     ERROR = h.diagnostics.severities["error"],
                     WARNING = h.diagnostics.severities["warning"],
-                }
+                },
             })
             params.messages = params.output
                     and params.output.files


marwan38:
@sirbrillig Do you use phpcsfixer or phpcbf, by chance, with null-ls as well? They don't output anything, similar to the issue with phpcs. Wondering if they're caused by the same thing?

gbprod:
Which version of php cs are you running ?
Seems that stderr output is used since the 4.0's version of phpcs, maybe we should have 2 built-ins ?

marwan38:
I discovered the source of this issue. Both phpcs and fixer work perfectly fine now (main branch, not this PR).

I had `xdebug` listening during all requests, and that caused it to error out when running php. You can see this error here: https://github.com/jose-elias-alvarez/null-ls.nvim/issues/170#issuecomment-917251210.

I just disabled xdebug, and reverted to the main branch (not this PR), and it works just fine! @sirbrillig I suspect you have a similar thing that is causing php to error out something, like for example, xdebug like me, or you have a duplicate extension on.

@gbprod FYI.

sirbrillig:
@gbprod 

> Which version of php cs are you running ?

Ah, interesting! I was running phpcs 3.5.4.

However, I tried version 3.6.0 and I get a very different problem! There seems to be a bug in phpcs 3.6 where the output includes some status output. For example, here's what I get when running `phpcs` manually (with the same flags as `null-ls`) with 3.6:

```
E 1 / 1 (100%)


{"totals":{"errors":43,"warnings":88,"fixable":120},"files":{"STDIN":{"errors":43,"warnings":88,"messages":[ ... ]}}}
```

See that `E 1 / 1 (100%)`? That looks like a regression. That's causing the output to be invalid JSON and probably breaks the parser. I've just created [a bug report for this issue](https://github.com/squizlabs/PHP_CodeSniffer/issues/3436), but in the meantime it's probably best to use phpcs 3.5.4.

> Seems that stderr output is used since the 4.0's version of phpcs, maybe we should have 2 built-ins ?

I don't see a 4.0 version in the [phpcs releases](https://github.com/squizlabs/PHP_CodeSniffer/releases) and I don't see anything in the changelog that mentions changing its output to stderr (which IMO would be unexpected since it would make it difficult to eg: use the linter with a pager).

Looking through the commits, I did find that [status output is being moved to stderr](https://github.com/squizlabs/PHP_CodeSniffer/commit/d4f33dbf45aa5bb78e077b0e73306746c8bd98cf), although the actual output is still being printed to stdout, but that is for an unreleased version (perhaps that's what you're using).

@marwan38 

Thank you for the good idea! I do not have xdebug turned on, and I can see that phpcs does output to stdout and not stderr on my machine. For example, if I run `phpcs --report=json path/to/file.php > test_file`, then all output is directed to the test file even though I'm not redirecting stderr (`>` in unix only redirects stdout). If you run the above command, do you still see the linting output? What appears in your test file?

> Both phpcs and fixer work perfectly fine now (main branch, not this PR).

This is very interesting. Again, there's a lot I don't know about this tool, but if I'm right and phpcs is sending its output to stdout, then [these lines](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/da8bb757c630b96fd26030df56fd3a070fbf16a1/lua/null-ls/helpers.lua#L171-L174) should cause the output to be erased. I'm very curious how you're able to get output to appear in null-ls. 

> I suspect you have a similar thing that is causing php to error out something

Wouldn't that do the opposite and cause the output to go to stderr? 🤔

sirbrillig:
> See that E 1 / 1 (100%)? That looks like a regression. That's causing the output to be invalid JSON and probably breaks the parser. I've just created a bug report for this issue, but in the meantime it's probably best to use phpcs 3.5.4.

Apparently I was mistaken about this. The issue also occurs in phpcs 3.5.4, but the correct way to consume raw JSON output (or one of the correct ways) from phpcs is to use the `-q` option to silence the status reporting. I added that option to this PR.

sirbrillig:
Ok, now I don't know what's happening anymore 😂  I just wrote all the above and finished my updates, and the phpcs builtin was working on this branch.

I made some updates to my phpcs standards configuration locally, and now I'm finally able to reproduce what you described above @marwan38... for some reason the phpcs messages output is going to stderr. I'm going to look more into how the external command gets called to see if I can understand why it sometimes was going to stderr and sometimes to stdout.

sirbrillig:
Ah, it's because of `check_exit_code`. phpcs returns a 1 or 2 exit code depending on if there are warnings or errors being reported, but this builtin considers a non-1 exit code to mean that the command itself failed, causing its stdin output to go to stderr.

Probably at various times while testing this PR and the main branch, all of us have had a different set of errors or warnings being displayed, sometimes causing them to go to stderr and sometimes to stdout. 

While we could ignore those error codes specifically, they aren't clearly documented on the site and really we just want to know if phpcs itself failed. Fortunately, [there's a way to do that](https://github.com/squizlabs/PHP_CodeSniffer/issues/2412#issuecomment-476234808): the `ignore_warnings_on_exit` and `ignore_errors_on_exit` runtime config options. I've added those and now this PR should work for everyone.

marwan38:
@sirbrillig Sounds like you had a rough night :). Are the rest of the changes, apart from ignore flag, still need to be added in?

sirbrillig:
Haha. Yeah, this was breaking my brain a little bit.

Yes, I believe all the changes are needed, but for different reasons. I edited this PR description to try to explain the choices I made (`-q` to make sure JSON is always valid, ignore warnings to make the exit code 0, set output to stdout because it is, map the right severity and code).

gbprod:
Hi ! 
I've just test your PR and it works fine for me using phpcs 3.6 ! 
I think we can merge it  @jose-elias-alvarez :)
Thanks @sirbrillig @marwan38

marwan38:
Does any of this stuff need to be applied for phpcsfixer or phpcbf as well?

sirbrillig:
> Does any of this stuff need to be applied for phpcsfixer or phpcbf as well?

I've never used `phpcsfixer` but IIRC it's totally unrelated to `phpcs`, which doesn't mean it might not have similar problems but they're probably unique to that tool and are worth exploring in their own issue.

`phpcbf` _is_ part of `phpcs`, and you're right. I see some of the same issues there in my testing. Unlike `phpcs`, we cannot set options to ignore error codes there so we have to detect the error code itself. I added support for that in a separate PR https://github.com/jose-elias-alvarez/null-ls.nvim/pull/186

jose-elias-alvarez:
Thanks to everyone for investigating and working on this. Much appreciated!

PR #186, opened 2021-09-13
Allow phpcbf builtin to consider exit codes 1 or 2 as success

sirbrillig:
Related to https://github.com/jose-elias-alvarez/null-ls.nvim/pull/185

Like `phpcs`, `phpcbf` returns a 1 or 2 exit code if it detects issues that need to be changed. The downside is that this exit code can be interpreted as meaning that the command failed, when in fact it did not (the result is that the stdout output ends up being moved to stderr by the formatter and not used).

This PR alters the phpcbf formatting builtin so that it will treat a 1 or 2 error code as a success, allowing formatting to proceed.


jose-elias-alvarez:
Thank you!

PR #188, opened 2021-09-13
Add "stylelint" diagnostics and formatter

andersevenrud:
This PR has the following:

* Adds "stylelint" diagnostics builtin
* Adds "stylelint" formatting builtin
* Updates documentation

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1472,3 +1472,22 @@ local sources = { null_ls.builtins.diagnostics.phpcs }
 - `command = "phpcs"`
 - `args = { "--report=json", "-s", "-" }`
 
+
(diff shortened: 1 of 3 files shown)


gegoune:
@andersevenrud Just FYI, there is [stylelint-lsp](https://github.com/bmatcuk/stylelint-lsp) which can be easily used with [lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#stylelint_lsp).

andersevenrud:
@gegoune Woah, I didn't know about that one. Thanks a lot!

jose-elias-alvarez:
Thank you! It looks like stylelint-lsp is a more powerful option, but I think it's helpful to have this, too.

PR #190, opened 2021-09-14
Do not exit early when `suppress_errors = true`

overhacked:
The previous behavior was to return from the generator early if
there is any `error_output` from stderr even if
`suppress_errors = true` (given the default `from_stderr = false`).

This didn't really make sense, because aborting on an error was
already covered by `check_exit_code`, which returns an error
before the generator even starts processing output.

In the case that a command doesn't use exit codes (always returns 0),
the `raw` output method could be used to check the error output.

### DIFF
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -173,7 +173,10 @@ M.generator_factory = function(opts)
                     error_output = nil
                 end
 
-                if error_output and not (suppress_errors or format == output_formats.raw or format == output_formats.json_raw) then
+                if
+                    error_output
+                    and not (suppress_errors or format == output_formats.raw or format == output_formats.json_raw)
+                then
                     error("error in generator output: " .. error_output)
                     done()
                     return
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -173,10 +173,8 @@ M.generator_factory = function(opts)
                     error_output = nil
                 end
 
-                if error_output and not (format == output_formats.raw or format == output_formats.json_raw) then
-                    if not suppress_errors then
-                        error("error in generator output: " .. error_output)
-                    end
+                if error_output and not (suppress_errors or format == output_formats.raw or format == output_formats.json_raw) then
+                    error("error in generator output: " .. error_output)
                     done()
                     return
                 end


overhacked:
This is a real issue with `ansible-lint`, which insists on putting output on stdout, even with the `-q` (quiet) flag, such as:
```
WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: playbook.yml
WARNING  Listing 18 violation(s) that are fatal
```

The second one particularly creates a problem, because I don't want null-ls to detect an "error" and stop processing `ansible-lint`'s output. I really want it to process stdout so I can see those 18 "violations" as LSP diagnostics.

jose-elias-alvarez:
I'm not sure I understand the situation – if the linter is outputting to `stderr`, why not use `from_stderr` or set `format` to `raw`? 

In my experience, a lot of sources have ambiguous exit codes. The current behavior of `suppress_errors` is meant to handle cases like a user formatting a file that a formatter can't parse, since what we want to is suppress (i.e. prevent reporting) the error and return early. 

If what you want to do isn't possible with the current options, I think we should add another option instead of modifying the current one, but again, I'm not sure I understand, so I can't say for sure.

overhacked:
I understand now the reason for bailing out on `stderr` output even when the exit code is `0`. The problem with `ansible-lint` is that it always outputs to both `stderr` *and* `stdout`. The lints I want `null-ls` to consume show up on `stdout`, but there's always some not-so-useful stuff also on `stderr`.

<details>
<summary>Example Output</summary>

```
(stderr) WARNING  Overriding detected file kind 'yaml' with 'playbook' for given positional argument: -
(stderr) WARNING  Listing 18 violation(s) that are fatal
(stdout) :1: [meta-no-tags] [HIGH] Tags must contain lowercase letters and digits only, invalid: 'FreeBSD'
(stdout) playbook.yml:40: [deprecated-local-action] [MEDIUM] Do not use 'local_action', use 'delegate_to: localhost'
(stdout) playbook.yml:44: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) playbook.yml:45: [deprecated-local-action] [MEDIUM] Do not use 'local_action', use 'delegate_to: localhost'
(stdout) playbook.yml:50: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) playbook.yml:51: [deprecated-local-action] [MEDIUM] Do not use 'local_action', use 'delegate_to: localhost'
(stdout) playbook.yml:103: [deprecated-local-action] [MEDIUM] Do not use 'local_action', use 'delegate_to: localhost'
(stdout) playbook.yml:126: [no-handler] [MEDIUM] Tasks that run when changed should likely be handlers
(stdout) playbook.yml:230: [unnamed-task] [MEDIUM] All tasks should be named
(stdout) roles/my-role/meta/main.yml:1: [role-name] [HIGH] Role name my-role does not match ``^[a-z][a-z0-9_]+$`` pattern
(stdout) roles/my-role/tasks/aws_s3.yml:8: [empty-string-compare] [HIGH] Don't compare to empty string
(stdout) roles/my-role/tasks/main.yml:30: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) roles/my-role/tasks/main.yml:41: [unnamed-task] [MEDIUM] All tasks should be named
(stdout) roles/my-role/tasks/main.yml:45: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) roles/my-role/tasks/main.yml:50: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) roles/my-role/tasks/main.yml:63: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) roles/my-role/tasks/main.yml:68: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect
(stdout) roles/my-role/tasks/main.yml:80: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect

```

</details>

In particular, the `Listing 18 violation(s) that are fatal` "error" line can't be suppressed, which is funny, because isn't listing violations the whole point of a linter?

jose-elias-alvarez:
Got it. This is pretty weird behavior from the linter, but I imagine it's not unique, either. Could we add a new config option to handle cases like this? Something like `ignore_errors` seems logical.

overhacked:
I like the idea of another option, but the name `suppress_errors` is already slightly confusing, because it isn’t obvious whether the suppressing is of error output (right) or of error behavior (wrong). “**Suppress**” and “**ignore**” are also easily confused words. What if the new option is named `stop_on_error` and the default is `true`?

jose-elias-alvarez:
I like the idea of adding a new output format, but since we haven't yet encountered any linters where we care about both `stderr` and `stdout` (if I understand correctly, in this case we just want to discard `stderr`), I'd prefer to save it for the future and go with the first option for now.

jose-elias-alvarez:
In the process of fixing a separate linter, #218 added an `ignore_stderr` option that should handle this case as well. Could you check if that works for you?

overhacked:
#218 does the trick. Updated to `main` and changed my `ansible-lint` linter. Thanks @Mange & @jose-elias-alvarez!

FollieHiyuki:
@overhacked sorry for bugging you. Can you share how you made `ansible-lint` working with `null-ls`?

Thanks in advance.

PR #184, opened 2021-09-11
Fix shellcheck column range for tab indents

dialetic:
Fixes #183.

PR #192, opened 2021-09-15
Follow shellcheck external sources

Gelio:
This PR adds support for following external sources for [shellcheck](https://www.mankier.com/1/shellcheck).

I have added expanding `$DIRNAME` to the parent directory of the buffer. I could not find another way to make it work.

The newly added test passes. The only tests that do not pass also did not pass on `main`, so I don't think my changes are the root cause.

### DIFF
--- a/lua/null-ls/builtins/diagnostics.lua
+++ b/lua/null-ls/builtins/diagnostics.lua
@@ -184,7 +184,7 @@ M.shellcheck = h.make_builtin({
     filetypes = { "sh" },
     generator_opts = {
         command = "shellcheck",
-        args = { "--format", "json1", "-" },
+        args = { "--format", "json1", "--source-path=$DIRNAME", "--external-sources", "-" },
         to_stdin = true,
         format = "json",
         check_exit_code = function(code)
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -31,6 +31,9 @@ local parse_args = function(args, params)
         if string.find(arg, "$FILENAME") then
             arg = u.string.replace(arg, "$FILENAME", params.bufname)
         end
+        if string.find(arg, "$DIRNAME") then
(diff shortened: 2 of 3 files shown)


jose-elias-alvarez:
Thank you, this seems reasonable. Just to verify, since I'm far from a `shellcheck` expert:

1. Did the existing `$ROOT` argument replacement not work?
2. Will this affect existing users in terms of errors / performance?

I just merged #184, which contained another fix for `shellcheck` arguments, so I think this needs to be rebased, too.

Gelio:
I'm not an shellcheck expert myself, but my take on the questions is:

> 1. Did the existing `$ROOT` argument replacement not work?

Generally, I have seen `shellcheck source` comments that point to the other script using a relative path from the source script's location. If we passed `$ROOT`, if I'm not mistaken, it would use the current working directory of `null-ls`, which may not be the same as the script's parent directory. For example, when starting neovim in the repository's root (`/abc`) and opening a file `/abc/somedir/script.sh`), if the file used

```sh
#shellcheck source=./another-script.sh
```

then using `$ROOT` would resolve the path to `/abc/another-script.sh`, whereas using `$DIRNAME` would resolve to `/abc/somedir/another-script.sh` (which looks to be how many people see it should behave https://github.com/koalaman/shellcheck/issues/769#issuecomment-277707597)

> 2. Will this affect existing users in terms of errors / performance?

No clue, actually. However, I believe that when you add those `shellcheck source` comments, you expect shellcheck to also inspect those files. Otherwise, there is little benefit to adding those comments, so people may as well not add them.

I was not running this PR as my active one, so I can't say with certainty how the performance is. However, running the `shellcheck` from the CLI I didn't find a difference with and without `--external-sources`

jose-elias-alvarez:
Got it, thanks for confirming. I think the changes make sense, so we'll see if there's any issues later on.

Gelio:
Alrighty, thanks for quick responses :+1:

PR #193, opened 2021-09-15
fix: pass filename to shfmt

Gelio:
Helps shfmt respect .editorconfig

Reference: https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#parser-flags


jose-elias-alvarez:
Makes sense, thank you!

PR #203, opened 2021-09-19
RFC - Dynamic source conditions

levouh:
Adds a `dynamic_condition` that is evaluated at "runtime" for all sources that discerns whether or not it will be run each time they have the potential to be run.

See associated discussion [here](https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/202) as to why this might be useful.

Some things still to do, but curious on your thoughts @jose-elias-alvarez before messing with getting these going:

- [x] Add tests

### DIFF
--- a/lua/null-ls/generators.lua
+++ b/lua/null-ls/generators.lua
@@ -18,6 +18,12 @@ M.run = function(generators, params, postprocess, callback)
         for _, generator in ipairs(generators) do
             table.insert(futures, function()
                 local copied_params = vim.deepcopy(params)
+
+                local runtime_condition = generator.opts and generator.opts.runtime_condition
+                if runtime_condition and not runtime_condition(copied_params) then
+                    return
+                end
+
                 local to_run = generator.async and a.wrap(generator.fn, 2) or generator.fn
                 local ok, results = pcall(to_run, copied_params)
                 a.util.scheduler()
--- a/doc/HELPERS.md
+++ b/doc/HELPERS.md
@@ -36,6 +36,7 @@ helpers.generator_factory({
     timeout, -- number (optional)
     to_temp_file, -- boolean (optional)
     use_cache, -- boolean (optional)
+    runtime_condition, -- function (optional)
 })
 ```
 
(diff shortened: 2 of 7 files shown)


lua/null-ls/generators.lua
```diff
@@ -94,7 +94,12 @@ end
 
 M.get_available = function(filetype, method)
     return vim.tbl_filter(function(generator)
-        return not generator._failed and u.filetype_matches(generator.filetypes, filetype)
+        local dynamic_condition = true
+        if generator.opts.dynamic_condition then
+            dynamic_condition = generator.opts.dynamic_condition()
```
levouh, line 99: File type matching should likely be done first to avoid calling these functions unnecessarily.

jose-elias-alvarez:
This is a great idea and works well. Strictly speaking, since null-ls allows running generators on demand, something like this is already possible, but I think this is a clean addition and far more user-friendly. 

I wonder if it might be worth it to move the `dynamic_condition` check to the generator loop in `generators.run`. This would give the callback access to `params`, which I imagine might be useful to simplify checking whether a source should run for the given file or not. Maybe something like this:

```diff
diff --git a/lua/null-ls/generators.lua b/lua/null-ls/generators.lua
index 41510af..9aea507 100644
--- a/lua/null-ls/generators.lua
+++ b/lua/null-ls/generators.lua
@@ -18,6 +18,12 @@ M.run = function(generators, params, postprocess, callback)
         for _, generator in ipairs(generators) do
             table.insert(futures, function()
                 local copied_params = vim.deepcopy(params)
+
+                local dynamic_condition = generator.opts.dynamic_condition
+                if dynamic_condition and not dynamic_condition(copied_params) then
+                    return
+                end
+
                 local to_run = generator.async and a.wrap(generator.fn, 2) or generator.fn
                 local ok, results = pcall(to_run, copied_params)
                 a.util.scheduler()
@@ -92,19 +98,9 @@ M.run_registered_sequentially = function(opts)
     M.run_sequentially(generators, make_params, postprocess, callback, after_all)
 end
```

What do you think? Other than that, I'm happy to merge this once we have tests (unit tests and a simple E2E test would be ideal).

levouh:
Makes sense to me, having access to `params` would be nice as well. I also think a name like `runtime_condition` might be better as that's what we're doing.

I'll look at this patch and adding tests to get it through sometime this week. Thanks for the feedback.

levouh:
Let me know how that looks @jose-elias-alvarez.

jose-elias-alvarez:
Looks great, thank you for working on this (and for the improvements to the documentation)!

PR #210, opened 2021-09-27
Add standardrb formatter

Mange:
[Standardrb](https://github.com/testdouble/standard) is a Ruby style guide, formatter and fixer based on Rubocop.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -948,6 +948,24 @@ local sources = { null_ls.builtins.formatting.shellharden }
 - `command = "shellharden"`
 - `args = { "--transform", "$FILENAME" }`
 
+#### [standardrb](https://github.com/testdouble/standard)
+
+##### About
+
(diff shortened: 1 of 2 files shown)


Mange:
This does not appear to work correctly for me. I'll investigate it a bit and see if I can get it to work. It's exceedingly hard to debug issues here.

In either case, I'll also probably add a diagnostic implementation for Standardrb and Rubocop as well.

jose-elias-alvarez:
Do you have any idea at what point it's failing? Using the `debug` option should give you an idea of the command's output and what null-ls is doing with it. Piping to `awk` is something we also do for `rubocop`, but it seems like a potential failure point that we can avoid by using Lua to parse the output.

Mange:
Thanks. With `debug` I could find the root of the error. Turns out the `rubocop` formatter was broken, so my copy with adjustments also didn't work properly.

I've implemented working diagnostics adapters for both in my local branch and will fix the formatters as well and push a new version of this.

PR #218, opened 2021-09-29
Fix rubocop formatter, add diagnostics, add standardrb for both

Mange:
Retrying #210.

# Fixing `rubocop`

Rubocop outputs both the violations it detected and the file contents to STDOUT. The command tried to pipe this to awk in order to cut off the first half and only output the formatted file contents.

This did not work as the command wasn't run in a shell and the pipes and shell constructs were passed as real command line arguments, which Rubocop confused to be other files to check. The `--stdin` option is not allowed together with filenames, and so Rubocop exited with an error.
Adding the `--stderr` option to Rubocop makes it output only the file contents to STDOUT and the violations to STDERR.

Added a new option called `ignore_stderr` that resets STDERR after running the command so the command doesn't fail because of operator output.

Now the formatter works correctly.

# Adding `standardrb` formatter

[`standardrb` is another Ruby formatter, based on Rubocop](https://github.com/testdouble/standard), but with a slightly more modern take on the style guide, as well as fewer rules enabled.

# Adding diagnostics for both

Adds both a `rubocop` and a `standardrb` diagnostics builtin. They output their "offenses" as JSON, which is then picked up and emitted.

**Rubocop diagnostics**

Screenshot_2021-09-29_22-54-49

**standardrb diagnostics**

Screenshot_2021-09-29_22-54-16

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -947,6 +947,24 @@ local sources = { null_ls.builtins.formatting.shellharden }
 - `command = "shellharden"`
 - `args = { "--transform", "$FILENAME" }`
 
+#### [standardrb](https://github.com/testdouble/standard)
+
+##### About
+
+Ruby Style Guide, with linter & automatic code fixer. Based on Rubocop.
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.formatting.standardrb }
+```
+
+##### Defaults
+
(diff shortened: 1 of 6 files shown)


lynndylanhurley:
This fix works for me. Thanks @Mange !!

jose-elias-alvarez:
Looks great! Thanks for looking into this so thoroughly.

PR #232, opened 2021-10-04
feat(markdownlint): add rule info

younger-1:
Add rules info like `MD012/no-multiple-blanks` as `diagnostics.code`. All rules are in [here](https://github.com/DavidAnson/markdownlint/#rules--aliases)

This is what it look in [trouble.nvim](https://github.com/folke/trouble.nvim).

Without `diagnostics.code`
mdlint0

With `diagnostics.code`
mdlint2


jose-elias-alvarez:
I went ahead and fixed the tests that were failing because of this, but it looks good. Thank you!

PR #242, opened 2021-10-06
Add some more prettier supported filetypes

mcchrish:
Also copied the same filetype set to prettier_d_slim assuming everything should work there too.


mcchrish:
Mentioned and recommended `prettierd` in the about sections.

jose-elias-alvarez:
Thanks! Let's hope the `prettier_d_slim` PR gets merged, too.

PR #251, opened 2021-10-08
feat(builtins): create copies from with() method

### LINKED ISSUE #247: Declare multiple time the same formatter but with different settings
### Discussed in https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/238

<sup>Originally posted by **Nargonath** October  5, 2021</sup>
I'm using `prettierd` as a formatter for many different filetypes. I use the default builtin `prettierd` object for `yaml`, `json` and `markdown` however for `javascript` I use a `conditional` since I need to either use `eslint_d` or `prettierd` depending on the configuration file present in the project root. Here is my attempt:

```lua
nullLs.config({
	sources = {
		require("null-ls.helpers").conditional(function(utils)
			local usesEslint = vim.fn.glob(".eslintrc*") ~= ""
				or (
					utils.root_has_file("package.json")
					and vim.fn.filereadable("package.json")
					and vim.fn.json_decode(vim.fn.readfile("package.json"))["eslintConfig"]
				)

			return usesEslint and nullLs.builtins.formatting.eslint_d
				or nullLs.builtins.formatting.prettierd.with({
					filetypes = { "javascript" },
				})
		end),
		nullLs.builtins.formatting.prettierd.with({
			filetypes = { "json", "markdown", "yaml" },
		}),
	},
})
```

The problem is that the `prettierd` formatter is not used for `json`, `markdown` or `yaml` anymore but only for `javascript` which means only the first conditional declaration is used. IMO I cannot merge both `prettierd` sources since I always want to use `prettierd` for files other than `javascript` and conditionally use it for `javascript`.</div>

Right now there is no _builtin_ way of doing it. You can find in the discussion @jose-elias-alvarez's suggestion using `vim.deepcopy` which works but leaves lots of boilerplate in userland.

My suggestion would be to transform `with` and make it immutable. That way every time you call `with` you actually receives in return a deep copy of the source object (by default the built-in one). This would allow you to declare multiple times the same tool with different settings while also keeping the same API. I would suggest making the option `name` mandatory with the new `with` to facilitate debugging but that would be a breaking change. If we do it, we might as well rename `with` to something like `fork` or `clone`.

jose-elias-alvarez:
Closes #247.

As discussed in the linked issue, this PR makes it so that the `with` method for built-ins automatically returns a copy of the built-in, making it possible to register the same built-in multiple times with different config options. I also changed how the registration check works to allow registering multiple copies with the same name, which avoids the need to change the name when making a copy.

One pitfall is that if a user registers the same built-in multiple times with, say, an overlapping condition, the source will naturally be registered multiple times, and that fact won't be immediately obvious (since it will only show up once in `:NullLsInfo`). For example, this config will cause `stylua` to run twice if Neovim is opened in a working directory containing both `stylua.toml` and `other-project-marker`:

```lua
local b = null_ls.builtins

local sources = {
    b.formatting.stylua.with({
        condition = function(utils)
            return utils.root_has_file("stylua.toml")
        end,
    }),
    b.formatting.stylua.with({
        condition = function(utils)
            return utils.root_has_file("other-project-marker")
        end,
    }),
}
```

That's arguably just an issue with the user's configuration, though, and I think avoiding the need to set a unique name is convenient (and matches user expectations), but if this is a problem, we could require unique names and warn the user in this situation. 

@Nargonath Let me know what you think.

### DIFF
--- a/lua/null-ls/config.lua
+++ b/lua/null-ls/config.lua
@@ -65,11 +65,11 @@ local should_register = function(source, filetypes, source_methods)
 
     for _, method in ipairs(source_methods) do
         config._methods[method] = config._methods[method] or {}
-        if config._methods[method][name] then
+        if config._methods[method][name] and not source._is_copy then
             return false
         end
 
-        config._methods[method][name] = filetypes
+        config._methods[method][name] = vim.list_extend(config._methods[method][name] or {}, filetypes)
     end
 
     return true
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -317,48 +317,53 @@ M.make_builtin = function(opts)
     })
 
     builtin.with = function(user_opts)
-        builtin.filetypes = user_opts.filetypes or builtin.filetypes
+        local builtin_copy = vim.deepcopy(builtin)
+        setmetatable(builtin_copy, getmetatable(builtin))
+        builtin_copy._is_copy = true
+
+        builtin_copy.filetypes = user_opts.filetypes or builtin_copy.filetypes
 
         -- Extend args manually as vim.tbl_deep_extend overwrites the list
         if
             user_opts.extra_args
             and (type(user_opts.extra_args) == "function" or vim.tbl_count(user_opts.extra_args) > 0)
         then
-            local builtin_args = builtin._opts.args
-            local extra_args = user_opts.extra_args
+            local original_args = builtin_copy._opts.args
+            local original_extra_args = user_opts.extra_args
 
-            builtin._opts.args = function(params)
-                local builtin_args_cpy = type(builtin_args) == "function" and builtin_args(params)
-                    or vim.deepcopy(builtin_args)
+            builtin_copy._opts.args = function(params)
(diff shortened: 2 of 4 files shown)


jose-elias-alvarez:
> I haven't looked at the code yet but from what you wrote in the description what if we just add an index automatically to the name of the duplicated config? This way the user would not need to change the name themselves but they would still appear separately on `:NullLsInfo`? Do you think that would work? If they don't like these automatically generated tests, they'll be able to override them themselves through the `with` options.

This is definitely an option. The disadvantage is that it might be a little confusing for users, since the meaning of, say, `eslint (1)` wouldn't be immediately clear. We could increment the index only if there is a conflict with a previously registered source, but I'd rather keep it simple for now and change this if we find that this is causing issues. Does that sound okay?

Nargonath:
I agree @jose-elias-alvarez let's keep it simple for now regarding the copy naming, we'll see how it goes.

Thanks for landing this. I got busy with life and didn't get the chance to revisit before you needed to merge. I'm looking forward to try this new behavior. 😄

PR #252, opened 2021-10-08
Builtin source for Cppcheck

bernardassan:
Add builtin source for cppcheck and documentation in BUILTIN.md

### DIFF
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -51,7 +51,7 @@ M.handler = function(original_params)
     end
 
     local params = u.make_params(original_params, methods.map[method])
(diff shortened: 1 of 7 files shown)


jose-elias-alvarez:
null-ls diagnostic positions should (theoretically!) be consistent, since it expects generators to return 1-indexed diagnostics and then converts them to 0-indexed LSP diagnostics, but apart from potential bugs, there's also a lot of inconsistency in how sources themselves index positions. Feel free to open an issue for this and any other broken / misaligned sources, since I'm sure there's a lot I haven't caught. I'll also make sure to clamp positions to prevent errors in cases like this where the source's behavior is incorrect.

PR #253, opened 2021-10-10
feat(diagnostics.severities): Added fallback severity for unspecified codes

### LINKED ISSUE #196: Error yielded by flake8 plugins are discarted
### FAQ

- [X] I have checked [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

### Neovim Version

NVIM v0.6.0-dev+267-g077150448 Build type: RelWithDebInfo LuaJIT 2.0.5 Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions     -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security     -fstack-clash-protection -fcf-protection -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/build/neovim-git/src/build/config -I/build/neovim-git/src/neovim-git/src -I/usr/include -I/build/neovim-git/src/build/src/nvim/auto -I/build/neovim-git/src/build/include Compiled by builduser  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/usr/share/nvim"  Run :checkhealth for more info

### Steps to reproduce

- Install `flake8` and `flake8-bugbear`
- Use this snippet:

```python
for _ in list():
    pass
```

### Expected behavior

Should show a diagnostic on line on line 1:

```console
$ flake8 test.py
test.py:1:10: C408 Unnecessary list call - rewrite as a literal.
```

### Actual behavior

No diagnostics are shown

### Debug log

null-ls log: /home/hugo/.cache/nvim/null-ls.log
Detected filetype: python

3 source(s) active for this buffer:

Diagnostics: flake8
Formatting: isort, black

### Help

### Implementation help

It's my understanding that null-ls tries to match the error code as defined here:

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/deff230ec0d19f59e3a71f8ad3de34fef938c8f6/lua/null-ls/builtins/diagnostics.lua#L312-L315

The error codes for plugins all use letters _different_ from the ones flake8 itself uses, so ideally, we'd need to fall back to `error` if the letter is not on that map.

That map looks wrong anyway though, it maps `F` errors to `info`, but those are sometimes fatal errors, for example, calling an undefined function is `F821: undefined name 'nonexistant'`. (it seems that `F` contains a  mix of `Formatting` and `Fatal`.

jbarap:
Resolves #196.

As discussed in the issue, flake8 is currently constrained to match only a subset of the possible error codes, this PR changes the regex to match any letter, thus allowing error codes from extensions that prefix the code by a single capital letter.

This also introduces a special severity key "_fallback", that specifies the severity level that should be used in case the error code isn't found on the user specified or default severity mappings. The default fallback value is "error", just as it was before the pr, but now the user can choose on a per-linter basis how this is handled.

I'm not as familiar with the project, so the way the fallback key was implemented may have other side effects. If this is the case, please feel free to scrap the PR in favor of another implementation or edit it accordingly.

### DIFF
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -376,7 +376,6 @@ M.diagnostics = (function()
         ["warning"] = 2,
         ["information"] = 3,
         ["hint"] = 4,
-        ["_fallback"] = 1,
     }
 
     local default_json_attributes = {
@@ -436,7 +435,7 @@ M.diagnostics = (function()
         for attr, adapter in pairs(attr_adapters) do
             entries[attr] = adapter(entries, content_line) or entries[attr]
         end
-        entries["severity"] = entries["severity"] or default_severities["_fallback"]
+        entries["severity"] = entries["severity"] or default_severities["error"]
 
         -- Unset private attributes
         for k, _ in pairs(entries) do


JA-Bar:
> How can I indicate to from_pattern, that all unmatched codes/severities should fall back to something else than info?

This behavior is now possible by passing the key `["_fallback"]` to `overrides.severities` in the `overrides` argument of `from_pattern`. So we can do:

```lua
    [[:(%d+):(%d+): ((%u)%w+) (.*)]],
    { "row", "col", "code", "severity", "message" },
    {
        severities = {
            E = h.diagnostics.severities["error"],
            W = h.diagnostics.severities["warning"],
            F = h.diagnostics.severities["information"],
            D = h.diagnostics.severities["information"],
            ["_fallback"] = h.diagnostics.severities["warning"],
            },
        }
    ),
```
If the user wanted unspecified codes to fall in a warning severity.

As for the issue of mixing formatting and fatal:

> That map looks wrong anyway though, it maps F errors to info, but those are sometimes fatal errors, for example, calling an undefined function is F821: undefined name 'nonexistant'. (it seems that F contains a mix of Formatting and Fatal.

I believe this is a flake8 thing, it uses the same code prefix (F) for minor formatting issues and for fatal errors. To fix this we could look at the range of numbers that follow the code letter and assign the ranges to either an `error` or an `information` severity, but I believe that would be another PR. Right now I left the default F error codes severity as it was, but we could change it for now to `error` to reflect that those codes may contain fatal errors.

jose-elias-alvarez:
Understood, thanks for the description! I'll merge this once you've reverted the `default_severities` change.

jose-elias-alvarez:
Great, thank you!

JA-Bar:
Oh sorry! did I miss something?

yujinyuz:
Thanks @jose-elias-alvarez it's working now.

Can we somehow configure the fallback for a formatter/linter? Coz I feel like it's more of a warning than an error

PR #257, opened 2021-10-13
Add option to pass function to set process cwd

### LINKED ISSUE #256: Set Cwd for Spawn
### Issues

### Feature description

Allowing the option to set the `cwd` for spawning the process would add support for tools that rely on a directory other than the project root.

For my context:
I am working with a monorepo, and one of the included projects is an Elixir application. That application is an umbrella project, meaning there is a root application with sub-applications. To run the linter, `mix credo`, it must be run from the root of the Elixir umbrella application, not the project root and not the closest "application" when moving upwards.

The `cwd` is currently constructed in the `generator_factory` as `cwd = client and client.config.root_dir or vim.fn.getcwd()`.

Because of the dynamic nature, the most flexible approach seems to be allowing a function option taking the root and params that can return the cwd to pass to spawning the process.

### Help

Yes

### Implementation help

I have a PoC working locally with essentially this

```lua
local root = client and client.config.root_dir or vim.fn.getcwd()
local spawn_opts = {
    cwd = opts.cwd and opts.cwd(root, params) or root,
```

Assuming the approach is acceptable, I can make a PR for it while still working on the diagnostics separately.

aoswalt:
Closes #256

### DIFF
--- a/doc/HELPERS.md
+++ b/doc/HELPERS.md
@@ -38,6 +38,7 @@ helpers.generator_factory({
     to_temp_file, -- boolean (optional)
     use_cache, -- boolean (optional)
     runtime_condition, -- function (optional)
+    cwd, -- function (optional)
 })
 ```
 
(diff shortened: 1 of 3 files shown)


polyzen:
Hoping this is the solution to get selene to work with a config in a subdir. 🤞

PR #275, opened 2021-10-21
fix(builtins): Make finding credo cwd smarter

aoswalt:
Make the cwd smarter by searching upwards from the buffer but limited to
the project root and then trying to find the most specific match. It
should prevent excessive searches and erroneous other matches like
adjacent projects.

Switching to `json_raw` allows for silencing errors in projects that
do not use creo, but any other errors should still be reported.

Also use stdin to get more rapid feedback when changing files.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -257,7 +257,8 @@ local sources = { null_ls.builtins.formatting.cmake_format }
 
 Static analysis for `elixir` files for enforcing code consistency.
 
-- Searches _downward_ from the project root, so it should work for most projects, including umbrella projects with a single credo config. The `cwd` parameter may be overwritten for other project structures such as to search upwards for umbrella projects with per-application configurations.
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
Two questions:

1. nvim-lspconfig has some utilities for finding paths based on patterns. I see [here](https://github.com/neovim/nvim-lspconfig/blob/6882cfe458388b6e3db6b77fc1b368f635a9a6af/lua/lspconfig/elixirls.lua#L9) that `elixir-ls` uses one of those utilities, so would that solve this case, too?
2. I'm not sure that suppressing the error is the right approach, since ideally we don't want the source to be registered at all if it can't run. At the moment, the recommended approach would be to use `condition` to determine whether to register the source (for example, I set up a condition for ESLint to check for a config file). Is there a specific file or other marker that can tell us if the command will run successfully?

PR #281, opened 2021-10-24
fix(logger): configurable logging behavior

### LINKED ISSUE #280: Show error messages when command failed?
### Issues

### Feature description

I wish there were some way to display error messages when underlying command (e.g., formatter) failed. Currently, when something goes wrong and the formatter fails, nothing happens silently.

For example, I had an older version of `isort` (or in case `isort` is not installed -- which I think is a very common failure case) which exited with non-zero-exit code, after the formatting request `:lua vim.lsp.buf.formatting_sync()`.

By turning on the debug mode I was able to see the log message and figured out why the formatter was failing.
```
[DEBUG Sat Oct 23 18:39:25 2021] .../.vim/plugged/null-ls.nvim/lua/null-ls/utils.lua:51: error output: /usr/local/miniconda3/lib/python3.7/site-packages/setuptools/distutils_patch.py:26: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  "Distutils was imported before Setuptools. This usage is discouraged "
Traceback (most recent call last):
  File "/usr/local/miniconda3/bin/isort", line 11, in <module>
    sys.exit(main())
  File "/usr/local/miniconda3/lib/python3.7/site-packages/isort/main.py", line 343, in main
    SortImports(file_=file_, write_to_stdout=True, **arguments)
TypeError: type object got multiple values for keyword argument 'write_to_stdout'
```

To make troubleshooting even more convenient, can we detect such failure cases when the prettier/linter program has failed? If we could get the error message without turning on the debug mode, that'd be even better. Thanks!

kylo252:
- Set the correct stack level in plenary.log
- Use all the levels of logging with the ability of using the console
- Switch `echo_error` to use `Log:warn()`
- Switch most of `u.debug_log()` to `Log:trace()` while keeping some of them as `Log:debug()`

Fixes #280

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1426,7 +1426,7 @@ A linter for the `javascript` ecosystem.
 - To use local (project) executables, use the integration in
   [nvim-lsp-ts-utils](https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils)
   or try the [ESLint language
-  server](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#eslint).
+  server](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#eslint).
 
 ##### Usage
 


kylo252:
@jose-elias-alvarez, anything I can do to make integrating this easier?

lua/null-ls/config.lua
```diff
@@ -6,7 +6,18 @@ local defaults = {
...
     debug = false,
+    log = {
+        ---@usage disable logging completely
+        enable = true,
+        ---@usage set logging level
+        --- possible valuse: {"error", "warn", "info", "debug", "trace"}
+        level = "warn",
+        ---@usage whether to print the output to neovim's
+        --- possible values: 'sync','async', false
+        use_console = "sync",
```
jose-elias-alvarez, line 19: I like the option to use the console, but it's a surprising change from the current behavior, so I think it should be off by default.

lua/null-ls/diagnostics.lua
```diff
@@ -58,8 +59,8 @@ M.handler = function(original_params)
         params = params,
         postprocess = postprocess,
         callback = function(diagnostics)
-            u.debug_log("received diagnostics from generators")
-            u.debug_log(diagnostics)
+            Log:trace("received diagnostics from generators")
```
jose-elias-alvarez, line 62: This should be `Log:debug` to be consistent with the other methods, right?

lua/null-ls/info.lua
```diff
@@ -30,7 +31,7 @@ M.show_window = function()
 
     local bufnr = api.nvim_get_current_buf()
     if not client or not lsp.buf_is_attached(bufnr, client.id) then
-        u.echo("WarningMsg", "failed to get info: buffer is not attached")
+        Log:warn("failed to get info: buffer is not attached")
```
jose-elias-alvarez, line 34: I think this should always output to `messages` regardless of the user's log settings, since it's easy to imagine a situation where the command fails silently.

lua/null-ls/init.lua
```diff
@@ -30,6 +30,7 @@ M.config = function(user_config)
     require("null-ls.handlers").setup()
 
     vim.cmd("command! NullLsInfo lua require('null-ls').null_ls_info()")
+    vim.cmd("command! NullLsLog lua vim.fn.execute('edit ' .. require('null-ls.logger').get_path())")
```
jose-elias-alvarez, line 33: Great idea!

jose-elias-alvarez (commented):
Sorry for the delayed review. I like a lot of the changes here, but I think I need to know more about the intended use cases for this level of configurability. As far as I know, these are the ways in which people are using logging now:

1. Users who believe (correctly or incorrectly) that something is broken. In these cases, we want to give them as much information as possible to either solve it on their own or help us figure it out. 
2. Contributors who are working on a new built-in or fixing an existing one and want to check output. Here, they're only interested in the output, which corresponds to the `trace` level in this PR, so they're getting everything anyways.
3. Contributors who are working on a new feature. They may prefer to use `print` statements, but if they use logging at all, they want to see everything.

 Basically: It seems to me that if you're looking at the log in the first place, you probably want to see everything in there, so what do you envision as the use case here?

Also, I don't know if I agree with moving the current warnings into the log. When a user registers a source and runs the corresponding LSP command, they expect it to work, and if it doesn't I prefer to give immediate feedback about that regardless of their log settings. Likewise, errors are critical enough that we always want to show them, so I don't know if I see the point of adding them into the log.

kylo252:
> 1. Users who believe (correctly or incorrectly) that something is broken. In these cases, we want to give them as much information as possible to either solve it on their own or help us figure it out.

That's exactly my aim here. I want to convert as much as I can from the prints and echos into log messages.

> 2. Contributors who are working on a new built-in or fixing an existing one and want to check output. Here, they're only interested in the output, which corresponds to the `trace` level in this PR, so they're getting everything anyways.

Correct.

> 3. Contributors who are working on a new feature. They may prefer to use `print` statements, but if they use logging at all, they want to see everything.

I'm not sure if you're unaware of this, but the logger is actually using the console. Furthermore, the logging functions offer more flexibility than a normal print statement would. 

> Also, I don't know if I agree with moving the current warnings into the log. When a user registers a source and runs the corresponding LSP command, they expect it to work, and if it doesn't I prefer to give immediate feedback about that regardless of their log settings. 

The logger is always set to `warn` by default, so the warnings will always show, and they're immediate since I'm using the console in "sync" mode. But you also have the flexibility of log-level to `info` or `debug` level, and then monitoring either the console or the file in the case of a more involved functionality that is not behaving as expected.

> Likewise, errors are critical enough that we always want to show them, so I don't know if I see the point of adding them into the log.

This is actually critical to your very first point about _helping us figure it out_. It's much easier when you can tell them to show you the log, rather than asking them to wait and/or copy the error lines from their console. What if neovim froze/crashed for some reason?

> Basically: It seems to me that if you're looking at the log in the first place, you probably want to see everything in there, so what do you envision as the use case here?

I think there's a middle ground between seeing a lot of useful info vs actually seeing too much. I moved what you had as `debug`  to `trace` because I don't think a regular user is exactly interested in the content of the RPC exchanges, not to mention it was flooding the logfile way too fast. However, I still kept that option since it's useful for certain situations.

lua/null-ls/code-actions.lua
```diff
@@ -28,8 +29,8 @@ M.handler = function(method, original_params, handler)
             params = params,
             postprocess = postprocess,
             callback = function(actions)
-                u.debug_log("received code actions from generators")
-                u.debug_log(actions)
+                Log:trace("received code actions from generators")
```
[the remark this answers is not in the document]
kylo252: I'm using this https://github.com/luarocks/lua-style-guide#variable-names
> When doing OOP, classes should use `CamelCase`.

lua/null-ls/config.lua
kylo252: If you set it to `false` then you won't be able to see the warnings. This should mirror the default behavior unless you're changing the log-level, in which case you're free to use the console or not.

lua/null-ls/logger.lua
```diff
@@ -1 +1,71 @@
...
+        -- plenary uses lower-case log levels
+        self.__handle[level:lower()](msg)
+        return
+    end
+    local status_ok, plenary = pcall(require, "plenary")
+    local c = require("null-ls.config")
+    local log_level = "warn"
+    if c.get().debug then
+        --- "trace" level takes precdence
+        log_level = c.get().log.level and "trace" or "debug"
```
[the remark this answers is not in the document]
kylo252: By default, `config.debug = false`, so this branch wouldn't even execute until you set it to `true` for [legacy reasons](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/281#issuecomment-951169885).

lua/null-ls/diagnostics.lua
kylo252: This gets triggered waaaay too often to be useful. That's why I moved it to `trace` so it's not slowing down everything.

lua/null-ls/info.lua
kylo252: A user would need to explicitly turn off the console option, and we can even avoid exposing that option altogether.

By the way, the same logger logic is used by Packer, Telescope and quite a few others.

lua/null-ls/code-actions.lua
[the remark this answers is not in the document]
jose-elias-alvarez: If you see the examples [further down](https://github.com/Olivine-Labs/lua-style-guide#naming-conventions) I think that's referring to the class itself. This and [other style guides](https://github.com/Olivine-Labs/lua-style-guide#naming-conventions) use lowercase instances from what I can tell.

lua/null-ls/diagnostics.lua
jose-elias-alvarez: That's a fair argument, but I'm not a fan of the inconsistency.

lua/null-ls/completion.lua
```diff
@@ -12,8 +13,7 @@ M.handler = function(method, original_params, handler)
             method = methods.map[method],
             params = params,
             callback = function(results)
-                u.debug_log("received completion results from generators")
-                u.debug_log(results)
+                Log:debug("received completion results from generators" .. vim.inspect(results))
```
jose-elias-alvarez, line 16: This should also be split into one statement to confirm that results were received and another statement for the results themselves, right?

lua/null-ls/helpers.lua
```diff
@@ -159,10 +160,12 @@ M.generator_factory = function(opts)
             opts.command = command
         end
 
-        assert(
-            vim.fn.executable(command) > 0,
-            string.format("command %s is not executable (make sure it's installed and on your $PATH)", command)
-        )
+        if vim.fn.executable(command) ~= 1 then
```
jose-elias-alvarez, line 163: This path needs to either throw an error or otherwise prevent the generator from running, since attempting to run non-executable commands causes unpredictable behavior and is harder to catch.

jose-elias-alvarez:
I don't know if we're going to completely see eye-to-eye on this, but I do think there's a lot of useful stuff here, so I'll break down exactly what I would want to merge this: 

1. Moving the detailed description of log configuration out of the basic FAQ to avoid confusion
2. No logging at all by default
3. Logging defaults to a file, not the console
4. Always output critical warnings and errors to the console, regardless of the user's settings

If any of those is a deal-breaker, I'd consider another solution like allowing users to inject their own logger.

lua/null-ls/code-actions.lua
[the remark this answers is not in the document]
kylo252: It's fine with me either way, let's make it lower case if you feel it works better with your current code-format.

lua/null-ls/helpers.lua
kylo252: We probably need to reverse the logic here:
```lua
if not _validated then
    validate_opts(params)
end
```
But I'm still not sure how to interrupt the fn function., or did you mean we should just throw an `assert`?

kylo252:
Github is not happy about me using the default branch name, I don't even know how I did that 😄 
I'm just going to create another PR.

PR #288, opened 2021-10-27
perf: modularize built-ins

jose-elias-alvarez:
The library of built-ins has grown beyond what I imagined. Currently, built-ins are modularized by method, which means that requiring and registering a single built-in runs `helpers.make_builtin` once for every other built-in for the same method. The performance impact isn't huge, but it'll naturally grow bigger as more built-ins are added, and working with built-in files is also becoming tedious.

This PR lays out a simple structure to completely modularize built-ins using metatables. Finishing this will require a couple of hours of copy-pasting but shouldn't be too difficult. 

- [x] Formatting
- [x] Diagnostics
- [x] Code actions
- [x] Hover
- [x] Completion
- [x] Update documentation


doc/BUILTINS.md
```diff
@@ -1018,7 +1039,7 @@ local sources = { null_ls.builtins.formatting.shfmt }
 
 ##### Defaults
 
-- `filetypes = { "sh" }`
+- `filetypes = { "sh", "zsh" }`
```
gegoune, line 1042: I am not sure if shfmt supports zsh actually.
https://github.com/mvdan/sh/issues/120

jose-elias-alvarez: That's strange – I've been using `shfmt` for `zsh` since we added it about 2 weeks ago and haven't run into any problems. I don't know much about the project, but maybe the issue you linked is about a different feature?

gegoune: Hmm, not so sure, I just tried it myself for the first time and it reformatted:
```zsh
if (( $+commands[ls] )); then
```
to 
```zsh
if (($ + commands[ls])); then
```

which is not valid and errors with `bad math expression`. There is no mention of official support for zsh in shift's repository, so I would recommend not setting it as default option here.

jose-elias-alvarez: Got it – I'll remove it from the defaults.

gegoune: Thanks, amazing refactor btw, excellent work on null-ls, using it every day!

PR #301, opened 2021-10-30
feat: Initial implementation for completion support

Furkanzmc:
Just creating a PR to get some early feedback. I thought I could base the completion feature on the way hover works.
I'm not yet so sure about extending the `generator_opts` to support external processes for completion yet. I want to understand your point of view on this feature before investing too much time in the wrong direction.

I started testing it with my own config [here](https://github.com/Furkanzmc/dotfiles/blob/master/vim/lua/vimrc/null_ls_sources.lua#L158).

I want to express my appreciation for this plug-in again. It's awesome work! I'm looking forward to pushing it to the limits.

Let me know what you think of my changes and the direction you want to take.

-----

I didn't add any tests yet. I'll do that once I get your feedback.

-----

**Q_1**: Is there a way to report the results as they become available instead of waiting for all the generators to finish running?
**Q_2**: I'm thinking of adding the ability to support external sources for completion from nvim-cmp. I tried the buffer sources by just copying it to my config and it works well without any changes. The only trouble is when cmp-buffer repository is added to the plugins, it'll look for `cmp` module and it is possible it doesn't exist. Do you have any ideas about this? Do you think it makes sense?

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1903,6 +1903,11 @@ If you are using a completion plugin, they can also leverage these custom comple
 local sources = { null_ls.builtins.completion.spell }
 ```
 
-##### Defaults
+If you want to disable spell suggestions when `spell` options is not set, you can use the
+following snippet:
 
-- `null_ls.builtins.completion.spell` requires `spell` option to be set.
+```lua
+runtime_condition = function(_)
+    return vim.opt_local.spell:get()
+end
+```
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1887,23 +1887,29 @@ local sources = { null_ls.builtins.hover.dictionary }
 
 ### Completion
 
+#### Tags
+
 ##### About
 
-Completion sources provide an easy way for you to add custom completion to your project. When
-Neovim runs `omnifunc`, the null-ls will return completion item from registered sources.
+Tags source for completion. Only works if you have `tags` options set.
 
-If you are using a completion plugin, they can also leverage these custom completion sources.
-
-##### Usage
-
-- Proof-of-concept for completion functionality.
-- See the completion section of [the documentation](MAIN.md) for details.
+###### Usage
 
 ```lua
-local sources = { null_ls.builtins.completion.spell, null_ls.builtins.completion.tags }
+local sources = { null_ls.builtins.completion.tags }
 ```
 
-Tags source won't be available if you don't have the `tags` option set for Neovim,
+#### Spell
+
+##### About
+
+Spell suggestions completion source.
+
+###### Usage
+
+```lua
+local sources = { null_ls.builtins.completion.spell }
+```
 
 If you want to disable spell suggestions when `spell` options is not set, you can use the
 following snippet:
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1884,3 +1884,26 @@ local sources = { null_ls.builtins.hover.dictionary }
 ##### Defaults
 
 - `filetypes = { "txt", "markdown" }`
+
+### Completion
+
+##### About
+
+Completion sources provide an easy way for you to add custom completion to your project. When
+Neovim runs `omnifunc`, the null-ls will return completion item from registered sources.
+
+If you are using a completion plugin, they can also leverage these custom completion sources.
+
+##### Usage
+
+- Proof-of-concept for completion functionality.
+- See the completion section of [the documentation](MAIN.md) for details.
+
+```lua
+local sources = { null_ls.builtins.completion.tags, null_ls.builtins.completion.spell }
+```
+
+##### Defaults
+
+- `null_ls.builtins.completion.tags` requires `tags` option to be set.
+- `null_ls.builtins.completion.spell` requires `spell` option to be set.
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1900,10 +1900,9 @@ If you are using a completion plugin, they can also leverage these custom comple
 - See the completion section of [the documentation](MAIN.md) for details.
 
 ```lua
-local sources = { null_ls.builtins.completion.tags, null_ls.builtins.completion.spell }
+local sources = { null_ls.builtins.completion.spell }
 ```
 
 ##### Defaults
 
-- `null_ls.builtins.completion.tags` requires `tags` option to be set.
 - `null_ls.builtins.completion.spell` requires `spell` option to be set.
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1900,8 +1900,10 @@ If you are using a completion plugin, they can also leverage these custom comple
 - See the completion section of [the documentation](MAIN.md) for details.
 
 ```lua
-local sources = { null_ls.builtins.completion.spell }
+local sources = { null_ls.builtins.completion.spell, null_ls.builtins.completion.tags }
 ```
+
+Tags source won't be available if you don't have the `tags` option set for Neovim,
 
 If you want to disable spell suggestions when `spell` options is not set, you can use the
 following snippet:
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -27,7 +27,8 @@ M.tags = h.make_builtin({
             for _, tag in ipairs(tags) do
                 table.insert(words, tag.name)
             end
-            words = table.uniq(words)
+
+            words = utils.table.uniq(words)
             for _, word in ipairs(words) do
                 table.insert(items, {
                     label = word,
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -21,7 +21,7 @@ M.spell = h.make_builtin({
                 return items
             end
 
-            done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = true })
+            done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = false })
         end,
         async = true,
     },
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -1,0 +1,36 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+local utils = require("null-ls.utils")
+
+local COMPLETION = methods.internal.COMPLETION
+
+local M = {}
+
+M.tags = h.make_builtin({
+    method = COMPLETION,
+    filetypes = {},
+    name = "tags",
+    generator = {
+        fn = function(params, done)
+            local tags = vim.fn.taglist(params.word_to_complete)
+            local words = {}
+            local items = {}
+            for _, tag in ipairs(tags) do
+                table.insert(words, tag.name)
+            end
+            words = utils.tbl_uniq(words)
+            for _, word in ipairs(words) do
+                table.insert(items, {
+                    label = word,
+                    insertText = word,
+                })
+            end
+
+            done(items)
+        end,
+        async = true,
+        use_cache = true,
+    },
+})
+
+return M
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -6,36 +6,6 @@ local COMPLETION = methods.internal.COMPLETION
 
 local M = {}
 
-M.tags = h.make_builtin({
-    method = COMPLETION,
-    filetypes = {},
-    name = "tags",
-    runtime_condition = function(_)
-        return #vim.fn.tagfiles() > 0
-    end,
-    generator = {
-        fn = function(params, done)
-            local tags = vim.fn.taglist(params.word_to_complete)
-            local words = {}
-            local items = {}
-            for _, tag in ipairs(tags) do
-                table.insert(words, tag.name)
-            end
-            words = utils.tbl_uniq(words)
-            for _, word in ipairs(words) do
-                table.insert(items, {
-                    label = word,
-                    insertText = word,
-                })
-            end
-
-            done(items)
-        end,
-        async = true,
-        use_cache = true,
-    },
-})
-
 M.spell = h.make_builtin({
     method = COMPLETION,
     filetypes = {},
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local utils = require("null-ls.utils")
 
 local COMPLETION = methods.internal.COMPLETION
 
@@ -16,9 +17,15 @@ M.tags = h.make_builtin({
     },
     generator = {
         fn = function(params, done)
+            -- Tags look up can be expensive.
+            if #params.word_to_complete < 4 then
+                done({ { items = {}, isIncomplete = false } })
+                return
+            end
+
             local tags = vim.fn.taglist(params.word_to_complete)
             if tags == 0 then
-                done({ items = {}, isIncomplete = false })
+                done({ { items = {}, isIncomplete = false } })
                 return
             end
 
@@ -36,7 +43,7 @@ M.tags = h.make_builtin({
                 })
             end
 
-            done({ items = items, isIncomplete = false })
+            done({ { items = items, isIncomplete = #items == 0 } })
         end,
         async = true,
     },
@@ -48,7 +55,7 @@ M.spell = h.make_builtin({
     name = "spell",
     generator = {
         fn = function(params, done)
-            local candidates = function(entries)
+            local get_candidates = function(entries)
                 local items = {}
                 for k, v in ipairs(entries) do
                     items[k] = { label = v, kind = vim.lsp.protocol.CompletionItemKind["Text"] }
@@ -57,7 +64,8 @@ M.spell = h.make_builtin({
                 return items
             end
 
-            done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = false })
+            local canditates = get_candidates(vim.fn.spellsuggest(params.word_to_complete))
+            done({ { items = canditates, isIncomplete = #canditates } })
         end,
         async = true,
     },
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -27,7 +27,6 @@ M.spell = h.make_builtin({
             done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = true })
         end,
         async = true,
-        use_cache = true,
     },
 })
 
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -1,10 +1,45 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
-local utils = require("null-ls.utils")
 
 local COMPLETION = methods.internal.COMPLETION
 
 local M = {}
+
+M.tags = h.make_builtin({
+    method = COMPLETION,
+    filetypes = {},
+    name = "tags",
+    generator_opts = {
+        runtime_condition = function(_)
+            return #vim.fn.tagfiles() > 0
+        end,
+    },
+    generator = {
+        fn = function(params, done)
+            local tags = vim.fn.taglist(params.word_to_complete)
+            if tags == 0 then
+                done({ items = {}, isIncomplete = false })
+                return
+            end
+
+            local words = {}
+            local items = {}
+            for _, tag in ipairs(tags) do
+                table.insert(words, tag.name)
+            end
+            words = table.uniq(words)
+            for _, word in ipairs(words) do
+                table.insert(items, {
+                    label = word,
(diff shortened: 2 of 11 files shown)


jose-elias-alvarez:
This is pretty cool! I think before I comment more on this, I'd need to know about what sort of use cases you have in mind for this. Pure Lua completion is probably best handled via plugins like nvim-cmp, and if the goal is to capture the output of spawned processes, I think we'd have to be careful about performance. 

To answer your questions for now:

1. We don't have an official way to do this, but I imagine you could hijack `postprocess` for now. If it's useful, we could add this as a proper option.
2. It looks like this is how nvim-cmp sources are structured, so besides a PR to conditionally require `cmp` (which I doubt would be accepted), you could do something like this, but it's pretty dirty:

```lua
-- run before requiring the nvim-cmp source
local ok, cmp = pcall(require, "cmp")
package.loaded["cmp"] = ok and cmp or { register_source = function() end }
```

Furkanzmc:
> This is pretty cool! I think before I comment more on this, I'd need to know about what sort of use cases you have in mind for this.

I love null-ls because I can extend the LSP capabilities with my own custom needs without much effort. It's standard since it implements LSP, and If I switch from one completion plugin to the other my custom completion code will not have to change as long as I have null-ls.

The trouble with the completion plugins is that they all reinvent the wheel in their own way, and the source for one of those plugins is useless for the other. If I have my own completion functions, I don't want it to break when I switch plugins. This is really a lost potential for those plugins and a lot of duplicated effort. I prefer basing it on a constant that doesn't change when I change plugins. Which is why I was using efm-langserver before. I implemented my own hover and completion sources, but since it's an external process I'm limited in what I can do.

I'd be preaching the choir here, but null-ls is superior to efm-langserver *because* it has direct access to Neovim.

I think completion itself should not be implementation dependant now that we have LSP. Completion sources can be implemented with null-ls by external plugin authors and these sources would be useful whatever plugin you use. In the short term, this gives users a very easy way to add custom LSP completion (like the list of Git authors in their team, or in my case I wrote a function that completes based on mnemonics). In the long run, I think null-ls can be used as the base completion "engine" and plugin authors can extend null-ls's capabilities with enter caching, pop-ups, documentations etc.

I'll give a personal example, I don't use any completion plugins. I rely on LSPs and ins-completion for my workflow (and I'm not the only one as evident by the popularity of supertab.vim and vim-mucomplete). But sometimes I come across some good completion sources, but they are useless to me because I don't use the dependant plugin. But if those sources were implemented in null-ls, I would be able to use it easily. And the plugin authors would still benefit from it because that source's results come directly from an LSP (null-ls).

> Pure Lua completion is probably best handled via plugins like nvim-cmp,

I think completion plug-ins should handle customization of completion options like popups, auto complete etc.

> and if the goal is to capture the output of spawned processes, I think we'd have to be careful about performance.

I actually don't think we'd benefit from having this feature, but I was thinking of adding for sake of uniformity with the other sources (format, hover, etc)

> 1. We don't have an official way to do this, but I imagine you could hijack `postprocess` for now. If it's useful, we could add this as a proper option.

Thanks for the explanation. I think this would be very useful but I wouldn't want to add it as part of this PR (If you want to move forward with it).

> 2. It looks like this is how nvim-cmp sources are structured, so besides a PR to conditionally require `cmp` (which I doubt would be accepted), you could do something like this, but it's pretty dirty:

Just to be clear, I was just thinking of this as a helper method to just enable a smooth transition from other completion sources to this one. It's not really a priority for this PR.

jose-elias-alvarez:
I see! The idea makes a lot of sense, and I think there's definitely room in the ecosystem for a lighter alternative to a full-on completion plugin that makes it easier to create / register additional sources. 

I'm happy to review and merge this whenever it's done. FYI, I briefly tested out an external source and it seems to be working fine, so I don't think there's any additional work needed to implement it:

```lua
M.test = h.make_builtin({
    method = COMPLETION,
    filetypes = {},
    name = "test",
    generator = h.generator_factory({
        command = "ls",
        args = { vim.fn.expand("$HOME") },
            local items = {}
            for _, item in ipairs(vim.split(params.output, "\n")) do
                if item ~= "" then
                    table.insert(items, { label = item, kind = vim.lsp.protocol.CompletionItemKind["File"] })
                end
            end
            done(items)
        end,
    }),
})
```

Furkanzmc:
> I see! The idea makes a lot of sense, and I think there's definitely room in the ecosystem for a lighter alternative to a full-on completion plugin that makes it easier to create / register additional sources. 

I'll use this source for a while to see how robust it is for me and then I'll contact some of the completion source plugin developers to see if they'd be open to supporting null-ls for this. Or I'll just fork and do it myself. :D

 
> I'm happy to review and merge this whenever it's done. FYI, I briefly tested out an external source and it seems to be working fine, so I don't think there's any additional work needed to implement it:

I'm extremely new to the source code, so I just implemented it the way I saw being the most straightforward. If you are OK with the current implementation, I have nothing more to add to it. I don't want to turn it into a big PR.

I can add anything else that come after with a separate PR.

What do you think?

jose-elias-alvarez:
Sure, works for me. Before merging, we should add tests (the ones for the hover method should provide a good example) and maybe another built-in source that doesn't rely on tags so other users can more easily try out the new functionality (and hopefully get inspiration to create their own sources). A toy source like the `dictionary` hover source would be great. Let me know if you have any ideas.

lua/null-ls/utils.lua
```diff
@@ -119,6 +119,13 @@ M.make_params = function(original_params, method)
         params.range = M.range.from_lsp(original_params.range)
     end
 
+    if params.lsp_method == methods.lsp.COMPLETION then
+        local pos = vim.api.nvim_win_get_cursor(0)
```
jose-elias-alvarez, line 123: Could we reuse the previous declaration of `pos`?

jose-elias-alvarez:
I played around with this a bit more locally and it seems to be working well for `omnifunc` completion but is broken with nvim-cmp. Each keystroke throws this error, which is coming from [this line](https://github.com/hrsh7th/cmp-nvim-lsp/blob/accbe6d97548d8d3471c04d512d36fa61d0e4be8/lua/cmp_nvim_lsp/source.lua#L82) in the cmp-nvim-lsp source:

```
Error detected while processing TextChangedI Autocommands for "*":
E5108: Error executing lua ...ck/packer/start/cmp-nvim-lsp/lua/cmp_nvim_lsp/source.lua:82: attempt to index local 'c' (a boolean value)
```

Since the goal (as I understand it) is to provide a plugin-agnostic completion option, we should probably check this out (and ideally check if other popular completion plugins work, though that's not a priority for this PR).

Furkanzmc:
> Since the goal (as I understand it) is to provide a plugin-agnostic completion option, we should probably check this out (and ideally check if other popular completion plugins work, though that's not a priority for this PR).

Makes sense. I tested it with nvim-cmp and coq_nvim to see if it works (https://github.com/jose-elias-alvarez/null-ls.nvim/pull/301/commits/83b599129a132bacc96bb7724f334a5aa39cc293 fixes the problem you reported.). It seems to work mostly fine. There's a problem where the plugins don't send a request to update the completion list when I enter new characters, but I'll figure that out with a different PR. I pushed some changes for early review. I'll look into the tests tomorrow.

Thanks for the feedback!

lua/null-ls/rpc.lua
```diff
@@ -8,6 +8,15 @@ local capabilities = {
     executeCommandProvider = true,
     documentFormattingProvider = true,
     documentRangeFormattingProvider = true,
+    completionProvider = {
+        -- FIXME: How do we decide what trigger characters are?
```
Furkanzmc, line 12: Maybe we can use a Vim option for this? I'm not sure about the right approach here. I'm guessing maybe this is causing the completion plugins to not send a request sometimes. @jose-elias-alvarez

lua/null-ls/builtins/completion.lua
```diff
@@ -0,0 +1,34 @@
...
+
+local M = {}
+
+M.spell = h.make_builtin({
+    method = COMPLETION,
+    filetypes = {},
+    name = "spell",
+    runtime_condition = function(_)
+        return vim.opt_local.spell:get()
+    end,
```
jose-elias-alvarez, line 15:
```suggestion
    generator_opts = {
        runtime_condition = function(_)
            return vim.opt_local.spell:get()
        end,
    },
```

This would need to be put inside `generator_opts` to work. Also. I'm not sure if we need to have this at all – at the moment, we don't have any built-ins with default conditions, so I think we could just mention this in the documentation as a (reasonable) suggestion.

lua/null-ls/builtins/completion.lua
```diff
@@ -0,0 +1,34 @@
...
+                    items[k] = { label = v, kind = vim.lsp.protocol.CompletionItemKind["Text"] }
+                end
+
+                return items
+            end
+
+            done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = true })
+        end,
+        async = true,
+        use_cache = true,
```
jose-elias-alvarez, line 30: `use_cache` only works for generators created using `generator_factory` (and would cache by command, so it wouldn't work in this case).

lua/null-ls/generators.lua
```diff
@@ -38,12 +39,20 @@ M.run = function(generators, params, postprocess, callback)
                     return
                 end
 
-                for _, result in ipairs(results) do
+                if params.method == methods.internal.COMPLETION then
```
jose-elias-alvarez, line 42: Is there a way to change the structure of `results` to avoid having to handle this as a special case?

lua/null-ls/rpc.lua
jose-elias-alvarez: Testing with nvim-cmp, the majority of these options don't seem to make a difference. Apart from `labelDetailsSupport`, I'm getting requests on every keystroke regardless of these values, so I think we can take them out for now.

jose-elias-alvarez (commented):
The `spell` source works well! I'm not seeing any issues with requests not getting sent – in fact, nvim-cmp sends so many requests that it causes slowdown when typing. What issues are you seeing with completion?

There's also a bad interaction with nvim-cmp where each keystroke sends a request *and* triggers a change in the document, which sends a `textDocument/didChange` notification and causes even more slowdown, regardless of `completeopt`. This is really strange, since a) the document is not actually changing, and b) these changes should be debounced by [this code in Neovim core](https://github.com/neovim/neovim/blob/5ce35abae6427dac7ac2b147bae8f16adcd8ddff/runtime/lua/vim/lsp.lua#L376). Until we figure this out, it might be worth mentioning it somewhere (and verify if the same thing happens with coq).

Also, could we add a mention of the new functionality on the main README.md page?

lua/null-ls/generators.lua
Furkanzmc: I could change it to `return { { items={}, isIncomplete=false } }` instead of `return { items={}, isIncomplete=false }` but I found that to be awkward. I could make the change if you insist. I think the second one is better just from a mindset perspective.

Furkanzmc:
> The `spell` source works well! I'm not seeing any issues with requests not getting sent – in fact, nvim-cmp sends so many requests that it causes slowdown when typing. What issues are you seeing with completion?

My bad. I was misinterpreting the behavior. When `isIncomplete` is `false`, it's pretty fast for tags too. Only becomes a problem if `isIncomplete` is `true` for each invocation.

> Also, could we add a mention of the new functionality on the main README.md page?

Will do.

lua/null-ls/rpc.lua
Furkanzmc: I'm getting problems with coq_nvim when I highlight a completion item. It unconditionally sends a `completionItem/resolve` request and we don't support that yet. I'll create an issue for this on coq_nvim so the author can do a conditional request instead.

lua/null-ls/rpc.lua
```diff
@@ -8,6 +8,15 @@ local capabilities = {
...
+        triggerCharacters = { ".", ":", "-" },
```
jose-elias-alvarez, line 13: So this seems to be working now in that requests are only sent when the trigger character matches, but as a result using sources like `spell` with nvim-cmp is pretty broken, since only the first typed letter will trigger completion. A potential solution would be to trigger on every character, then let each source optionally define a custom set of trigger characters. For example, the spell source could trigger only on letters. I don't know if that would be performant enough to provide a good experience, though.

I'm not sure what's the best solution here, but if we can't think of anything for now, I don't think we should specifically advertise compatibility with on-each-character completion plugins until we have this figured out.

lua/null-ls/utils.lua
```diff
@@ -119,6 +119,12 @@ M.make_params = function(original_params, method)
         params.range = M.range.from_lsp(original_params.range)
     end
 
+    if params.lsp_method == methods.lsp.COMPLETION then
+        local line = vim.api.nvim_get_current_line()
+        local line_to_cursor = line:sub(1, pos[2])
+        params.word_to_complete = line:sub(vim.fn.match(line_to_cursor, "\\k*$") + 1, params.col)
```
jose-elias-alvarez, line 125: We don't currently use any Vim regular expressions in the codebase, so if possible could we use a Lua string matcher here?

lua/null-ls/builtins/completion.lua
```diff
@@ -0,0 +1,65 @@
...
+                done({ items = {}, isIncomplete = false })
+                return
+            end
+
+            local words = {}
+            local items = {}
+            for _, tag in ipairs(tags) do
+                table.insert(words, tag.name)
+            end
+            words = table.uniq(words)
```
jose-elias-alvarez, line 30: This is undefined with LuaJIT – did you mean to use your `tbl_uniq` util instead?

lua/null-ls/utils.lua
```diff
@@ -164,6 +170,13 @@ M.table = {
         end
         return replaced
     end,
+    extend = function(source, target)
```
jose-elias-alvarez, line 173: Would `vim.tbl_extend` remove the need for this?

jose-elias-alvarez (commented):
Great work on this! 

It seems to me that on-demand completion is working well, but we still have some issues to work out when it comes to autocompletion. I don't think there's necessarily a need to solve those issues in this PR, but in that case, I think it would be best to clarify that currently, this is *only* meant to work with on-demand, `omnifunc`-style completion. Let me know what you think.

lua/null-ls/rpc.lua
Furkanzmc: > So this seems to be working now in that requests are only sent when the trigger character matches, but as a result using sources like `spell` with nvim-cmp is pretty broken, since only the first typed letter will trigger completion.

I think the reason you were experiencing that was because of `isIncomplete=false` for spell. If I set that to true, then we don't have the same problem. I can't produce a problem with auto complete using coq_nvim.

lua/null-ls/utils.lua
Furkanzmc: The documentation states that it works for map like tables. But I need this for array like tables. I don't think it will work for this case. Or maybe I'm not familiar enough with `tbl_extend`.

jose-elias-alvarez: In that case, `vim.list_extend` should work.

lua/null-ls/rpc.lua
jose-elias-alvarez: Alright, so it seems that we need to set `isIncomplete` based on the availability of candidates, since without it autocompletion won't actually work with nvim-cmp (performance issues aside). Does something like this make sense?

```diff
diff --git a/lua/null-ls/builtins/completion.lua b/lua/null-ls/builtins/completion.lua
index 237aad2..eb2cde3 100644
--- a/lua/null-ls/builtins/completion.lua
+++ b/lua/null-ls/builtins/completion.lua
@@ -47,7 +47,7 @@ M.spell = h.make_builtin({
     name = "spell",
     generator = {
         fn = function(params, done)
-            local candidates = function(entries)
+            local get_candidates = function(entries)
                 local items = {}
                 for k, v in ipairs(entries) do
                     items[k] = { label = v, kind = vim.lsp.protocol.CompletionItemKind["Text"] }
@@ -56,7 +56,11 @@ M.spell = h.make_builtin({
                 return items
             end

-            done({ items = candidates(vim.fn.spellsuggest(params.word_to_complete)), isIncomplete = false })
+            local candidates = get_candidates(vim.fn.spellsuggest(params.word_to_complete))
+            done({
+                items = candidates,
+                isIncomplete = #candidates > 0,
+            })
         end,
         async = true,
     },
```

Also, I haven't set up coq, but are you saying that autocompletion works there without `isIncomplete`?

lua/null-ls/completion.lua
```diff
@@ -0,0 +1,50 @@
...
+                u.debug_log(results)
+                if #results == 0 then
+                    handler({})
+                else
+                    local items = {}
+                    local isIncomplete = false
+                    for _, result in ipairs(results) do
+                        isIncomplete = isIncomplete or result.isIncomplete
+
+                        vim.validate({
```
jose-elias-alvarez, line 24: I think we can take out validation. This handler can run on every keystroke, so every millisecond counts, and the general rule we've followed so far is that sources are responsible for ensuring their return values match what's expected.

lua/null-ls/completion.lua
```diff
@@ -0,0 +1,50 @@
...
+                            items = { result.items, "table" },
+                            isIncomplete = { result.isIncomplete, "boolean" },
+                        })
+                        for index, item in ipairs(result.items) do
```
jose-elias-alvarez, line 28: Same thing re: validation here. Also, do we need to handle the case where `item` is a string given that the specification says it has to be an object? If not, we could just use `vim.list_extend` without iterating over the items at all.

lua/null-ls/rpc.lua
Furkanzmc: > Alright, so it seems that we need to set `isIncomplete` based on the availability of candidates, since without it autocompletion won't actually work with nvim-cmp (performance issues aside). Does something like this make sense?

I think this makes more sense for tags source. With spell, any new key can change the suggestions. But with tags, we'll get more or less the same items.

> Also, I haven't set up coq, but are you saying that autocompletion works there without `isIncomplete`?

It works with both if `isIncomplete` is set to `false` appropriately. I was just setting it to `true` before, and that was causing performance problems. And setting it to false all the time causes missing completion requests. So, what you suggest is the way to go.

Also, I'm not getting the error for `completionItem/resolve` any more. All seems to be working fine for now.

lua/null-ls/utils.lua
Furkanzmc: I switched to using `vim.regex`. Since `\k` uses `iskeyword` option, it's pretty easy to parse the line (And it could also be helpful for language specific options too). Otherwise I'd need to write a long string matcher and I'm not really good at those. Is this solution sufficient for you?

lua/null-ls/completion.lua
Furkanzmc: I put the validation under `config.debug` condition. Is that OK?

doc/BUILTINS.md
```diff
@@ -1884,3 +1884,32 @@ local sources = { null_ls.builtins.hover.dictionary }
 ##### Defaults
 
 - `filetypes = { "txt", "markdown" }`
+
+### Completion
```
jose-elias-alvarez, line 1888: I think the structure here doesn't match what we use for the rest of this document. We should remove (or move) the general description and instead just describe the two specific sources we're adding in this PR.

lua/null-ls/builtins/completion.lua
```diff
@@ -0,0 +1,74 @@
...
+            local get_candidates = function(entries)
+                local items = {}
+                for k, v in ipairs(entries) do
+                    items[k] = { label = v, kind = vim.lsp.protocol.CompletionItemKind["Text"] }
+                end
+
+                return items
+            end
+
+            local canditates = get_candidates(vim.fn.spellsuggest(params.word_to_complete))
```
jose-elias-alvarez, line 67:
```suggestion
            local candidates = get_candidates(vim.fn.spellsuggest(params.word_to_complete))
```

lua/null-ls/builtins/completion.lua
```diff
@@ -0,0 +1,74 @@
...
+            done({ { items = canditates, isIncomplete = #canditates } })
```
jose-elias-alvarez, line 68:
```suggestion
            done({ { items = candidates, isIncomplete = #candidates } })
```

lua/null-ls/completion.lua
```diff
@@ -0,0 +1,47 @@
...
+                u.debug_log(results)
+                if #results == 0 then
+                    handler({})
+                else
+                    local items = {}
+                    local isIncomplete = false
+                    for _, result in ipairs(results) do
+                        isIncomplete = isIncomplete or result.isIncomplete
+
+                        if config.debug then
```
jose-elias-alvarez, line 25: This is a good idea! Once this is merged I'll add something similar for the other methods.

jose-elias-alvarez (changes requested):
This looks great! Once we make the final changes to the documentation and fix a typo, this is ready to merge IMO.

lua/null-ls/builtins/completion.lua
Furkanzmc: Ironic, making spelling mistakes in the spelling source. :D

jose-elias-alvarez:
Awesome, thank you!

jose-elias-alvarez (approved):

PR #303, opened 2021-11-01
Pass `--stdin-filename` for black

bellini666:
Black accepts --stdin-filename since its 21.5b0 release: https://github.com/psf/black/releases/tag/21.5b0

### DIFF
--- a/lua/null-ls/builtins/formatting.lua
+++ b/lua/null-ls/builtins/formatting.lua
@@ -32,6 +32,8 @@ M.black = h.make_builtin({
     generator_opts = {
         command = "black",
         args = {
+            "--stdin-filename",
+            "$FILENAME",
             "--quiet",
             "-",
         },


jose-elias-alvarez:
I'm not familiar with the Python ecosystem, but if a user were on an old version of `black`, would it be simple for them to update it, or could it potentially cause conflicts with projects using an older version? Also, how do older versions behave when passed this flag? May 2021 is pretty recent, so I wouldn't want to cause issues.

bellini666:
Hey @jose-elias-alvarez ,

Actually it is a bit earlier than that, from https://github.com/psf/black/releases/tag/21.4b0 (it was me that contributed to that changed and I don't even remember lol https://github.com/psf/black/pull/1780)

But I know what you mean... If someone is using an old version of `black` it would give an error saying that `--stdin-filename` doesn't exist.

Black considers itself still in beta and thus, people should be using its latest version, but maybe we should wait for some more time to make it default here.

Maybe a hint on BUILTINS.md would be enough? Point there that it is recommended for the user to put both this and `--fast` at extra_args?

jose-elias-alvarez:
If it's been almost a year and `black` is explicitly in beta, then I think it's safe to add the argument. Thank you for clarifying and working on this!

PR #309, opened 2021-11-03
fix(builtins): let `fnlfmt` know which buffer to run on

kaile256:
Omit `--fix` flag, which modifies file directly, but doesn't buffer.


jose-elias-alvarez:
Yep, I can see that this was broken. Thank you for the fix!

PR #313, opened 2021-11-03
fix: check fileformat to determine newline char

### LINKED ISSUE #312: File with CRLF (line endings) breaks eslint_d
### FAQ

### Issues

### Neovim Version

0.5.1

### null-ls config

// from [LunarVim](https://www.lunarvim.org/languages/#at-a-glance)

local linters = require "lvim.lsp.null-ls.linters"
linters.setup {
  {
    exe = "eslint_d",
    filetypes = { "typescript", "javascript", "javascriptreact", "typescriptreact" },
  },
}

### Steps to reproduce

1. Setup eslint_d / eslint inside neovim config
2. Create a file with CRLF (windows file endings, \r\n)
3. File should be in the project folder with eslint config
4. Inside eslint config have this inside [rules](https://eslint.org/docs/rules/linebreak-style) - linebreak-style: ["error", "windows"]
5. Open file and see errors

### Expected behavior

No errors must be shown, as the file contains correct line endings.

### Actual behavior

 But for some reason, eslint says that line ending is LF (UNIX one)
image

### Debug log

probable fine (?)

### Help

### Implementation help

_No response_

jose-elias-alvarez:
Might fix #312. Needs tests if it does.

### DIFF
--- a/lua/null-ls/diff.lua
+++ b/lua/null-ls/diff.lua
@@ -1,0 +1,156 @@
+local M = {}
+
+local function first_difference(old_lines, new_lines)
+    local line_count = math.min(#old_lines, #new_lines)
+    if line_count == 0 then
+        return 1, 1
+    end
+
+    local start_line_idx
+    for i = 1, line_count do
+        start_line_idx = i
+        if old_lines[start_line_idx] ~= new_lines[start_line_idx] then
+            break
+        end
+    end
+
+    local old_line = old_lines[start_line_idx]
+    local new_line = new_lines[start_line_idx]
+    local length = math.min(#old_line, #new_line)
+
+    local start_col_idx = 1
+    while start_col_idx <= length do
+        if string.sub(old_line, start_col_idx, start_col_idx) ~= string.sub(new_line, start_col_idx, start_col_idx) then
+            break
+        end
+        start_col_idx = start_col_idx + 1
+    end
+
+    return start_line_idx, start_col_idx
+end
+
+local function last_difference(old_lines, new_lines, start_char)
+    local line_count = math.min(#old_lines, #new_lines)
+    if line_count == 0 then
+        return 0, 0
+    end
+
+    local end_line_idx = -1
+    for i = end_line_idx, -line_count, -1 do
+        if old_lines[#old_lines + i + 1] ~= new_lines[#new_lines + i + 1] then
+            end_line_idx = i
+            break
+        end
+    end
+
+    local old_line
+    local new_line
+    if end_line_idx <= -line_count then
+        end_line_idx = -line_count
+        old_line = string.sub(old_lines[#old_lines + end_line_idx + 1], start_char)
+        new_line = string.sub(new_lines[#new_lines + end_line_idx + 1], start_char)
+    else
+        old_line = old_lines[#old_lines + end_line_idx + 1]
+        new_line = new_lines[#new_lines + end_line_idx + 1]
+    end
+
+    local old_line_length = #old_line
+    local new_line_length = #new_line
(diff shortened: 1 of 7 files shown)


jose-elias-alvarez:
So the first fix I attempted works when the file first opens but fails as soon as it's changed because of [this line](https://github.com/neovim/neovim/blob/f26b391317281d0520056eaa8312cf25e654149f/runtime/lua/vim/lsp.lua#L274) in Neovim core. null-ls tries to get the file's content from the LSP params when it's available to save an API call, but until this is fixed upstream, the solution is to always get content manually.

The following Neovim diff fixes the issue:

```diff
diff --git a/runtime/lua/vim/lsp.lua b/runtime/lua/vim/lsp.lua
index 01e801415..20d9f650b 100644
--- a/runtime/lua/vim/lsp.lua
+++ b/runtime/lua/vim/lsp.lua
@@ -271,9 +271,10 @@ end
 ---@param bufnr (number) Buffer handle, or 0 for current.
 ---@returns Buffer text as string.
 local function buf_get_full_text(bufnr)
-  local text = table.concat(nvim_buf_get_lines(bufnr, 0, -1, true), '\n')
+  local eol_char = nvim_buf_get_option(bufnr, 'fileformat') == 'dos' and '\r\n' or '\n'
+  local text = table.concat(nvim_buf_get_lines(bufnr, 0, -1, true), eol_char)
   if nvim_buf_get_option(bufnr, 'eol') then
-    text = text .. '\n'
+    text = text .. eol_char
   end
   return text
 end
```

I'm not sure if a PR would be accepted since it doesn't handle `\r`, which [the specification](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocuments) also defines as a valid character. This is currently causing an identical issue for the ESLint language server, though, so it's worth a shot.

jose-elias-alvarez:
Still waiting for feedback but I'm pretty sure this is the right fix, and it's now waiting on neovim/neovim#16252 (I also adapted the implementation a bit based on the solution there). 

An consequence of merging this PR is that 0.5.0 and 0.5.1 will be *slightly* slower in the interest of correctness, since those versions will always get buffer content directly instead of pulling it from the LSP params when possible, but I think that's acceptable.

jose-elias-alvarez:
The linked PR was merged, a consequence of which was the removal of `vim.lsp.util.compute_diff`. This PR now includes an independent version of the old `compute_diff` utility, which does the job for now and closes #328.

Keep-Simple:
@jose-elias-alvarez  Hi, thanks for the work! So now it should work on nvim 0.5.x and 0.6.0?

jose-elias-alvarez:
@Keep-Simple null-ls no longer does any special handling of line endings thanks to neovim/neovim#16334, so everything should work as expected.

Keep-Simple:
Awesome, already tested and it's working! Can I donate you somehow?

jose-elias-alvarez:
Haha, don't worry about it, just glad to know it's working.

Keep-Simple:
Hey, sorry to disturb you again.
I really want to try your typescript utils -
https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils.
But I find it difficult to integrate with the LunarVim project -
https://github.com/LunarVim/LunarVim

Maybe you know something about it? Thanks!

jose-elias-alvarez:
I don't know much about LunarVim, but based on [this page](https://www.lunarvim.org/plugins/01-installing.html#example), it looks like you can add plugins using Packer. You'd then have to extend the default `on_attach` callback for `tsserver. [This PR](https://github.com/LunarVim/LunarVim/pull/1859) may have some info.

PR #316, opened 2021-11-04
statix: output change from stdout to stderr

meck:
Statix changed default diagnostics output in https://github.com/nerdypepper/statix/commit/1a97cce01f8e49b33bf28cbcdfeb3c8aefd809a5


jose-elias-alvarez:
Thank you for the quick fix!

Issue #166, opened 2021-09-08
Support running linters on `textDocument/didSave`


jose-elias-alvarez:
#69 describes the issues with `golangci-lint`, a linter that doesn't support `stdin` and also needs to run on a full project for context (meaning temp files won't work unless we create them within the project, which could cause other issues). Due to the timing of `textDocument/didChange` notifications, linting files on disk doesn't provide a good experience (the diagnostics "get stuck" and don't update as expected), which is why the PR is on hold. 

One solution would be to allow linters to only run on `textDocument/didSave` notifications. At the moment, it isn't possible to run both types of linters at the same time, since diagnostics from `didChange` linters continually clobber the ones from `didSave` linters, but neovim/neovim#15585 introduces the idea of diagnostic namespaces, which will solve this issue perfectly. 

The question is then whether supporting this is a good idea. The experience between the two types of linters is inconsistent, and I wonder whether it would cause confusion (or cause users to think that null-ls itself is broken). However, it seems like this isn't a problem for the VS Code ecosystem – for example, as far as I can tell the [golangci-lint integration for VS Code](https://golangci-lint.run/usage/integrations/) only runs on save. 

Since there seems to be demand to run these tools, I think it's a good idea to support them once the Neovim PR is merged, as long as we improve the documentation to avoid unnecessary bug reports / complaints. 

Edit: This could also be relevant to #155, though I think using temp files is always a better option if it's possible to do so.

PR #318, opened 2021-11-05
Add golangci-lint for golang linting

zalegrala:


zalegrala:
I can't tell what the linter is complaining about.  Any clues?

jose-elias-alvarez:
Thank you for the PR! I think the linter has an issue with the number of spaces? I checked out the PR locally and `stylua` reported the same issue, so I think you just have to run the file through `stylua`. 

Also: We had a [previous PR](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/69) that included this linter, and we've yet to merge it because of the issues discussed there. Could you verify whether your source updates diagnostics on change *before* saving the file? If not, I think it might be better to wait until I (or someone else) gets around to working on #166.

zalegrala:
The diagnostics show up before save.  If I open a new file I see the diagnostics as soon as golangci-lint can run.  I've pushed an update after running with stylua.

zalegrala:
The linter is now failing for an unrelated change.  Do you want me to address here as well?

jose-elias-alvarez:
Apologies – I should have been more specific. The issue isn't whether diagnostics are shown when first opening a file but rather whether diagnostics are updated when making a change to a file (before writing it to the disk). I briefly tested out and found that this PR suffers from the same issue. 

I'm not opposed to adding support for this class of linters in the future but am not sure about merging this right now, mostly since the inconsistency in observed behavior may lead users to conclude that something is broken. As I mentioned in the linked issue, it seems that the `golangci-lint` extension for VS Code only runs on save, so I think that's an acceptable compromise once we've gotten that implemented.

zalegrala:
Oh I see what you're saying.  If the buffer gets updated, I don't think the changes are reflected, since the execution is performed based on what is on disk.  I suppose I didn't notice since that's how the tool operates.  I do appreciate when the linters can be updated before saving, but as mentioned in the linked issues, this seems to be a tool limitation for this linter.  I was hopeful that this would be the fast path to being able to use golangci-lint for my config, but if I need to come up with another solution in the short term that's not a problem.  Let me know if I can do anything else to help this along.

jose-elias-alvarez:
Sure, thanks. In the meantime, if you're happy with the current functionality you can register it as a custom source:

```lua

local golangci_lint = {
    filetypes = { "go" },
    generator = null_ls.generator({
        command = "golangci-lint",
        to_stdin = true,
        from_stderr = false,
        args = {
            "run",
            "--out-format",
            "json",
            "$DIRNAME",
            "--path-prefix",
            "$ROOT",
        },
        format = "json",
            return code <= 2
        end,
        on_output = function(params)
            local diags = {}
            for _, d in ipairs(params.output.Issues) do
                if d.Pos.Filename == params.bufname then
                    table.insert(diags, {
                        row = d.Pos.Line,
                        col = d.Pos.Column,
                        message = d.Text,
                    })
                end
            end
            return diags
        end,
    }),
}
-- add to other sources or register individually
null_ls.register(golangci_lint)
```

kylo252:
I was checking the docs for this at https://golangci-lint.run/usage/integrations:
> Using it in an editor without `--fast` can freeze your editor. 

So maybe you should add it. You can also see all the flags that are used by [flycheck](https://github.com/weijiangan/flycheck-golangci-lint/blob/8e446c68311048f0b87febf8ef0379e29d358851/flycheck-golangci-lint.el#L78-L92) as a reference.

zalegrala:
I've added `--fast` though that appears to be guidance for vscode.  Should we close this out?  I see that its already out of date with what is in main.

jose-elias-alvarez:
It's up to you – to resolve conflicts you'd want to move the source into its own file, which is the structure we're using since d78734baeeedae8e73c76680726dfae57ed82d09. Once we support running on `textDocument/didSave`, you'd just have to change the method and this would be ready to merge.

zalegrala:
Ah okay thank you, just a new structure.  I've updated to accommodate.

lua/null-ls/builtins/diagnostics/golangci_lint.lua
```diff
@@ -0,0 +1,37 @@
+M.golangci_lint = h.make_builtin({
+    method = DIAGNOSTICS,
+    filetypes = { "go" },
+    generator_opts = {
+        command = "golangci-lint",
+        to_stdin = true,
+        from_stderr = false,
```
marcomayer, line 7: Adding `ignore_stderr = true,` might be a good idea to not have linter deprecation warnings in the console (to test use --enable-all as an arg to golangci-lint.

lua/null-ls/builtins/diagnostics/golangci_lint.lua
```diff
@@ -0,0 +1,37 @@
...
+        end,
+        on_output = function(params)
+            local diags = {}
+            for _, d in ipairs(params.output.Issues) do
+                print(d.Pos.Filename)
+                if d.Pos.Filename == params.bufname then
+                    table.insert(diags, {
+                        row = d.Pos.Line,
+                        col = d.Pos.Column,
+                        message = d.Text,
```
marcomayer, line 29: By default, all issues are "errors" which I think is a bit too hard. Not the perfect solution but maybe better, for now, is to default to warnings:
`severity = h.diagnostics.severities["warning"],`
For this to work you need `local h = require("null-ls.helpers")`

marcomayer (commented):
Would love to finally see golangci-lint in null-ls, here are some suggestions.

lua/null-ls/builtins/diagnostics/golangci_lint.lua
zalegrala: Ah, I'd added `suppress_errors = true` in my local version which seemed to suppress the deprecation.  Is `ignore_stderr` preferred?  The warnings all seem to go to stderr, while the json goes to stdout.

zalegrala: I see too that there is some json to collect warnings, but not sure how to represent that in the linter.
```
...
"Report": {
    "Warnings": [
      {
        "Tag": "runner",
        "Text": "The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive."
      },
      {
        "Tag": "runner",
        "Text": "The linter 'interfacer' is deprecated (since v1.38.0) due to: The repository of the linter has been archived by the owner. "
      }
    ],
```

lua/null-ls/builtins/diagnostics/golangci_lint.lua
zalegrala: I'd seen some usage of `severity` in other linters, but wasn't clear on its use.  What is that doing?

lua/null-ls/builtins/diagnostics/golangci_lint.lua
marcomayer: `suppress_errors` got deprecated and completely replace with `ignore_stderr`. Not sure about the deprecated warnings...from what I know it's not so clear how to properly deal with those even from a golangci-lint perspective. I guess in some new version they'll be removed so I just ignore the warnings for now...

zalegrala: Updating the configuration fixes the issue, I was mostly just pointing out that the information that is sent to stderr also seems to be present in the json.  So if there were a preferred way to access that from vim, then maybe we could do it.  I too mostly ignore them too, but perhaps I shouldn't, or perhaps users might want to know somehow.  I was slightly wondering if there was a way to get those into a quickfix list or something.

jose-elias-alvarez: We don't currently do anything with warnings like this unless they prevent the linter from running in the first place, so I think it's safe to ignore them. @marcomayer Are you able to test out #339?

lua/null-ls/builtins/diagnostics/golangci_lint.lua
marcomayer: When running the null-ls linter you'll have the results shown as diagnostics in nvim. There are different error levels aka serenity-levels there. By default they're seen as errors which is not true for most of the linters. An error would be something that really breaks the code like usage of a non-existing variable or a syntax error. But something like "this line is too long" or "you should add a comment here" doesn't break the code so you don't want it popping up as an error but as a warning. So adding severity explicitly changes the default from error to warning in this case.

I think you can set severity per golangci-linter config and it would then actually show up in the json results (by default it's empty). So ideally you'd check if there's something in there and only set the warning default in case it's not.

lua/null-ls/builtins/diagnostics/golangci_lint.lua
marcomayer: Great so you really want to add this to #339, also the default severity. I'm happy to test-drive this next week at work. For now, I can already say that the diagnostics do show up when saving a file but disappear as soon as I edit anything. That's at least unexpected and at least for me would be not something I'd like to work with :/

zalegrala:
Fixed in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/339

PR #320, opened 2021-11-08
fix(diagnostics): don't call handler if buffer changed since request

### LINKED ISSUE #307: Cancelling of linter requests when new changes are made
### Issues

### Feature description

When a linter is slow, it may take a while for its diagnostics to appear. If you make new changes in the mean time, chances are those diagnostics are no longer relevant. In addition, these changes produce new diagnostics. The result is that once your first set of diagnostics are displayed, new diagnostics may be displayed almost immediately after.

As an example, consider this series of events:

1. `debounce = 500`
2. You make some changes, then wait half a second
3. The linter needs at least one second to produce its diagnostics
4. You make some new changes in the mean time
5. This time the linter is much faster

The result is that the diagnostics from step 3 are displayed at some point, then are cleared and the diagnostics from step 4 are displayed. As this process involves changing signs, underlines, etc this can be quite jarring.

It would be nice if null-ls could somehow either cancel outstanding requests, or just ignore the diagnostics produced if a new request for diagnostics is made.

This NeoVim PR might be useful for this: https://github.com/neovim/neovim/pull/15949

### Help

No

### Implementation help

_No response_

jose-elias-alvarez:
Closes #307. As discussed there, the ideal solution is to cancel pending requests completely, but until we can achieve that, this may provide a better experience for slow linters.


jose-elias-alvarez:
My first attempt at this didn't make a lot of sense, but I think it's okay now. I tried artificially adding a delay to callbacks to simulate slow linters and it definitely seems less jarring. Could you try this out when you can and let me know if it solves your issue? @YorickPeterse

YorickPeterse:
@jose-elias-alvarez This seems to work based on testing with RuboCop. Thanks!

PR #322, opened 2021-11-09
Revert d78734baeeedae8e73c76680726dfae57ed82d09 - modularize buildins

Grzanka99:
This PR reverts modularize of builins and fix broken null-ls (fixing issue from screenshot). I think atleast for a while (until resolve issue) tha's good solution.
image


jose-elias-alvarez:
I think @bpinto's comment on the commit is right and this issue is caused by attempting to register a source that doesn't exist. On my end, replacing the nonexistent `whitespace` source with `trim_whitespace` fixes the issue. Could you verify whether that fixes it for you as well?

jose-elias-alvarez:
I think the cause of the issue is [this line](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/e7ca5a8a7d34d533c191399226705c5d9fb97d4b/lua/null-ls/builtins/init.lua#L17), since when the module doesn't exist, `builtin` is actually a string containing an error message. [This loop](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/e7ca5a8a7d34d533c191399226705c5d9fb97d4b/lua/null-ls/config.lua#L146) iterates over the table, which now includes the error string (previously it would have been `nil`, so the loop would have skipped it). 

Once you verify that correcting the name of the source fixes this issue, I'll change the return value so that the error doesn't happen even in this case (and add a warning message so users know they're trying to register a nonexistent source).

jose-elias-alvarez:
2faae9c70997485caef0feaf15e8729e821977e9 confirms that `null_ls.builtins.formatting.whitespace` was causing the error you saw. Now, you'll get a warning when attempting to access a nonexistent builtin, but this same error shouldn't occur (and the rest of your config should work as normal).

Grzanka99:
Thx, just weird error with this naming :P first when I started to using null-ls in documentation was trim_whitespace but only "whitespace" was working :P
Anyway, after renaming, everything works fine

PR #325, opened 2021-11-09
feat(builtins): add mypy built-in

### LINKED ISSUE #139: Add support for mypy

nawordar:
Closes #139.

Here is my attempt to add the mypy built-in. It differs from #155 in these ways:

- instead of using a temporary file, it uses a `--command` parameter with a file content,
- it includes exit code 2 as a positive result,
- it also parses diagnostic codes.

As for the exit code, I will ask upstream if they could add a separate exit code for syntax errors.

**EDIT:** I've just noticed that the spawn `mypy` processes don't close, so they fill up the whole CPU

### DIFF
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -1,0 +1,47 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    method = DIAGNOSTICS,
+    filetypes = { "python" },
+    generator_opts = {
+        command = "mypy",
+        args = {
+            "--hide-error-codes",
+            "--hide-error-context",
+            "--no-color-output",
+            "--show-column-numbers",
+            "--show-error-codes",
+            "--no-error-summary",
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks for this! I'm testing this out with [this repo](https://github.com/markkohdev/mypy-example), and on my end the `mypy` processes seem to close correctly. I wonder what's causing the difference in behavior?

jose-elias-alvarez:
I suggest we merge it and keep an eye on it. Does that sound good to you?

nawordar:
Yep

WhyNotHugo:
Thanks for this, works like a charm!

PR #326, opened 2021-11-09
Handle compile warnings preceeding credo output

aoswalt:
When working on a project that logs out a warning if loading a
dev.secret.exs file, I noticed that vscode users did not have any
issues. Taking a look through the plugin, it looks for the first
occurance of a `{` and parses from there.

https://github.com/pantajoe/vscode-elixir-credo/blob/main/src/execution.ts#L44


jose-elias-alvarez:
Great – makes sense and I appreciate the test coverage. Thank you!

PR #330, opened 2021-11-10
Completion source for vim-vsnip

oncomouse:
This adds a completion source [vim-vsnip](https://github.com/hrsh7th/vim-vsnip).

### DIFF
--- a/lua/null-ls/builtins/completion/vsnip.lua
+++ b/lua/null-ls/builtins/completion/vsnip.lua
@@ -1,0 +1,29 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local COMPLETION = methods.internal.COMPLETION
+
+return h.make_builtin({
+	method = COMPLETION,
+	filetypes = {},
+	name = "vsnip",
+	generator = {
+		fn = function(params, done)
+			local items = {}
+			local snips = vim.fn["vsnip#get_complete_items"](params.bufnr)
+			local targets = vim.tbl_filter(function(item)
+				return string.match(item.word, "^" .. params.word_to_complete)
+			end, snips)
+			for _, item in ipairs(targets) do
+				table.insert(items, {
+					label = item.abbr,
+					detail = item.menu,
+					kind = vim.lsp.protocol.CompletionItemKind["Snippet"],
+				})
+			end
+			done({ { items = items, isIncomplete = false } })
+		end,
+		async = true,
+	},
+})
+
--- a/lua/null-ls/builtins/completion/vsnip.lua
+++ b/lua/null-ls/builtins/completion/vsnip.lua
@@ -4,26 +4,25 @@ local methods = require("null-ls.methods")
 local COMPLETION = methods.internal.COMPLETION
 
 return h.make_builtin({
-	method = COMPLETION,
-	filetypes = {},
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/completion/vsnip.lua
```diff
@@ -0,0 +1,29 @@
...
+				return string.match(item.word, "^" .. params.word_to_complete)
+			end, snips)
+			for _, item in ipairs(targets) do
+				table.insert(items, {
+					label = item.abbr,
+					detail = item.menu,
+					kind = vim.lsp.protocol.CompletionItemKind["Snippet"],
+				})
+			end
+			done({ { items = items, isIncomplete = false } })
```
Furkanzmc, line 24: `isIncomplete` is used for caching the results of completion requests. It might be better if you can return `isIncomplete = #items == 0` rather than always returning false. This might not matter for a small list of snippets, and I doubt one's snippet list will grow to millions, but just mentioning to set precedence for the new completion sources to come.

I'm not familiar with vsnip, so my assumption can be incorrect, for tags and spell sources we are returning `isIncomplete = true` only if the number of completion items is `0`. For example in tags source, I get roughly the same list if I complete for `boo` or `boost`, so this mechanism works.

I notice that vsnip source for nvim-cmp also does the same thing you do and returns false all the time, so it might just be the right thing to do. 

Otherwise the PR looks good to me. Do you mind adding it to the [BUILTINS.md](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md) documentation as well?

What do you think?

oncomouse: That makes sense. I wasn't sure what that check was for, so it wasn't a vsnip-based decision. I've made that change.

jose-elias-alvarez:
Looks good and works well on my end. I wonder if it's worth mentioning in the documentation that actually expanding snippets requires setting up a keymap, or do you think that would be clear enough to users?

Furkanzmc:
> Looks good and works well on my end. I wonder if it's worth mentioning in the documentation that actually expanding snippets requires setting up a keymap, or do you think that would be clear enough to users?

That's a good point, I had not thought of it. I think in the BUILTINS.md documentation, that should be mentioned. It's also worth noting that null-ls would not give any snippet related functionality (like jumping to marks etc).

oncomouse:
That makes sense. I'll note it. One other documentation question: there's a plugin called [vim-vsnip-integ](https://github.com/hrsh7th/vim-vsnip-integ) that adds support for LSPs that return snippets (such as the vscode servers). As far as I can tell, this plugin is NOT required because null-ls isn't actually returning the snippets.

My question is: is it clearer to ignore this question and say nothing or to mention that the integration plugin is not needed?

jose-elias-alvarez:
Well, if a user installs the integration plugin and it doesn't cause any issues, it's not a big issue, right? The goal is for users to understand what they can and can't expect from the completion source, so I'd trim it down to something like this:

> Registering this source will show available snippets in the completion list, but vim-vsnip is responsible for expanding them. See [vim-vsnip's documentation](https://github.com/hrsh7th/vim-vsnip#2-setting) for setup instructions.

jose-elias-alvarez:
Thanks! I think this is all set now.

PR #332, opened 2021-11-11
feat: improve source API

jose-elias-alvarez:
The goal of this PR is to simplify how sources are registered to clean up a messy portion of the code base and enable new features, namely:

- Disabling filetypes per source (i.e. enable all filetypes besides a subset)
- More transparent querying of registered sources (no need to go through the info API, which isn't really ideal for consumption)
- Disabling sources
- Registering an arbitrary number of sources with the same name (which lets through duplicates, but that hasn't been an issue AFAIK)
- ~~Running sources on demand (not really related, but something I want to include)~~ out of scope and not really consistent with the goal of LSP-like behavior

Currently, generators are stored independently and anonymously, making it impossible to disable them or otherwise manipulate them once registered. Coupling sources and their generators is both logical and solves the issue, so we now store the sources themselves in the config object. We also avoid registering filetypes and methods independently and simply check the sources themselves for availability (with a simple transformation applied on registration to make this check easier).

I also have to think about how to handle disabling sources. ~~Users will most likely want to disable them by name~~ Edit: I think the best solution is a simple query mechanism that either specifies a string (assumed to be a name) or an object specifying name, method, and / or ID that then disables all matching sources. 

Finally, I want to add test coverage to the `info`, `rpc`, and `lspconfig` modules . Working on this project has taken up a good chunk of my life lately, so I want to finish this, clean up a few outstanding feature requests, and take a break. 

TODO:

- [x] Add missing coverage
- [x] Add `disable` method
- [x] Document source API

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -80,6 +80,19 @@ If you see `filetypes = {}` in a source's description, that means the source is
 active for all filetypes by default. You may want to define a specific list of
 filetypes to avoid conflicts or other issues.
 
+You can also pass a list of specifically disabled filetypes:
+
+```lua
+local sources = {
+    null_ls.builtins.code_actions.gitsigns.with({
+        disabled_filetypes = { "lua" },
+    }),
+}
+```
+
+null-ls is always inactive in non-file buffers (e.g. file trees, finders) so
+theres's no need to specify them here.
+
 ### Arguments
 
 To add more arguments to a source's defaults, use `extra_args`:
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -124,3 +124,6 @@ end
 If null-ls is already running but you want to stop it, you can use the methods
 provided by nvim-lspconfig (`:LspStart`, `:LspStop`, and `:LspRestart`) to
 control its behavior.
+
+You can also deregister sources using the source API, as described in
+[SOURCES](SOURCES.md).
--- a/doc/MAIN.md
+++ b/doc/MAIN.md
@@ -95,6 +95,9 @@ my_source.filetypes = { "lua", "teal" }
 my_source.filetypes = {}
 ```
 
+Sources can also include a list of `disabled_filetypes`. null-ls checks these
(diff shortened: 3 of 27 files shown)


jose-elias-alvarez:
This should be good to go. I'm not seeing any issues on my end, so I'm going to merge it and deal with anything I missed as we go.

David-Else:
Thanks for all the amazing hard work you have done on this!

I have read the new explanations. My current config is:

```lua
local null_ls = require 'null-ls'

local sources = {
  -- Use denols instead for 'javascript', 'javascriptreact', 'typescript', 'typescriptreact'
  null_ls.builtins.formatting.prettier.with {
    filetypes = {
      'vue',
      'svelte',
      'css',
      'scss',
      'html',
      'json',
      'yaml',
      'markdown',
    },
  },

  null_ls.builtins.formatting.stylua.with {
    extra_args = { '--config-path', vim.fn.expand '~/.stylua.toml' },
  },
  null_ls.builtins.formatting.shfmt.with {
    extra_args = { '-i', '4' },
  },
  null_ls.builtins.diagnostics.shellcheck,
  null_ls.builtins.diagnostics.vale,
}
null_ls.config { sources = sources }
```
I think the new `disabled_filetypes` might be a big improvement. At the moment I am copy/pasting the `filetypes` for Prettier from the docs and removing 'javascript', 'javascriptreact', 'typescript', 'typescriptreact' to 'disable' them. I assume the following now works, or maybe I need to add something else?

```lua
  null_ls.builtins.formatting.prettier.with {
    disabled_filetypes = { 'javascript', 'javascriptreact', 'typescript', 'typescriptreact' },
  },
```

Have I understood correctly, and could you look at my config and tell me if it is written in an optimized up-to-date way, could I refactor it to be cleaner since this merge?

jose-elias-alvarez:
Yep, `disabled_filetypes` should work as you expect (I'm doing the same thing on my end). The deregister API is meant to completely wipe out sources, but toggling them is also a likely (or even likelier) use case, and it should be simple to add support for it. I'll open an issue.

PR #335, opened 2021-11-11
fix(logger): configurable logging behavior

### LINKED ISSUE #280: Show error messages when command failed?

kylo252:
- Set the correct stack level in plenary.log
- Use all the levels of logging with the ability of using the console
- Switch `echo_error` to use `Log:warn()`
- Switch most of `u.debug_log()` to `Log:trace()` while keeping some of them as `Log:debug()`

Fixes #280

- fix(logger): configurable logging behavior
- fix(logger): restore the quick debug toggle
- fix(logger): rebase
- chore: use lower-case for logger instances
- fix(logger): explicit mention of empty completion
- chore: naming convention

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,22 @@ memory without any external processes, in most cases it should run faster than
 similar solutions. If you notice that performance is worse with null-ls than
 with an alternative, please open an issue!
 
+### Debugging
+
+1. Enable debug logging
+
+   ```lua
+   require("null-ls").config({
+      debug = true,
+   })
+
+   ```
+
+2. Use `:NullLsLog` to open the log file.
+
+As with LSP logging, `debug` will slow down Neovim. Make sure to disable the
+option after you've collected the information you're looking for.
+
 ## Tests
 
 The test suite includes unit and integration tests and depends on plenary.nvim.
--- a/lua/null-ls/completion.lua
+++ b/lua/null-ls/completion.lua
@@ -1,6 +1,7 @@
 local u = require("null-ls.utils")
 local methods = require("null-ls.methods")
 local config = require("null-ls.config").get()
+local log = require("null-ls.logger")
 
 local M = {}
 
@@ -12,11 +13,12 @@ M.handler = function(method, original_params, handler)
             method = methods.map[method],
             params = params,
             callback = function(results)
-                u.debug_log("received completion results from generators")
-                u.debug_log(results)
                 if #results == 0 then
+                    log:debug("received no completion results from generators")
                     handler({})
                 else
+                    log:debug("received completion results from generators")
+                    log:trace(vim.inspect(results))
                     local items = {}
                     local isIncomplete = false
                     for _, result in ipairs(results) do
--- a/lua/null-ls/config.lua
+++ b/lua/null-ls/config.lua
@@ -4,7 +4,18 @@ local defaults = {
     diagnostics_format = "#{m}",
     debounce = 250,
     default_timeout = 5000,
+    ---@usage setting this to true will enable debug logging
     debug = false,
+    log = {
+        ---@usage disable logging completely
+        enable = true,
+        ---@usage set logging level
+        --- possible valuse: {"error", "warn", "info", "debug", "trace"}
+        level = "warn",
+        ---@usage whether to print the output to neovim's
+        --- possible values: 'sync','async', false
+        use_console = "async",
+    },
     -- prevent double setup
     _setup = false,
     -- force using LSP handler, even when native API is available (e.g diagnostics)
@@ -77,6 +88,9 @@ M.setup = function(user_config)
 
     local validated = validate_config(user_config)
     config = vim.tbl_extend("force", config, validated)
+
+    -- FIXME: validating log options should maintain any default options
+    config.log = vim.tbl_extend("keep", config.log, defaults.log)
 
     if config.sources then
         require("null-ls.sources").register(config.sources)
--- a/lua/null-ls/formatting.lua
+++ b/lua/null-ls/formatting.lua
@@ -1,5 +1,6 @@
 local u = require("null-ls.utils")
 local methods = require("null-ls.methods")
+local log = require("null-ls.logger")
 
 local api = vim.api
 
@@ -51,8 +52,8 @@ M.apply_edits = function(edits, params)
     -- formatting and rangeFormatting handlers are identical
     local handler = u.resolve_handler(params.lsp_method)
 
-    u.debug_log("received edits from generators:")
-    u.debug_log(edits)
+    log:debug("received edits from generators")
+    log:trace(edits)
 
     local diffed_edits = {}
     for _, edit in ipairs(edits) do
@@ -72,8 +73,8 @@ M.apply_edits = function(edits, params)
         restore_win_data(marks, views, bufnr)
     end)
 
-    u.debug_log("successfully applied edits:")
-    u.debug_log(diffed_edits)
+    log:debug("successfully applied edits")
+    log:trace(diffed_edits)
 end
 
 M.handler = function(method, original_params, handler)
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -160,16 +160,15 @@ M.generator_factory = function(opts)
             opts.command = command
         end
 
-        local err_msg = ""
+        local err_msg
         if vim.fn.executable(command) ~= 1 then
             err_msg = string.format(
                 "command %s is not executable (make sure it's installed and on your $PATH)",
                 command
             )
-            log:debug(err_msg)
         end
 
-        assert(err_msg, "")
+        assert(not err_msg, err_msg)
         assert(not from_temp_file or to_temp_file, "from_temp_file requires to_temp_file to be true")
 
         _validated = true
--- a/lua/null-ls/helpers.lua
+++ b/lua/null-ls/helpers.lua
@@ -1,6 +1,7 @@
 local u = require("null-ls.utils")
 local c = require("null-ls.config")
 local s = require("null-ls.state")
+local log = require("null-ls.logger")
 
 local api = vim.api
 local validate = vim.validate
@@ -159,10 +160,16 @@ M.generator_factory = function(opts)
             opts.command = command
         end
 
-        assert(
-            vim.fn.executable(command) > 0,
-            string.format("command %s is not executable (make sure it's installed and on your $PATH)", command)
-        )
+        local err_msg = ""
+        if vim.fn.executable(command) ~= 1 then
+            err_msg = string.format(
+                "command %s is not executable (make sure it's installed and on your $PATH)",
+                command
+            )
+            log:debug(err_msg)
+        end
+
+        assert(err_msg, "")
         assert(not from_temp_file or to_temp_file, "from_temp_file requires to_temp_file to be true")
 
         _validated = true
@@ -177,8 +184,8 @@ M.generator_factory = function(opts)
             end
 
             local wrapper = function(error_output, output)
-                u.debug_log("error output: " .. (error_output or "nil"))
-                u.debug_log("output: " .. (output or "nil"))
+                log:trace("error output: " .. (error_output or "nil"))
+                log:trace("output: " .. (output or "nil"))
 
                 if ignore_stderr then
                     error_output = nil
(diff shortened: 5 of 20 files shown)


kylo252:
> 1. Moving the detailed description of log configuration out of the basic FAQ to avoid confusion

Yeah, that's always fine. I just didn't know what where it should go.

> 2. No logging at all by default
> 3. Logging defaults to a file, not the console
> 4. Always output critical warnings and errors to the console, regardless of the user's settings

I feel like I'm missing something here. When `log:error` gets triggered it will output to **both** the file, and the console by default.
Here's a screenshot comparison.

image

The one at the top is using the logger, you can tell by the timestamp, while the one on the bottom is using just the `main` branch. The difference is that now I can actually show you this one as well, because it got recorded. Please note that only warnings and errors are shown in either the console or the file, because the logger will simply ignore the request if it doesn't fall within the configured level (note: I had a bug in the code that was ignoring that log-level 😅)

```logtalk
[ERROR Thu 11 Nov 2021 06:30:45 PM CET] ...e/pack/packer/start/null-ls.nvim/lua/null-ls/helpers.lua:164: command markdownlint is not executable (make sure it's installed and on your $PATH)
[WARN  Thu 11 Nov 2021 06:30:50 PM CET] ...site/pack/packer/start/null-ls.nvim/lua/null-ls/loop.lua:99: command timed out after 5000 ms
```

Now let's consider that something a bit more tricky is happening, and you want to see why `prettier` is hanging, then I can turn the logging up a notch to `debug` and either monitor the console or the logfile directly. Or take it to the max and use `trace` level.

This is just a QoL change for everyone involved, for you as the plugin author, and for me as both as a user of your plugin and in the context of maintaining LunarVim.

Of course, it's all up to you at the end, I just figured I need to give you a more concrete example. I can also do another recording if you think it's useful.

README.md
```diff
@@ -281,6 +281,33 @@ memory without any external processes, in most cases it should run faster than
 similar solutions. If you notice that performance is worse with null-ls than
 with an alternative, please open an issue!
 
+### Debugging
+
+- Enable debug logging
```
jose-elias-alvarez, line 286: I strongly prefer for this to only mention the `debug` option. Given that setting `debug = true` will output everything to the console unless otherwise specified, I think we can trim down these instructions to just mention `:messages` and the new `:NullLsLog` command.

lua/null-ls/completion.lua
```diff
@@ -12,11 +13,11 @@ M.handler = function(method, original_params, handler)
             method = methods.map[method],
             params = params,
             callback = function(results)
-                u.debug_log("received completion results from generators")
-                u.debug_log(results)
                 if #results == 0 then
+                    log:debug("no results were received completion results from generators")
```
jose-elias-alvarez, line 17:
```suggestion
                    log:debug("received no completion results from generators")
```

lua/null-ls/formatting.lua
```diff
@@ -51,8 +52,8 @@ M.apply_edits = function(edits, params)
     -- formatting and rangeFormatting handlers are identical
     local handler = u.resolve_handler(params.lsp_method)
 
-    u.debug_log("received edits from generators:")
-    u.debug_log(edits)
+    log:debug("received edits from generators:")
```
jose-elias-alvarez, line 55:
```suggestion
    log:debug("received edits from generators")
```

lua/null-ls/formatting.lua
```diff
@@ -77,8 +78,8 @@ M.apply_edits = function(edits, params)
         restore_win_data(marks, views, bufnr)
     end)
 
-    u.debug_log("successfully applied edits:")
-    u.debug_log(diffed_edits)
+    log:debug("successfully applied edits:")
```
jose-elias-alvarez, line 81:
```suggestion
    log:debug("successfully applied edits")
```

lua/null-ls/loop.lua
```diff
@@ -95,7 +96,7 @@ M.spawn = function(cmd, args, opts)
 
     if timeout then
         timer = M.timer(timeout, nil, true, function()
-            u.debug_log("command timed out after " .. timeout .. " ms")
+            log:warn("command timed out after " .. timeout .. " ms")
```
jose-elias-alvarez, line 99: Could we also add the name of the command here?

lua/null-ls/config.lua
```diff
@@ -6,7 +6,18 @@ local defaults = {
...
     debug = false,
+    log = {
+        ---@usage disable logging completely
+        enable = true,
+        ---@usage set logging level
+        --- possible valuse: {"error", "warn", "info", "debug", "trace"}
+        level = "warn",
+        ---@usage whether to print the output to neovim's
+        --- possible values: 'sync','async', false
+        use_console = "sync",
```
jose-elias-alvarez, line 19: Is there any disadvantage to setting this to `sync` by default?

lua/null-ls/helpers.lua
```diff
@@ -159,10 +160,12 @@ M.generator_factory = function(opts)
             opts.command = command
         end
 
-        assert(
-            vim.fn.executable(command) > 0,
-            string.format("command %s is not executable (make sure it's installed and on your $PATH)", command)
-        )
+        if vim.fn.executable(command) ~= 1 then
+            log:error(
```
jose-elias-alvarez, line 164: Yes, this needs to prevent the generator from running. Since all of the `validate` statements will already throw errors, I think it's fine to change this back to `assert`. We can wrap `validate_opts` in `pcall` and catch the error so we can log it, but that's a bigger change so we can leave it out of this PR.

jose-elias-alvarez (changes requested):
Thanks for clarifying the rationale behind these changes. Apart from the comments and a few merge conflicts, I ran into an issue while testing this locally with `debug = true`. The output from formatters fills up the whole screen every time I format the file, which is pretty annoying when combined with formatting on save. 

I think changing `debug = true` so that it both sets the appropriate log level and disables `use_console` is a sensible solution, especially since I think users would appreciate the option to copy-paste the log output from `:NullLsLog`.

jose-elias-alvarez:
Also, I forgot to mention, but you might have seen how in the new completion source, we are using the `debug` flag to run some additional validation in handlers, and I want to expand this to the other handlers, too. I think keeping `debug` and using it as an all-purpose, tell-me-anything-that-might-be-going-wrong option is still a good idea.

README.md
kylo252: How about about something like this? Otherwise, feel free to write it here and and I can just copy it, or you can also push the commit yourself.

lua/null-ls/loop.lua
kylo252: Yeah, I've missed this one in the rebase, ` u.debug_log` doesn't exists anymore.

lua/null-ls/config.lua
kylo252: Did you mean `async`? so `async` is faster, but I noticed that it doesn't quite show the errors/warnings in order, if that's something that matters to you. As long as the log level is `warn`/`error` then it doesn't really matter to use `async` because you should seldom have two consecutive errors. I'll set it back to `async` since we moved the `debug` option to be file-only, more or less.

lua/null-ls/loop.lua
kylo252: Actually, I'm not sure that we want this one to be a warning because I think it's triggering way too often for my liking. So I'll switch it to be `debug` level, for performance reasons.

kylo252: what do you think of `on_error()`? I based it on  [/runtime/lua/vim/lsp/rpc.lua#L228](https://github.com//neovim/neovim/blob/29cec32f44db358b0dfcd560d2ba8228b9d9de39/runtime/lua/vim/lsp/rpc.lua#L228), but I'm not sure if you want something more specific.

lua/null-ls/helpers.lua
kylo252: I think I found a good middle-ground. Let me know if you don't like it.

lua/null-ls/loop.lua
jose-elias-alvarez: Sorry, could you clarify what you mean by this? Seeing commands time out should be pretty rare if the command is running properly.

README.md
jose-elias-alvarez: Seems fine, thanks!

lua/null-ls/loop.lua
```diff
@@ -9,6 +9,12 @@ local close_handle = function(handle)
     end
 end
 
+local on_error = function(errkind, ...)
+    log:debug(string.format("[%q] has failed with: %q", errkind, table.concat(..., ", ")))
+    -- we need to make sure to raise an error
+    error(string.format("[%q] has failed with: %q", errkind, table.concat(..., ", ")))
+end
```
kylo252, line 16: @jose-elias-alvarez, This is the new `on_error()` handler I was talking about.
See how it's called in `handle_stdout` and `handle_stderr`.

jose-elias-alvarez: I see now, makes sense.

lua/null-ls/helpers.lua
```diff
@@ -159,10 +160,17 @@ M.generator_factory = function(opts)
...
+        local err_msg = ""
+        if vim.fn.executable(command) ~= 1 then
+            err_msg = string.format(
+                "command %s is not executable (make sure it's installed and on your $PATH)",
+                command
+            )
+            log:debug(err_msg)
+        end
+
+        -- https://github.com/jose-elias-alvarez/null-ls.nvim/pull/335/files#r747896325
```
jose-elias-alvarez, line 172:
```suggestion
```

jose-elias-alvarez (approved):
Looks good to me! I recently merged a few commits that added more log statements, so could you rebase this against `main`, make sure we've converted everything, and let me know when this is ready to merge?

kylo252:
@jose-elias-alvarez, I think that should cover it!

We should defintely add more log entres, but that can be done incrementally afterwards.

jose-elias-alvarez:
I went ahead and fixed up a few things I noticed while trying this out, but I think it's a nice improvement. Thanks again for your patience and work on this!

jose-elias-alvarez:
@kylo252 Do you know how to use the `lazy` and `file` methods on the Plenary logger (found [here](https://github.com/nvim-lua/plenary.nvim/blob/1c31adb35fcebe921f65e5c6ff6d5481fa5fa5ac/lua/plenary/log.lua#L177))? I tried the examples in the comments but wasn't able to get them to work. They'd be useful to have.

williamboman:
I never understood how plenary's lazy log was implemented, it would never print anything. I refactored it a bit to function in a way that I had expected it to. Examples:

```lua
log.lazy_debug(function ()
    local expensive_stuff = do_something()
    return "These return values follow the fmt_debug signature, with placeholders %s", expensive_stuff
end)
```

jose-elias-alvarez:
@williamboman Thanks for confirming – the nvim-lsp-installer implementation is a lot cleaner and makes more sense than Plenary's, so I'll see how we can fix ours up a bit. 

I'll also see if I can make sense of what Plenary is trying to do and open a PR (or an issue if I can't figure it out).

PR #343, opened 2021-11-13
Statix fix options

meck:
Statix changed it's default config again, but a stdin
has been added so we use this

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1769,7 +1769,7 @@ local sources = { null_ls.builtins.diagnostics.statix }
 
 - `filetypes = { "nix" }`
 - `command = "statix"`
-- `args = { "check", "--format=errfmt", "--", "$FILENAME" }`
+- `args = { "check", "--stdin", "--format=errfmt" }`
 
 #### [Stylelint](https://github.com/stylelint/stylelint)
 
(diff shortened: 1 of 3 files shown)


Furkanzmc:
All these fixes assume that the user of the plugin is using the latest version of the executable. Is it possible to use the `condition` function to register the correct arguments when the source is registered? @jose-elias-alvarez

jose-elias-alvarez:
We don't have an explicit stance on this, but for built-ins, I think we should generally stick with whatever works with the latest version unless there's a widespread reason making it impossible / undesirable for users to update. 

I know absolutely nothing about Statix but it seems to be very new (first commit on September 10) so does it seem reasonable to expect that everyone's on the latest version or can easily update?

meck:
Yes Its fairly new, @nerdypepper might chime in on the stability of the interface?

I suspect there wont be many users of previous versions as the application is for the Nix package manager/language, so the majority of users will install statix via Nix, which currently has the latest version in its unstable channel (not in any stable channel yet AFAIK). 

As null-ls is also in Nixpks a reasonable approach is to try to keep it compatible with the Statix version in nixpkgs master.

jose-elias-alvarez:
Makes sense, thanks for confirming!

PR #345, opened 2021-11-14
add .null-ls-root to LSP root patterns

kabouzeid:
Allows to explicitly set the project root for certain projects

### DIFF
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -127,3 +127,7 @@ control its behavior.
 
 You can also deregister sources using the source API, as described in
 [SOURCES](SOURCES.md).
+
+## Explicitly defining the project root
+
+Create an empty file `.null-ls-root` in the directory you want to mark as the project root for null-ls.
--- a/lua/null-ls/lspconfig.lua
+++ b/lua/null-ls/lspconfig.lua
@@ -41,7 +41,7 @@ function M.setup()
         cmd = { "nvim" },
         name = "null-ls",
         root_dir = function(fname)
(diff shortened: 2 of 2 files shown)


jose-elias-alvarez:
This is a good idea, thanks! Could we add a note to the documentation? We don't have an ideal place for it but maybe [CONFIG](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md) could work.

MunifTanjim:
Instead of hardcoding such a specific filename `.null-ls-root` (which is only ever useful for this plugin), can use a global variable maybe (e.g. `vim.g.null_ls_root_file`)? That way we can set that variable in our config or project local settings (exrc)?

jose-elias-alvarez:
@MunifTanjim Something like this is already possible to do via lspconfig:

```lua
require("lspconfig")["null-ls"].setup({
    root_dir = function(fname)
        return "my-root-dir"
    end,
})
```

Arguably, that also makes this PR unnecessary, though I can still see it as a convenient workaround.

kabouzeid:
There are several language servers that leverage this technique, see eg `.ccls-root` (https://github.com/MaskRay/ccls/wiki/FAQ#project-root-detection).

The problem with
```lua
require("lspconfig")["null-ls"].setup({
    root_dir = function(fname)
        return "my-root-dir"
    end,
})
```
is that it overwrites the original `root_dir` function which I would have to duplicate here.

jose-elias-alvarez:
Yep, I think this could be a simple solution for certain use cases. Thank you for this!

PR #346, opened 2021-11-15
Use `--stdin-path` argument in phpcbf to make exclude-pattern works

gbprod:
Ref: https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/341


gbprod:
I think we can merge it.

jose-elias-alvarez:
Sure, thank you for all of your help!

PR #349, opened 2021-11-15
feat(builtins): add prefer_local and only_local options

### LINKED ISSUE #289: Better support for monorepos
### Issues

### Feature description

Hi,

Where I work we use a lot of monorepos, and those doesn't even contain the same language (e.g. in a project the `frontend` directory will be a typescript project while the `backend` will be python)

The way null-ls works has 2 issues for monorepos like those:

1) Some linters/formatters needs to run with their `cwd` set to the directory where the configuration is defined. For example, inside the `backend` folder we have a `.flake8` file which configures flake8, but if null-ls runs it from the root (the directory above, where .git is) it will not use those configs.

I could get around this by passing a custom `cwd` function, but I'm not sure it is the proper way (I'll explain why bellow)

2) The major issue: We need to run the specific installed linters/formatters inside those projects. For example, we have some `flake8` plugins installed in the project which will not be included in the global flake8 binary. Idially the `flake8` installed in the project should be used.

To workaround that I did the something like this:

```lua
null_ls.config({
    sources = {
        null_ls.builtins.diagnostics.flake8.with({
            command = function()
                if ROOT_DIRS.python ~= nil then
                    local match = vim.fn.glob(path.join(ROOT_DIRS.python, ".venv", "bin", "flake8"))
                    if match ~= "" then
                        return match
                    end
                end
                return "flake8"
            end,
            cwd = function(params)
                return ROOT_DIRS.python or params.cwd
            end,
        }),
    }
})
```

What is the problem there? That `ROOT_DIRS.python` is defined like this:

```lua
ROOT_DIRS = {
    python = nil,
    node = nil,
}
nvim_lsp.pyright.setup({
    before_init = function(_, config)
        ROOT_DIRS.python = config.root_dir
    end,
})
```

So, I'm depending that the python LSP runs first because it will find the properly root dir for that specific file in the buffer, so that when null_ls is going to retrieve the command, I already have the python root dir.

I cannot simply set the root dir to something else because each one will have its own root and its own formatters/linters (in my example, I want to run `flake8`, `isort` and `black` iside the `backend` dir and `prettier` and `eslint` inside the `frontend` dir). I did the same for each source, I'm not showing everything here for simplicity =P

In summary, this works and I got everything running. But I'm afraid of that "if null-ls asks for the cmd first?" and also it doesn't seem to be a clean and elegand solution.

Maybe there's a way to improve that kind of workflow?

### Help

No

### Implementation help

_No response_

jose-elias-alvarez:
Closes #289.

This PR adds two new options for built-ins, `prefer_local` and `only_local`. It also exposes the underlying `dynamic_command` mechanism for generators created using `generator_factory`. 

When a source with `prefer_local = true` runs for the first time in a given buffer, null-ls will traverse that buffer's parent directories and try to find a local copy of `command`, stopping once it hits the client's root directory. If not found, it will fall back to a globally-available executable. The results of the search are cached and used whenever the source runs again. `only_local` works the same way without falling back (i.e. it only runs if a local copy is found).

Both options accept a string, which is treated as a prefix (see the documentation for an example).

If found, the `cwd` of the processes will be set to the directory in which the local copy was found (unless specifically set by the user).

There's (at least) one potentially confusing piece of behavior in this PR, which is that when `only_local` is set and no local executable is found, `:NullLsInfo` will still show the source as available for the current buffer. There are workarounds for that, but none of them are pretty, so I'd rather leave it as-is for now.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -158,6 +158,28 @@ See [CONFIG](CONFIG.md) to learn about the structure of `diagnostics_format`.
 Note that specifying `diagnostics_format` for a built-in will override your
 global `diagnostics_format` for that source.
 
+## Local executables
+
+To prefer using a local executable for a built-in, use the `prefer_local`
+option. This will cause null-ls to search upwards from the current buffer's
+directory, try to find a local executable at each parent directory, and fall
+back to a global executable if it can't find one locally.
+
+`prefer_local` can be a boolean or a string, in which case it's treated as a
+prefix. For example, the following settings will cause null-ls to search for
+`node_modules/.bin/prettier`:
+
+```lua
+local sources = {
+    null_ls.builtins.formatting.prettier.with({
+        prefer_local = "node_modules/.bin",
+    }),
+}
+```
+
+To _only_ use a local executable without falling back, use `only_local`, which
+accepts the same options.
+
 ## Conditional registration
 
 null-ls supports dynamic registration, meaning that you can register sources
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -180,6 +180,10 @@ local sources = {
 To _only_ use a local executable without falling back, use `only_local`, which
 accepts the same options.
 
+By default, these options will also set the `cwd` of the spawned process to the
+parent directory of the local executable (if found). You can override this by
+manually setting `cwd` to a function that should return your preferred `cwd`.
+
 ## Conditional registration
 
 null-ls supports dynamic registration, meaning that you can register sources
--- a/doc/HELPERS.md
+++ b/doc/HELPERS.md
@@ -38,6 +38,7 @@ helpers.generator_factory({
     use_cache, -- boolean (optional)
     runtime_condition, -- function (optional)
     cwd, -- function (optional)
+    dynamic_command, -- function(optional)
 })
 ```
 
@@ -193,6 +194,18 @@ Optional callback to set the working directory for the spawned process. Takes a
 single argument, `params`, which is a table containing information about the
 current editor state (described in [MAIN](./MAIN.md)). If the callback returns
 `nil`, the working directory defaults to the project's root.
+
+### dynamic_command
+
+Optional callback to set `command` dynamically. Takes two arguments, a string
+containing the base `command` and a `params` object containing information about
+the current buffer's state. The callback should return a string containing the
+command to run or `nil`, meaning that no command should run.
+
+`dynamic_command` runs every time its parent generator runs and can affect
+performance, so it's best to cache its output when possible.
+
+Note that setting `dynamic_command` will disable `command` validation.
 
 ## formatter_factory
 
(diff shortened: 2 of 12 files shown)


kylo252 (commented):
I think it might be better to scope this logic into specific sources instead of adding general complexity to all the sources.

I have been meaning to do something in the same vein but with quite a different implementation. Specifically, I wanted to port the resolve_command logic that is used in https://github.com/LunarVim/lunarvim/blob/b3538b63ce7626f269455881c7d659926ad4ee85/lua/lvim/lsp/null-ls/services.lua#L34.

I had a few ideas to approach this, but the easiest would be to always use the result of the function I used as the command for applicable sources. You can see how `prettier` has a `from_node_modules` resolver, so `flake8` would have a `from_python_env` or something along those lines. And this kind of approach is a lot more predictable, and will already support a few cases without worrying about explaining how it works to the users.

jose-elias-alvarez:
> I think it might be better to scope this logic into specific sources instead of adding general complexity to all the sources.

Apart from `dynamic_command`, this is just an opt-in wrapper around existing functionality. I like your solution (and have used something similar in my config) but it won't handle the cases described in #289 where each buffer potentially resolves to a different command, since `command` won't be re-evaluated. It also doesn't handle `cwd` modification. 

I'd frankly prefer to reduce complexity by leaving things like this up to users in the first place, but I can see that it's useful and don't want to just dump a block of code into a wiki page. If there's demand for it, in the future we could reuse the same functionality to create wrappers for common cases like `node_modules` and maybe even enable this as a default for some sources.

kylo252:
> I'd frankly prefer to reduce complexity by leaving things like this up to users in the first place, but I can see that it's useful and don't want to just dump a block of code into a wiki page. If there's demand for it, in the future we could reuse the same functionality to create wrappers for common cases like `node_modules` and maybe even enable this as a default for some sources.

Take a look at https://github.com/LunarVim/LunarVim/issues/1833, which I doubt can be solved without introducing something similar to what I suggested.

jose-elias-alvarez:
@kylo252 As far as I can tell, the error is coming from your validation, right? It seems like any solution has to handle 3 situations:

1. The user registers a linter but doesn't have it installed globally, and they don't understand why it's not working, so we should try to give them a hint (how null-ls normally works). 
2. The user wants to use local executables if available but fall back to global (handled by `prefer_local` in this PR).
3. The user only wants to use local executables and doesn't want to see warnings if the command isn't available globally (handled by `only_local` in this PR). 

It seems like the solution you guys are using tries to handle cases 1 and 2 but doesn't account for 3, and I don't think there's a good way to distinguish between cases 2 and 3 unless the user specifies. If we merged this, you could either directly expose `prefer_local` / `only_local`, wrap them, or do something else entirely. This is just meant to provide one solution to an apparently common issue.

phantomwhale:
I've got projects with Prettier in `/node_modules/.bin`, `/assets/node_modules/bin` and some without anything where I want to use my global binary.

This was exactly the change I was looking for - everything works now without me tweaking my LUA config each time - thanks!!!

Issue #194, opened 2021-09-15
Provide diagnostics for closed files in workspace


cblessing24:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Provide diagnostics for all files in the current workspace irrespective of whether they are open in neovim or not.

Currently null-ls only provides diagnostics for files that are open in neovim as far as I know. I think providing diagnostics for closed files as well would be very useful for checking all diagnostics in the current workspace via telescope for example. Other language servers (e.g. pyright) have this functionality already.

### Help

No

jose-elias-alvarez:
Sure, it's possible to implement something like this if a source supports it, though it would take some extra work, since the current diagnostics framework is set up to handle one buffer at a time. Is there a specific source you have in mind (e.g. a CLI program that can provide diagnostics for an entire project at once)? I know ESLint supports this, so I could try to put together a prototype when I have time.

cblessing24:
I mainly use flake8, mypy and pylint to lint my Python code. Maybe flake8 would be a good place to start because it is already supported?

jose-elias-alvarez:
Got it. For reference, how does pyright handle workspace diagnostics? Is it a separate command?

cblessing24:
Honestly, I am not sure. I guess it just publishes diagnostics for all python files in the workspace instead of just the open ones over the protocol.

jryom:
Just to chime in, `tsc` might be another good use for this. It's what [`coc-tsserver` uses to provide diagnostics for TypeScript project](https://github.com/neoclide/coc-tsserver/blob/68bb4634f345b70c8b2e8c48294dd3619eb5b810/package.json#L97-L101), using the `--watch` flag to update on workspace changes.

Would be nice to be able to configure something similar via `null-ls`.

jose-elias-alvarez:
@jryom Could you try out #351 and let me know what you think? Unfortunately the architecture of null-ls makes it impossible to get output from a continuously-running process the way `coc-tsserver` does, so the experience is definitely worse, but I'd like to know if you think it's useful enough to justify supporting it in spite of the limitations I describe there.

jose-elias-alvarez:
This feature is now supported in #376, which I plan on merging once null-ls fully migrates to Neovim 0.6.0.

PR #351, opened 2021-11-16
feat: support project diagnostics

jose-elias-alvarez:
Early attempt at closing #194.

This PR adds support for project diagnostics and a rudimentary `tsc` source to server as a proof-of-concept. Running `:lua require("null-ls.diagnostics").set_project_diagnostics()` will run registered generators for the new method based on the current filetype, load buffers by filename, set diagnostics per buffer, and send results to the quickfix list.

As a simple way to spawn processes and send their output to `vim.diagnostic`, it works, but it has quite a few issues:

- Each run is very, very slow (nearly 10 seconds on a small-to-medium TypeScript project). Diagnostics only update upon receiving generator results again, which is painful.
- ~~If an update removes all diagnostics in a buffer, `project_diagnostics()` won't clear stale diagnostics in that buffer.~~ Sort of solved by clearing diagnostics from all available sources before each run.
- Sources like `tsc`, which overlap with `tsserver` diagnostics, are particularly confusing, since the user sees two sets of identical diagnostics, one that updates on change... and one that mysteriously doesn't.
- null-ls requires processes to close in order to capture their output, so we can't do something like run `tsc --watch` and update when it outputs to `stdout`.

Because of these issues, I think there's a high change that users will conclude that something is broken. Sending diagnostics to the quickfix list is pretty convenient, and I can imagine a workflow where users fetch project diagnostics, solve issues using the quickfix list, then refresh them, but it's not ideal.

I'm going to sit on this for a bit and see I can come up with something better, but I'd say there's a 50% chance that I don't merge this because I'm not sure we can give users a good experience. Feedback / ideas / contributions are also very wellcome.


jose-elias-alvarez:
I used this yesterday to fix issues in a project, and in spite of its limitations it's pretty useful. I'm considering merging it as an experimental feature, but I wonder if the only reason I'm okay with the limitations is because I know why we have them. I still need to think more.

xenrox:
Hello, as you proposed in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/362 I tested this PR with staticcheck (https://github.com/xenrox/null-ls.nvim/tree/project-sc). This seems to work quite well, but I noticed a bug: Running `lua require("null-ls.diagnostics").get_project_diagnostics() ` does not only open the diagnostics in the quickfix list but in a new empty buffer as well. Closing this empty buffer, causes repeated runs of this command to fail with: 
```
E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/diagnostic.lua:925: Invalid bu
ffer id: 3                                                                               
stack traceback:                                                                         
        [C]: in function 'nvim_buf_clear_namespace'                                      
        /usr/share/nvim/runtime/lua/vim/diagnostic.lua:925: in function 'hide'           
        /usr/share/nvim/runtime/lua/vim/diagnostic.lua:1068: in function 'hide'          
        /usr/share/nvim/runtime/lua/vim/diagnostic.lua:1342: in function 'reset'         
        ...ck/packer/start/null-ls.nvim/lua/null-ls/diagnostics.lua:34: in function 'hide
_source_diagnostics'                                                                     
        ...ck/packer/start/null-ls.nvim/lua/null-ls/diagnostics.lua:208: in function 'cle
ar_project_diagnostics'                                                                  
        ...ck/packer/start/null-ls.nvim/lua/null-ls/diagnostics.lua:160: in function 'get
_project_diagnostics'                                                                    
        [string ":lua"]:1: in main chunk                                
```
I will try to debug this issue further to check if maybe another plugin or setting is interfering.
Furthermore I would like a setting to disable the auto opening of the quickfix list and just work with the opened buffers.

jose-elias-alvarez:
@xenrox Thanks for trying this out! I don't use Go but am trying your branch out with [this sample repository](https://github.com/golang/example) and it's working well without the issue you're describing. Could it be that one of your diagnostics doesn't have a filename associated with it? That could result in creating an empty buffer. If that's the case, we should validate the filename when receiving diagnostics and also make sure the buffer is loaded when clearing diagnostics. 

I also agree that automatically opening the quickfix list should be optional, so I'll work on that too. In addition to the Lua API, I'm thinking of a `:NullLsProjectDiagnostics` command with a `!` version to not open the quickfix list – does that make sense to you?

xenrox:
> @xenrox Thanks for trying this out! I don't use Go but am trying your branch out with [this sample repository](https://github.com/golang/example) and it's working well without the issue you're describing. Could it be that one of your diagnostics doesn't have a filename associated with it? That could result in creating an empty buffer. If that's the case, we should validate the filename when receiving diagnostics and also make sure the buffer is loaded when clearing diagnostics.

The problem was caused by a plugin that modified the quickfix list. What still causes problems now is:
1. I generate project diagnostics
2. I close the quickfix list
3. Rerun diagnostic generation
4. Run into same error message as above

> I also agree that automatically opening the quickfix list should be optional, so I'll work on that too. In addition to the Lua API, I'm thinking of a `:NullLsProjectDiagnostics` command with a `!` version to not open the quickfix list – does that make sense to you?

Yes that sounds nice, although I would probably go more with something like `:NullLsProjectDiagnostics quickfix=false` (similar to how you can e.g. use themes in packer) in case you want to configure more settings in the future.

jose-elias-alvarez:
Hmm, I still can't replicate the issue. Based on your descriptions, it seems like there might be two issues:

1. One or more of the diagnostics is missing a filename, which causes an empty buffer to be loaded. We can add a simple level of validation to verify that `filename` is not `nil` and is not an empty string, but ultimately each source would be responsible for providing valid (and non-relative) filenames, which might be tricky. 
2. I suspect there may be an upstream issue with the diagnostic API, since a user ran into a similar issue with #350. I opened neovim/neovim#16391 with a description and reproduction.

Let me know if either of those explanations makes sense given what you are seeing.

xenrox:
I should not be the first issue, since I am running it on the same example repository as you and the diagnostics aren't missing any filenames. And I don't have problems with an empty buffer anymore but with closing the quickfix list. I attached a gif in case my description is a bit lacking.
record_1637520998

jose-elias-alvarez:
Are you also using bufferline.nvim? The issue in #350 turned out to be related to that, so I wonder if that's what's going on in your case, too. Either way I'll wait to see what happens with the upstream issue and keep working on this in the meantime.

xenrox:
Yes you are right, without bufferline it works as expected. Thanks for your help!

xenrox:
Works as expected with https://github.com/neovim/neovim/pull/16397.

jose-elias-alvarez:
Closing this in favor of #376.

PR #352, opened 2021-11-16
feat(formatting): add google-java-format

mtoohey31:
This PR adds a builtin formatting source for [google-java-format](https://github.com/google/google-java-format).

I believe I've followed the format for documentation properly, the only thing that might be a little out of the ordinary is that I added another section inside usage showing how to use it with the `--aosp` option since I think many people will want to do that. If there's a different heading that should go under, lmk!

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -641,6 +641,34 @@ local sources = { null_ls.builtins.formatting.golines }
 - `command = "golines"`
 - `args = {}`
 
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
I think this is a good place for the extra instructions if you think it's going to be a common request. Thank you!

mtoohey31:
> I think this is a good place for the extra instructions if you think it's going to be a common request.

Sounds good, I think it will be cause the default indentation is 2 spaces which is a little cramped.

Happy to help!

PR #353, opened 2021-11-16
perf: add update_on_insert option

### LINKED ISSUE #199: Avoid updating diagnostics in insert mode, part 2
#136 documents my first failed attempt to implement this. Since then, I've come up with the idea to handle this in the RPC module, where we could check the current mode, store notifications in state, and unconditionally flush any stored notifications on `InsertLeave`. This allows us to leave the diagnostics module untouched and could potentially be expanded to other methods in the future.

Remaining potential issues:
- Checking the current mode could be unreliable if there is a delay between changing the text and receiving the notification
- Making sure that editing the text and exiting insert mode doesn't trigger a double notification

I'll start prototyping this and open a PR soon.

jose-elias-alvarez:
Closes #199.

This was suspiciously easy to implement and seems to have no ill effects apart from the double notification issue. It also provides a gigantic performance boost when using an autocompletion plugin like nvim-cmp + a null-ls completion source, since it suppresses the LSP `didChange` notifications that would otherwise be sent on every keystroke.

I need to use this for a while and see if there's any bad behavior or edge cases that I'm missing, and as before I'd appreciate input from anyone having performance issues about whether this actually makes a difference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,13 +180,6 @@ local markdownlint = {
 null_ls.register(markdownlint)
 ```
 
-### Advanced usage
-
-This [ESLint
-integration](https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils/blob/main/lua/nvim-lsp-ts-utils/null-ls.lua)
-from one of my plugins demonstrates an advanced example of parsing JSON output
-from a command to generate code actions.
-
 ## FAQ
 
 ### Something isn't working! What do I do?
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1506,13 +1506,6 @@ local sources = { null_ls.builtins.diagnostics.cspell }
 
 A linter for the `javascript` ecosystem.
 
-- Note that the `null-ls` builtin requires your `eslint` executable to be
-  available on your `$PATH`.
-- To use local (project) executables, use the integration in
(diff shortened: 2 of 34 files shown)


numToStr:
This would also likely solve `eslint_d` diagnostics being very slow. I would be happy to give this is a shot :)

jose-elias-alvarez:
@numToStr Do you specifically experience lag when typing in insert mode? If so, let me know if you're able to try this PR out locally and if it makes a difference. I haven't had any performance issues so I really don't know if this helps.

jose-elias-alvarez:
@numToStr Sure, just rebased.

Are you using an autocomplete plugin? I've always experienced terrible performance with `typescript-language-server` + autocompletion, so I'd investigate that too. Either way, if the changes in this PR help performance a bit and don't have any downsides, I'd be happy to merge it.

numToStr:
I am using nvim-cmp for the completion.

> if the changes in this PR help performance a bit and don't have any downsides, I'd be happy to merge it.

Sure. I'll let you know about my findings :)

numToStr:
@jose-elias-alvarez I haven't faced any issues or any bugs. So, I think you can merge this if you want :)

jose-elias-alvarez:
Great! I've also been using it without any obvious issues so I'm going to merge it for now and see if any issues pop up when more people start using it.

PR #357, opened 2021-11-16
fix golangci-lint issues

marcomayer:
I had mentioned all of this in code-review comments in #318, for details please have a look there.
Here are the changes I suggested.

It's necessary to ignore stderr from golangci-lint for now as otherwise the deprecation warnings of various linters used by golangci-lint will cause issues.

Also as most of the linters show non-critical issues in the code, settings the default severity to warning makes sense. Otherwise all the warnings will show up as errors.


jose-elias-alvarez:
Thanks! Happy to get this in a separate PR from an actual user of the source.

PR #358, opened 2021-11-17
docs(builtins): fix `command` for `eslint_d` diagnostics

numToStr:


jose-elias-alvarez:
Whoops, thanks for catching this!

PR #360, opened 2021-11-18
Add PHP -l syntax checker

euoia:


lua/null-ls/builtins/diagnostics/php.lua
```diff
@@ -0,0 +1,31 @@
...
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    method = DIAGNOSTICS,
+    filetypes = { "php" },
+    generator_opts = {
+        command = "php",
+        to_stdin = false,
```
gbprod, line 11: For better performance, I think we should use stdin.

Try with `to_stdin = false`, `args = { "-l" }` and remove `to_temp_file`

lua/null-ls/builtins/diagnostics/php.lua
```diff
@@ -0,0 +1,31 @@
...
+        to_temp_file = true,
+        args = { "-l", "$FILENAME" },
+        format = "line",
+        check_exit_code = function()
+            return true
```
gbprod, line 16: [Documentation](https://www.php.net/manual/en/features.commandline.options.php) says that the shell result is `-1` if errors are found.
I think we should `return -1 == code` here

lua/null-ls/builtins/diagnostics/php.lua
```diff
@@ -0,0 +1,31 @@
...
+        end,
+        on_output = h.diagnostics.from_patterns({
+            {
+                -- Parse error: syntax error, unexpected '$appends' (T_VARIABLE), expecting function (T_FUNCTION) or const (T_CONST) in app/Config.php on line 16
+                pattern = [[Parse error: (.*) (%d+)]],
```
gbprod, line 21:
```suggestion
                pattern = [[PHP Parse error: (.*) (%d+)]],
```

lua/null-ls/builtins/diagnostics/php.lua
```diff
@@ -0,0 +1,31 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
```
gbprod, line 6: php linters errors are written to `stderr`, we should add `from_stderr = true`

lua/null-ls/builtins/diagnostics/php.lua
gbprod: much better : 

`pattern = [[PHP Parse error: (.*) in Standard input code on line (%d+)]],`

gbprod:
I've tried with this settings today and it works great : 

```lua
    generator_opts = {
        command = "php",
        to_stdin = true,
        args = { "-l" },
        format = "line",
        from_stderr = true,
          return code < 0
        end,
        on_output = h.diagnostics.from_patterns({
            {
                -- Parse error: syntax error, unexpected '$appends' (T_VARIABLE), expecting function (T_FUNCTION) or const (T_CONST) in app/Config.php on line 16
                pattern = [[PHP Parse error: (.*) in Standard input code on line (%d+)]],
                groups = { "message", "row" },
                overrides = {
                    diagnostic = { severity = h.diagnostics.severities["error"] },
                    offsets = { col = 1, end_col = 1 },
                },
            },
        }),
    },
```

lua/null-ls/builtins/diagnostics/php.lua
euoia: I think the "PHP" prefix may depend on php version. Which PHP version are you using? Which should we target?

euoia:
Is it preferable to pipe to stdin rather than just check the file as it is on disk? I would think the latter would be preferable.

Which version of PHP are you testing with, since there may be subtle differences between versions.

euoia:
I've tested using PHP 7.4 and PHP 8.0.

Regarding the exit code, I get 0 when there are no errors and 255 when there are errors. There's an exit code of 1 when the file doesn't exist. I'm not sure the shell will ever report a negative exit code.

Regarding the output, I am not seeing "PHP" at the start of the parse error message. On which  version of PHP are you seeing that? Of course, we can ignore it so both formats work.

Regarding use of stdin, see my comment above. I agree we should not create a temp file, but would expect it would be better to use the file on disk rather than pass to stdin. Of course happy to defer to your judgement.

Here's my updated version (not using stdin). I just need to optionally match the PHP at the start:

```

return h.make_builtin({
    method = DIAGNOSTICS,
    filetypes = { "php" },
    generator_opts = {
        command = "php",
        to_stdin = false,
        to_temp_file = false,
        args = { "-l", "$FILENAME" },
        format = "line",
            -- Code 0 means no syntax errors.
            -- Code 255 means syntax errors.
            -- Other codes mean something went wrong.
            return code == 0 or code == 255
        end,
        on_output = h.diagnostics.from_patterns({
            {
                -- Parse error: syntax error, unexpected '$appends' (T_VARIABLE), expecting function (T_FUNCTION) or const (T_CONST) in app/Config.php on line 16
                pattern = [[Parse error: (.*) in (.*) on line (%d+)]],
                groups = { "message", "junk", "row" },
                overrides = {
                    diagnostic = { severity = h.diagnostics.severities["error"] },
                    offsets = { col = 1, end_col = 1 },
                },
            },
        }),
    },
})
```

Can you also advice on whether using the "junk" group is the correct way to do this?

jose-elias-alvarez:
> Is it preferable to pipe to stdin rather than just check the file as it is on disk? I would think the latter would be preferable.

The issue with the latter is that diagnostics won't update until the file is saved. If supported, `stdin` will almost always provide a better experience (without the overhead of managing temp files). 

I can't comment on the PHP-specific stuff, but to answer the version question: Generally speaking (and I'll add a note about this to the documentation) we should target the latest version of linters unless there's a compelling and widespread reason to use a previous version. If previous versions can be supported by modifying arguments, we can mention that in the built-in documentation, but if they require a different parser, we'd need to create separate built-ins.

gbprod:
About exit codes, I've made some tests and you're right :) (I've tried with php 8.0)

lua/null-ls/builtins/diagnostics/php.lua
gbprod: Using stdin, on PHP 8.0, I got this : 

```bash
$ cat test.php | php -l
PHP Parse error:  syntax error, unexpected token "}", expecting ";" in Standard input code on line 18
Errors parsing Standard input code
```
Better to avoid the "PHP" prefix as the regex will work for both case.

gbprod:
Thanks for this, I'll make some tests with different versions and make a feedback asap

lua/null-ls/builtins/diagnostics/php.lua
euoia: How odd that our output is slightly different:
```
$ cat app/Config.php | php -l

Parse error: syntax error, unexpected token "=>", expecting "," or ";" in Standard input code on line 21
Errors parsing Standard input code
```

```
$ php -v
PHP 8.0.12 (cli) (built: Nov  5 2021 13:11:36) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.12, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.12, Copyright (c), by Zend Technologies
```

In any case, as you say, without the PHP prefix the regex will match both - I tested that.

euoia:
Based on feedback, I've updated the PR. Please have another look.

jose-elias-alvarez:
Thanks for working on this – will merge once @gbprod approves.

gbprod:
Hi! Made some test with this new version but it doesn't works for me :confused: 
It only works if I add `from_stderr = true,`, can you try with it @euoia ?
I will continue to test today and find out why it's not working for me...

lua/null-ls/builtins/diagnostics/php.lua
gbprod: Really strange ! Maybe some conf in php.ini ?

```
$ php -v
PHP 8.0.13 (cli) (built: Nov 22 2021 09:47:37) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.13, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.13, Copyright (c), by Zend Technologies
```

gbprod:
Ok ! I think I've got it ! Seems that's it's the `display_errors` php.ini setting.

```
$ php -i | grep display_errors
display_errors => Off => Off
```
IMHO, we should add `from_stderr = true` in generator settings and add `-d display_errors=stderr` to the php command : 

```lua
-- Tested using PHP 7.x and PHP 8.x.

return h.make_builtin({
    method = DIAGNOSTICS,
    filetypes = { "php" },
    generator_opts = {
        command = "php",
        -- Send file to stdin or else checking will only be done when the file is saved.
        to_stdin = true,
        from_stderr = true,
        args = { "-l", "-d", "display_errors=Off" },
        format = "line",
            -- Code 0 means no syntax errors.
            -- Code 255 means syntax errors.
            -- Other codes mean something went wrong.
            return code == 0 or code == 255
        end,
        on_output = h.diagnostics.from_patterns({
            {
                -- Example of an error when checking a file:
                -- Parse error: syntax error, unexpected '$appends' (T_VARIABLE), expecting function (T_FUNCTION) or const (T_CONST) in app/Config.php on line 16

                -- Example of an error when checking stdin:
                -- Parse error: syntax error, unexpected token "=>", expecting "," or ";" in Standard input code on line 21

                -- This pattern should match both.
                pattern = [[Parse error: (.*) in (.*) on line (%d+)]],
                groups = { "message", "junk", "row" },
                overrides = {
                    diagnostic = { severity = h.diagnostics.severities["error"] },
                    offsets = { col = 1, end_col = 1 },
                },
            },
        }),
    },
})
```

euoia:
Interesting! Here's my unmodified display_errors configuration on Mac OS 12.0.1 with PHP 8.0.13 installed via Homebrew:

```
$ php -i | grep display_errors
display_errors => STDOUT => STDOUT
```

If I set `display_errors=Off` I just get a generic error message instead of a specific syntax error:
```
$ php -d display_errors=Off -l app/Config.php
Errors parsing app/Config.php
```

And further, the output is still to stdout (redirecting to `/dev/null` should still show sent to stderr):
```
$ php -d display_errors=Off -l app/Config.php > /dev/null
$
```

Setting `display_errors=Stdout` or `display_errors=Stderr` does work expected though. So I think we should either:
1.  `from_stderr=false` and `display_errors=Stdout`
2. `from_stderr=true` and `display_errors=Stderrr`

I don't really know which is best, I guess it depends on whether php -l will put other stuff on either stdout or stderr that could interfere with the pattern matching.

gbprod:
It's a mess but it's a really interesting problem.

Options 1 and 2 are both working for me but... all errors are duplicated ! Maybe it's another setting... I've thought about `log_errors` directive and it seems to fix.

Maybe we just should use default php settings for settings that can change linter's behavior that are : 

    display_errors=stdout
    log_errors=Off

That's means that my preference will go to option 1 + `-d log_errors=Off`.
That works for me.

What do you think about that?

euoia:
How strange. When you say duplicated, can you show me what you mean? I assume you are saying that every error line is shown repeated twice. My php.ini enables log_errors:
```
$ php -i | grep log_errors
log_errors => On => On
log_errors_max_len => 1024 => 1024
```

Running the linter with log_errors On or Off doesn't seem to make any difference to the output on Stdout:
```
$ php -d display_errors=Stdout -d log_errors=Off -l app/Config.php

Parse error: syntax error, unexpected token "=>", expecting "," or ";" in app/Config.php on line 21
Errors parsing app/Config.php
$ php -d display_errors=Stdout -d log_errors=On -l app/Config.php

Parse error: syntax error, unexpected token "=>", expecting "," or ";" in app/Config.php on line 21
Errors parsing app/Config.php
```

Setting log_errors=Off does prevent the linting error being logged to my php-fpm.log, though. This seems consistent with the documentation that refers to the error log (and not to stdout or stderr):
https://www.php.net/manual/en/errorfunc.configuration.php#ini.log-errors

I wonder if it's a different setting, or something to do with the lint error itself, that's producing the duplicated errors for you?

gbprod:
Ok!
Thanks for the feedback (and sorry for my late answer).

I've figured out why I was getting all the errors twice : my `error_log` was set to null (which meens stderr).

My opinion is to set :

    display_errors=stdout
    log_errors=Off

to avoid local configuration based error log (and avoid logging to any log file). Do you agree ?

euoia:
That all makes sense, and yes, I agree those options look good. I've pushed the changed to the PR.

doc/BUILTINS.md
```diff
@@ -862,6 +862,24 @@ local sources = { null_ls.builtins.formatting.perltidy }
...
+
+```lua
+local sources = { null_ls.builtins.diagnostics.php }
+```
+
+##### Defaults
+
+- `filetypes = { "php" }`
+- `command = "php"`
+- `args = { "-l" }`
```
gbprod, line 881: I think we should update doc

gbprod (approved):
It's ok for me !

jose-elias-alvarez:
Great, thank you @gbprod for the thorough review! Once we update the documentation this looks good to me. (Since you already have a couple of examples in the comments for the built-in, you may also want to consider adding test coverage, but it's not strictly necessary.)

euoia:
Ok, I updated the documentation. Let me know if you'd like me to have a go at adding tests.

jose-elias-alvarez:
I think this is good as-is. Thanks again to both of you for working on this and figuring it out!

PR #361, opened 2021-11-18
fix(nginx_beautifier): wrong output

captainko:
Attempt to fix nginx-beautifier wrong output


jose-elias-alvarez:
Looks like this was pretty broken to begin with. Thank you for the fix!

PR #362, opened 2021-11-19
feat(diagnostics): add staticcheck

xenrox:
This PR will add staticcheck. Compared to the implementation I used for [nvim-lint](https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/linters/staticcheck.lua), the whole project will be linted instead of a single file. This prevents false positives like `undeclared name`.


jose-elias-alvarez:
Great, thank you! I think this would also be a good test case for #351 if you're willing to give that a shot, but this makes sense to run on save, so I'm happy to merge it as-is for now.

PR #369, opened 2021-11-20
feat(diagnostics): add revive

xenrox:
Like most golang linters revive can lint a whole project at once as well. So if you implement https://github.com/jose-elias-alvarez/null-ls.nvim/pull/351, I would probably adjust this.


jose-elias-alvarez:
Sure, does it make sense to merge this now and update it later if needed? I'm still not 100% sure about merging project diagnostics so that may be better but let me know.

xenrox:
Yes then it makes sense to merge it like this and I will update it in case you merge the project diagnostics feature.

PR #370, opened 2021-11-21
Additional message handlers for the flake8 diagnostic builtin

### LINKED ISSUE #365: Add ability to catch more flake8 codes
### Issues

### Feature description

Flake8 has a somewhat unusual severity/coding system in that it does not provide a more traditional info, warn, error style of message classification. The code system instead tags the origin of which flake8 plugin generated the diagnostic message. 

This presents a couple of difficulties when combined with null-ls. There doesn't seem to be an out of the box way to modify the `on_output` setting of a builtin, requiring replacing the entire builtin if you wanted to change the severity listing of a default source (in my case move F messages to warning or error). 

Additionally, flake8 has an extensive set of plugins which provide their own letter prefixes (bugbear for example uses B). The existing builtin ends up ignoring these messages as there is no default or catch-all severity handler. 

Neither of these is particularly problematic for me to work around as the solution to both is define my own diagnostic source based on the builtin, but there could be room for improvement. 

See https://flake8.pycqa.org/en/latest/user/violations.html#selecting-violations-with-flake8 and https://flake8.pycqa.org/en/latest/user/error-codes.html for some references on flake8 codes.

and https://github.com/PyCQA/flake8-bugbear

### Help

### Implementation help

A simple solution that would get most people 99% of the way there would perhaps be just to audit the existing flake8 plugin ecosystem and provide default severities for each currently used letter prefix code. It might also be preferable to change the current default severity of the F codes in particular, but it may be undesirable to make that change for other existing users. 

If we'd like to provide defaults for more error codes out of the box I'd be happy to do that leg work and upstream it.

bwells:
Adds error message capture for some of the most popular flake8
extensions.

Closes #365

### DIFF
--- a/lua/null-ls/builtins/diagnostics/flake8.lua
+++ b/lua/null-ls/builtins/diagnostics/flake8.lua
@@ -27,7 +27,7 @@ return h.make_builtin({
                     R = h.diagnostics.severities["warning"],
                     S = h.diagnostics.severities["warning"],
                     I = h.diagnostics.severities["warning"],
-                    C = h.diagnostics.severities["warning"]
+                    C = h.diagnostics.severities["warning"],
                 },
             }
         ),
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thank you for working on this!

PR #371, opened 2021-11-22
Specify filename when calling isort

cblessing24:
Currently `isort` is formatting stub files (`.pyi` extension) incorrectly because it doesn't know whether the input is a stub file or a regular Python file. This PR fixes that problem by specifying the filename when calling `isort`.


jose-elias-alvarez:
Great, thanks!

sirupsen:
@cblessing24 which version of `isort` are you using? I installed a recent version with `pip` and don't have `--filename`, so this didn't work for me. I copy-pasted the builtin and removed the `--filename $FILENAME` and it works for me.

PR #376, opened 2021-11-24
feat(diagnostics): project diagnostics v2

### LINKED ISSUE #194: Provide diagnostics for closed files in workspace

jose-elias-alvarez:
Closes #194.

While working on #351, I realized that a better way to support project diagnostics might be to allow ~~any~~ diagnostics sources to specify a filename / buffer number and use that to determine where to send diagnostics. This is a more logical approach for linters that run on save, such as the Go project linters we've recently added support for, and it removes the need for additional commands / special handling. **Update**: Sources must now use the `multiple_files` flag to signal that their results should apply to more than one file, but everything else worked out as planned.

~~It might still make sense to add a command that sends a synthetic notification to run project linters and open the quickfix list. The source API makes it easier to enable and disable sources, but running them on demand is still not that friendly, so a dedicated command might help here. This would also let users determine whether to run these sources on save or on demand depending on their workflow and environment.~~ **Update**: I'm not planning on implementing this, at least in this PR. Running on save is more "LSP-like" and provides a better experience. 

~~As a consequence of how null-ls runs generators, registering a quick on-change source and a slow on-save source delays diagnostics until both sources have returned. It might also be a good idea to wait until 0.6 release and refactor so that diagnostics are published as they're received, which should speed things up.~~ **Update**: This PR now depends on #388, which removes 0.5.1 support. Diagnostics are now published as they're received (instead of waiting for all generators to finish running), which is a nice improvement overall. 

This PR also includes a `tsc` source, which works the same but runs on save and now includes a simple check to mitigate the duplicate diagnostics issue. 

I'm pretty happy with this now, so I plan on using this for a couple of weeks until I can merge #388. Further feedback / testing from anyone who happens to see this would be greatly appreciated.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2024,6 +2024,28 @@ local sources = { null_ls.builtins.diagnostics.staticcheck }
 - `command = "staticcheck"`
 - `args = { "-f", "json", "./..." }`
 
+#### [tsc](https://www.typescriptlang.org/docs/handbook/compiler-options.html)
+
+##### About
+
+Parses diagnostics from the TypeScript compiler.
+
+##### Usage
+
+- Diagnostics from this source and `tsserver` are independent. If you have
+  `tsserver` configured to show diagnostics, you may see duplicates until buffer
+  diagnostics update.
+
+```lua
+local sources = { null_ls.builtins.diagnostics.tsc }
+```
+
+##### Defaults
+
+- `filetypes = { "typescript", "typescriptreact" }`
+- `command = "tsc"`
+- `args = { "--pretty", "false", "--noEmit" }`
+
 ### Code actions
 
 #### [ESLint](https://github.com/eslint/eslint)
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1956,8 +1956,8 @@ local sources = { null_ls.builtins.diagnostics.yamllint }
 **NOTE**: These sources depend on Neovim version 0.6.0 and are not compatible
 with previous versions.
 
-These sources run **only** on save, meaning that the diagnostics you see will
-not reflect changes to the buffer until you write the changes to the disk.
+These sources **do not run on change**, meaning that the diagnostics you see
+will not reflect changes to the buffer until you write the changes to the disk.
 
 You can configure built-in diagnostic sources to run on save by overriding
 `method`:
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -158,6 +158,19 @@ See [CONFIG](CONFIG.md) to learn about the structure of `diagnostics_format`.
 Note that specifying `diagnostics_format` for a built-in will override your
 global `diagnostics_format` for that source.
 
+### Diagnostics performance
+
+If you have performance issues with a diagnostic source, you can configure any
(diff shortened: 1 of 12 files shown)


jose-elias-alvarez:
The current implementation has a big flaw, namely that we don't know whether a source intended to send diagnostics for all buffers (in which case we should wipe out all existing diagnostics, then reset them) or for just one buffer (in which case we should only wipe out and reset diagnostics for the buffer in questions). That means sources need to opt in to multi-file diagnostics, which is a little less intuitive but most likely necessary. I'll continue working on the PR this week and make a decision about merging it.

jose-elias-alvarez:
@xenrox If you have the time / inclination to test this out with your Go sources, I'd really appreciate it. Adding support should be pretty trivial – you'd just need to set `multiple_files = true` and specify either a `bufnr` or `filename` for each diagnostic (you can use the [`tsc` source](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/project-diagnostics-v2/lua/null-ls/builtins/diagnostics/tsc.lua) as an example).

xenrox:
@jose-elias-alvarez: I have added support here: https://github.com/xenrox/null-ls.nvim/tree/project-diagnostics-go
Will test in the coming days.

xenrox:
So I have worked with project diagnostics for a few hours now and I quite like the usability. It makes it a lot easier to actively work on go projects to e.g. see if a change in one file introduced a regression in another one (with telescope).

jose-elias-alvarez:
Great to hear, no issues then? I've also found it quite useful and think it's a solid workflow for on-save linters. If everything is working for you too, I'll merge this after #388.

xenrox:
No, couldn't find any bugs. Thanks for your work!

Tylerc230:
Does `eslint` or `eslint_d` support `multiple_files`? I see it working for `tsc` but not these two.

jose-elias-alvarez:
> Does `eslint` or `eslint_d` support `multiple_files`? I see it working for `tsc` but not these two.

They do not - this is theoretically possible to add but would require adding separate sources, since we don't have a good way for a single built-in to support both cases. Vanilla ESLint would also be *extremely* slow when running across a whole project.

I'm generally hesitant to improve these sources given that the [ESLint language server](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#eslint) is in almost all cases a better option, but feel free to open an issue or PR if you think this would be useful.

toto1384:
Any updates on the multiple_files ?

jose-elias-alvarez:
> Any updates on the multiple_files ?

This needs to be explicitly supported by sources and is [already supported by a handful](https://github.com/jose-elias-alvarez/null-ls.nvim/search?q=multiple_files).

toto1384:
It is supported for tsc but i couldn't get it working

Issue #378, opened 2021-11-26
Cannot use null-ls with the latest `neovim/nvim-lspconfig`


captainko:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

  NVIM v0.6.0-dev+634-geb876a0a6 

### null-ls config

```lua
null_ls.config {
	debug = true,
	sources = {...}
}
lspconfig["null-ls"].setup { ... }
```

### Steps to reproduce

Run require('lspconfig')['null-ls'].setup {...}

### Debug log

```
Cannot access configuration for null-ls. Ensure this server is listed in `server_configurations.md` or added as a custom server.
packer.nvim: Error running config for null-ls.nvim: ...pack/packer/start/null-ls.nvim/lua/null-ls/lspconfig.lua:100: attempt to index field 'manager' (a nil value)
```

### Help

No

mjlbach:
Relevant lines of code for indexing the `lspconfig` table are here: https://github.com/neovim/nvim-lspconfig/blob/b2d302f1609a8267f4b855d8853fba65669aa8e5/lua/lspconfig.lua#L72-L85

numToStr:
So, the issue might be related to [rpc.lua](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/e76aefca283032e1d85a8e3094531124ff8172d3/lua/null-ls/rpc.lua#L56) looking for `null-ls` even before [registering](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/e76aefca283032e1d85a8e3094531124ff8172d3/lua/null-ls/init.lua#L31) itself.

jose-elias-alvarez:
@numToStr Thanks for looking into this, I’m with my family and away from my computer for the weekend so I can’t fix this directly. Does switching the order of the lines in init.lua make a difference? 

If the changes mean that the index metamethod shows an error message if the server is not yet configured, we either need to manually track whether the server has been configured or use a different method to access the config object. Reverting the rpc changes in 4e9e786e7621dcbe0c3d8d473fc434a2a2d2687b may also help.

mjlbach:
[...]

PR #381, opened 2021-11-26
fix: resolve root_dir for single files

kylo252:


kylo252:
@jose-elias-alvarez, I needed to make a small change to handle root_dir for single file mode. Please let me know if it fixes the issue.

jose-elias-alvarez:
I’m away for the weekend and won’t have access to a computer until Monday, so I can’t properly review this. Ideally we could just fix the indexing issue from #378 for now then go over these other changes later. Is that possible?

kylo252:
> I’m away for the weekend and won’t have access to a computer until Monday, so I can’t properly review this. Ideally we could just fix the indexing issue from #378 for now then go over these other changes later. Is that possible?

Was already doing what you wanted, see #382 :)

jose-elias-alvarez:
@kylo252 We can continue working on this if you like. Single-file support seems like a natural fit for null-ls, but I'm not familiar with the advantages / disadvantages.

kylo252:
> @kylo252 We can continue working on this if you like. Single-file support seems like a natural fit for null-ls, but I'm not familiar with the advantages / disadvantages.

It's basically that `lspconfig` has sort of dropped support for using `dirname`, there's even a CI check that will get triggered if you attempt to add it back again 😄 

Other than that, I'm not sure that there's much between them. 

One thing I've been having trouble with currently is that `luacheck` doesn't update its cwd, so it's not reading the current project `.luacheckrc`.  I'm going to test to see if this PR helps, but I'll open a new ticket to track it anyway.

jose-elias-alvarez:
It's very likely that I'll merge #389 soon, which should make this irrelevant. Not sure about the Luacheck `cwd` problem, but feel free to open an issue with a reproduction, since everything related to `root_dir` should work the same way.

PR #382, opened 2021-11-26
fix(lspconfig): update the setup workflow

kylo252:
Note: the single-files changes are now tracked separately.

### DIFF
--- a/lua/null-ls/init.lua
+++ b/lua/null-ls/init.lua
@@ -28,8 +28,8 @@ M.config = function(user_config)
     end
 
     c.setup(user_config or {})
-    require("null-ls.rpc").setup()
     require("null-ls.lspconfig").setup()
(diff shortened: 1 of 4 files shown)


jose-elias-alvarez:
Great, thanks! @numToStr Could you quickly verify that this doesn’t have the same RPC cmd issue?

kylo252:
The `format_on_save` now works. It also works with single files. I had just missed adding a check against this: https://github.com/jose-elias-alvarez/null-ls.nvim/pull/381/files#diff-f3780c8e3c0176358f1aff548b6d5267d747e62dc474b039d4a93e7d886ae008R104-R106

jose-elias-alvarez:
Okay, cool. Let’s merge this for now and take a look at those changes again when I can.

PR #384, opened 2021-11-28
feat(builtins): add shellcheck code actions

b0o:
Adds shellcheck code actions to disable the rule under the cursor, either at the line level or the file level. If a `disable=` directive already exists, the rule’s code is appended to it.

Note: Shellcheck allows arbitrary blank lines between line directives and the target line or the shebang/BOF and the file-level directive. This case is currently not handled intelligently; a new directive is added rather than updating the existing one.

I didn't add tests because I don't see any tests for other code actions builtins.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2134,6 +2134,24 @@ local sources = { null_ls.builtins.code_actions.refactoring }
 
 - `filetypes = { "go", "javascript", "lua", "python", "typescript" }`
 
+#### [shellcheck](https://www.shellcheck.net)
+
+##### About
+
+A shell script static analysis tool.
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.code_actions.shellcheck }
+```
+
+##### Defaults
+
+- `filetypes = { "sh" }`
+- `command = "shellcheck"`
+- `args = { "--format", "json1", "-" }`
+
 #### [statix](https://github.com/nerdypepper/statix)
 
 ##### About
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1796,7 +1796,7 @@ local sources = { null_ls.builtins.diagnostics.shellcheck }
 
 - `filetypes = { "sh" }`
 - `command = "shellcheck"`
-- `args = { "--format", "json1", "-" }`
+- `args = { "--format", "json1", "--source-path=$DIRNAME", "--external-sources", "-" }`
 
 #### [standardrb](https://github.com/testdouble/standard)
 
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
This is a cool idea and works well! The way Shellcheck handles blank lines is pretty weird. I guess we could get all the lines above the current line and iterate over them backwards to find existing comments if you think it's worth it (I personally would  
not leave blank lines in this situation).

b0o:
Ok, I think this is ready for review. Improvements in latest commit:

- Arbitrary number of blank or comment-only lines are allowed between disable directives and the target line
- Running a `Disable Line` action should not append to a file-level directive
- Refactored and cleaned up the code

jose-elias-alvarez:
I pushed a commit fixing the arguments in the documentation, since they were out-of-date for the diagnostics source, too. Also, I think it might be a good idea to describe exactly what the code actions do, since they're synthetic (i.e. not something Shellcheck can normally do). 

I see what you're saying about the first non-comment line. I wonder if it's possible to improve the behavior a bit, though. As a minimal example:

```sh
#!/bin/bash

for f in ".txt"; do
	echo "file"
done
```

Applying the 2 code actions to fix the 2 broken rules in the first line of the loop results in this:

```sh
#!/bin/bash

# shellcheck disable=2034
# shellcheck disable=2043
for f in ".txt"; do
	echo "file"
done
```

If we could change the heuristic so that the disable directives are merged (regardless of whether they're treated as file-level or line-level directives), it might be a nice improvement. I don't think it's a big deal, though, since it's an edge case and the syntax is simple enough that fixing it manually is quick.

b0o:
So after re-reading the ShellCheck documentation[^1] and doing some testing, it seems **every** directive between the beginning of the file and the first non-comment line is treated as file-level. I originally thought this only applied to the uppermost directive. I don't really like this behavior of ShellCheck's (ideally you could only put file-level directives right after the shebang), but what're you gonna do.

There is a workaround for this: to add a no-op command like `:` or `true` between file-level directives and the directive for the first line[^2][^3]. I don't really want to do this automatically as it might confuse the user.

In any case, I'd rather err on the side of caution. My main concern is that we don't want the user to unintentionally add a disable directive at the file-level when they intended to disable a line; if it goes unnoticed it could come back to bite them.

Given this, I think the best solution is to not offer a `Disable for Line` action for the first non-comment line at all; only `Disable for File` should be offered.

[^1]: https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-all-instances-in-a-file-044
[^2]: https://github.com/koalaman/shellcheck/issues/1770
[^3]: https://github.com/koalaman/shellcheck/issues/1877

b0o:
> Also, I think it might be a good idea to describe exactly what the code actions do, since they're synthetic

Good thinking, totally agree.

b0o:
> Given this, I think the best solution is to not offer a Disable for Line action for the first non-comment line at all; only Disable for File should be offered.

This is done in 155c587

b0o:
> Also, I think it might be a good idea to describe exactly what the code actions do, since they're synthetic

Done in 88a714b

jose-elias-alvarez:
The things linters do will never cease to amaze me, but I think your solution makes sense. Thanks for working on this!

PR #386, opened 2021-11-29
feature(yapf): allow range_format

meijieru:
Enable range formatting for yapf & black

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1374,6 +1374,8 @@ local sources = { null_ls.builtins.formatting.uncrustify }
 ##### About
 
 Formatter for `python` files
+- Supports both `textDocument/formatting` and `textDocument/rangeFormatting`.
+  - `textDocument/rangeFormatting` is line-based.
 
 ##### Usage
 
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Are you sure this works? I think this will make range formatting format the entire document, since the arguments aren't getting modified. I recommend taking a look at the [implementation for Prettier](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/prettier.lua) and see if we can do something similar for these formatters.

meijieru:
I hacked it and use it for several days.
Just find that they always format the entire file.
Sorry for that, and I will look at it later this week.

meijieru:
I fix the implementation for `yapf` and update the documentation.

There are some limitations though:
1. It is line-based.
2. Even range is given, it may change the out-of-range content like whitespace, which comes from the plugin itself.

meijieru:
For `black` we should wait for https://github.com/psf/black/issues/830

jose-elias-alvarez:
LGTM, thanks!

PR #387, opened 2021-11-30
feat: add pylama

klen:
[Pylama](https://github.com/klen/pylama) is python code audit tool. It's like `flake8` but has some unique features. The integration supports pylama v8.1.4+


jose-elias-alvarez:
Looks good, thank you!

PR #388, opened 2021-11-30
refactor!: remove 0.5.1 support

jose-elias-alvarez:
This PR removes support for Neovim 0.5.1, which simplifies the code base and enables new 0.6-exclusive features (e.g. #376). The goal is to merge this in a couple of weeks once the new version is more widespread and keep a branch pinned at the last commit before merging (with the caveat that I will no longer write or merge fixes for 0.5.1).


jose-elias-alvarez:
It seems like 0.6 availability is pretty decent already, so I'll plan on merging this at the end of the week.

Issue #383, opened 2021-11-27
Better sources discoverability


kylo252:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

It was possible to get a list of available sources even when they weren't registered, but somewhere between #332 and #321 that ability got removed. It may have never been intended to work that way from the beginning, but it was useful nevertheless. 

Before: 
![image](https://user-images.githubusercontent.com/59826753/143680979-6ebce372-de93-4a80-8aef-de913a2b3503.png)

That list is now always empty for non-registered sources, which isn't really that helpful.

### Help

Yes

### Implementation help

There are two options available:
- expose the config so that one can require it without setting up the source, see https://github.com/neovim/nvim-lspconfig/pull/1479
- create a static table that contains metadata about the available sources, see https://github.com/williamboman/nvim-lsp-installer/pull/159

jose-elias-alvarez:
The specific commit that removed this was d78734baeeedae8e73c76680726dfae57ed82d09. You can require a built-in without setting it up, actually (it’s only set up once you register it) but the only way I can think of to get a list of all available buiIt-ins would be to scan the file system. 

If it’s useful, I wouldn’t be opposed to a solution like the one you put together for lsp-installer.

kylo252:
Implemented in #397 🚀

PR #389, opened 2021-12-01
refactor!: remove lspconfig integration

jose-elias-alvarez:
Back in June, #18 added support for configuring null-ls via nvim-lspconfig. The hope was to eventually add a configuration for null-ls into lspconfig core, but that [seems unlikely](https://github.com/neovim/nvim-lspconfig/pull/1487). 

As mentioned in #18, configuring null-ls via lspconfig has advantages, but the dependency also opens us up to breakage caused by upstream changes, e.g. #378, as well as issues caused by misconfiguration (such as the mysteriously common practice of setting `autostart = true`). Removing the integration would also make configuration (slightly) less verbose and more clearly signal what does and doesn't work for null-ls (compared to an actual LSP server).

Right now, this PR removes the dependency on lspconfig to start the server and attach buffers. Apart from that, we still depend on lspconfig path utils (and the `:NullLsInfo` window creation mechanism), so we need to create our own version of those, and we need to update the documentation and fix a bunch of broken unit tests (integration tests are passing). 

If we're going to merge this, I also need to think about how to best signal to users that the integration has been removed, since it's a breaking change. Packaging this with #388 may be a good option.  

- [x] Replace path utils
- [x] Replace window creation mechanism for `:NullLsInfo`
- [x] Fix tests
- [x] Update documentation
- [x] Signal integration removal

Edit: This PR now also includes a performance boost to JSON-based sources thanks to `vim.json.decode` as well as a fix for #401 / better handling of dynamic source registration in general. Hopefully that helps offset the annoyance of requiring every user to update their config. 

Edit edit: While working on different branches, I noticed that the message about `config` deprecation is easy to miss when combined with the giant `lspconfig` error trace. I went back and forth on this, but I've refactored the code to instead show a deprecation message. After updating, users will see the following messages:

```
[null-ls] [WARN  12:35:53] ...site/pack/packer/start/null-ls.nvim/lua/null-ls/init.lua:60: config is deprecated; use setup instead
[null-ls] [WARN  12:35:53] ...site/pack/packer/start/null-ls.nvim/lua/null-ls/init.lua:39: lspconfig integration is deprecated; pass options to setup instead
```

I also added a workaround to try to maintain compatibility with `on_attach` (though I'm not confident it'll work perfectly). Obviously, the goal is to encourage users to move to the new configuration style ASAP, but I hope this will make the transition a little smoother.


jose-elias-alvarez:
A potential obstacle is that the lspconfig path utils are very good (certainly better than whatever I can come up with on my own, and far more battle-tested). I guess the easiest solution would be to simply include the relevant utils + license here, which would also protect us against future changes.

jose-elias-alvarez:
This is should be good to go. I want think about whether merging this makes sense, and if so I'll wait until #388 and #376 are merged to see if there's any issues from those PRs.

numToStr:
> I also need to think about how to best signal to users that the integration has been removed

You already have an *Breaking Change* issue. I think that would be enough.

Secondly, you should create a version tag before merging this so that the user could easily rollback if they can't update their config.

mjlbach:
FWIW I am in favor of this change. Lspconfig's path utils can be copied with the associated Apache 2.0 license, and eventually removed since they (at least, a version of them) will be added in core.

jose-elias-alvarez:
> You already have an _Breaking Change_ issue. I think that would be enough.
> 
> Secondly, you should create a version tag before merging this so that the user could easily rollback if they can't update their config.

Yep, you're right. I always feel weird about breaking changes, but that's the best I can do.

@mjlbach Thank you for the feedback, do you think adding [this paragraph](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/d7596a17859f52248984c5a12cbb28e43ea35d21/LICENSE#L26-L29) would be enough to handle the license issue? (Side note: Working on this refactor made me realize how impressive lspconfig is as a solution to the problem of "how do I seamlessly handle every language server + whatever weird things the server does + whatever weird things the user wants to do".)

kylo252:
I'm also in favor of this change, and I've  been wondering why it was still here.

As for `NullLsInfo`, let me know if you'd like to see something to what's implemented in LunarVim, see the screenshots from #383.  Alternatively, we can also use the simpler version that's used in `LspInfo`.

jose-elias-alvarez:
@kylo252 The LunarVim window is very nice and would be a solid improvement. For now, this PR just includes (most of) the code from the `:LspInfo` window.

jose-elias-alvarez:
This was a big refactor, but I've been using it all week and haven't run into any issues. Now that #376 and #388 are merged, I'm going to wait another week to let any issues from those PRs shake out, then merge this one.

sQVe:
@jose-elias-alvarez Cheers for all the amazing work you put into this plugin! 🎉 

I've previously been using `root_dir` when setting up `null-ls`, which now isn't available as an option anymore. I'm curious how the root now is resolved and if it's possible to configure it?

jose-elias-alvarez:
@sQVe Apologies, looks like gg`root_dir` fell through the cracks. 474a859f24cd6d8c7ef6c5942039d7dd201a2d09 should add the same functionality that was available via lspconfig.

kylo252:
@jose-elias-alvarez, does it make sense to expose other common settings, like `on_init` and `capabilities`? or maybe issue some warning that they're ignored, since `lspconfig` used to accept all of those.

jose-elias-alvarez:
@kylo252  `capabilities` shouldn't be exposed in the first place, I think, since it should depend on the sources a user has registered, but we can re-add `on_init`.

PR #391, opened 2021-12-01
fix(teal): avoid duplicate registration

meijieru:
Previously, the command is `tl` while the filename is `teal.lua`.
So it results in duplicate registration.
This PR fixes it by setting its name to `teal`

### DIFF
--- a/lua/null-ls/builtins/diagnostics/teal.lua
+++ b/lua/null-ls/builtins/diagnostics/teal.lua
@@ -4,6 +4,7 @@ local methods = require("null-ls.methods")
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
 return h.make_builtin({
+    name = "teal",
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
I don't mind this change, but what do you mean by duplicate registration?

meijieru:
For example, when I use this linter within LunarVim, I have the following issue, and every error will occur multiple time if I open multiple teal files.
I lazy load the linter by register the linter in `after/ftplugin/tl.lua`. This comes from LunarVim (seems the same as for null-ls itself from the doc) use the name to check whether it has been registered. Thus the mismatch cause duplicate registration.
image

jose-elias-alvarez:
I see, so LunarVim is already using the source API to check. In that case, this change makes sense, and we should probably audit other sources to make sure the file name and source name match, too.

Issue #148, opened 2021-09-03
Using GCC as a source of diagnostics?


IndianBoy42:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

I think that GCC  would be nice to use gcc as a source of diagnostics. Some of my c++ projects (particularly embedded projects) aren't that compatible with clang, and thus clangd sometimes doesn't work super well. Just using the GCC error messages would be great. 

### Help

Yes

### Implementation help

The main hard part is probably parsing the output of GCC, I guess we would have to use `-fdiagnostics-format=json`. Also we should look for `compile_commands.json` or `compile_flags.txt`, or maybe just run `cmake --build` or `make/ninja` and get the output

jose-elias-alvarez:
If it can output to JSON, it should be pretty straightforward to parse the output. I'm not at all familiar with the C++ ecosystem, so I can't do this on my own, but I can review / provide help with a PR for anyone who's interested in putting it together.

jose-elias-alvarez:
I believe [this file](https://github.com/dense-analysis/ale/blob/f0887d3e6178482255f11aa378124aef3699245f/ale_linters/cpp/cc.vim) and [this file](https://github.com/dense-analysis/ale/blob/f0887d3e6178482255f11aa378124aef3699245f/autoload/ale/handlers/gcc.vim) contain what ALE uses to spawn and parse GCC for C++ files (though like I said, I'm not familiar with the ecosystem, so I could be wrong). Not sure why they're not using the JSON option, but it could be a good place to start.

IndianBoy42:
I am starting on this work, I have one question, is there a builtin method for searching the working directory for things like `compile_commands.json` or cmake build directories?

jose-elias-alvarez:
nvim-lspconfig has [a bunch of useful utils](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/util.lua) for handling paths, so I imagine there's something there we can use (we have it as a dependency here, so we're free to use them unconditionally).

IndianBoy42:
So implementing this from scratch myself was quite hard, mainly to handle the project stuff, but recently I found [gccdiags](https://gitlab.com/andrejr/gccdiag) that can handle that for me. I just have to parse the output, they have some examples for `efm` and `diagnostic-languageserver`, I wonder if there is a straightforward translation from that?

[...]

PR #393, opened 2021-12-02
Add a builtin for using `gccdiag`

IndianBoy42:
Get diagnostics from gcc with the correct compiler arguments

Requires a `compiler_commands.json` file at the root of the working directory.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1374,6 +1374,7 @@ local sources = { null_ls.builtins.formatting.uncrustify }
 ##### About
 
 Formatter for `python` files
+
 - Supports both `textDocument/formatting` and `textDocument/rangeFormatting`.
   - `textDocument/rangeFormatting` is line-based.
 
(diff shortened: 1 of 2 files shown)


IndianBoy42:
Addresses #148

jose-elias-alvarez:
This looks good to me, do you think it's ready to merge? You mentioned in #148 that you wanted to get arguments from a JSON file. I don't know about the complexity involved, but it should theoretically be possible by setting `args` to a function that parses the file and returns the arguments (ideally caching them somehow to prevent parsing on each run).

IndianBoy42:
So now gccdiag handles parsing the compile_commands.json (usually generated by cmake) and invokes the compiler with the correct arguments. We don't have to do that ourselves

Oh you know what I forgot to link it properly in the PR description. 

https://gitlab.com/andrejr/gccdiag

I'm not invoking GCC directly as you see, gccdiag is basically a wrapper that looks through compile_commands and invokes gcc with the correct arguments

It's not perfect but very useful

jose-elias-alvarez:
Okay, got it. Do you think it might be a good idea to add a description of gccdiag to the documentation, too, in case users are unfamiliar with it?

IndianBoy42:
Yeah that makes sense, I'll do that, in a few hours

jose-elias-alvarez:
Great, thanks!

PR #396, opened 2021-12-03
feat: add :checkhealth support (WIP)

### LINKED ISSUE #392: Add `:checkhealth` support for checking if configured sources are executable
### Issues

- [x] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

I usually use `:checkhealth` to check if I all lspconfig executables `cmd` are executable.

_e.g._

```markdown
lspconfig: require("lspconfig.health").check()
========================================================================
  - INFO: clangd: configuration checked.
  - ERROR: arduino_language_server: The given command "arduino-language-server" is not executable.
  - INFO: cmake: configuration checked.
  - INFO: pyright: configuration checked.
  - ERROR: r_language_server: The given command "R" is not executable.
  - INFO: null-ls: configuration checked.
```

It would be helpful if I could do the same with null-ls, so that I could check if I have installed every executable that the plugin is configured to use.

This is useful if I am setting a new environment, or if I am not sure I have everything properly installed.

### Help

### Implementation help

I would like to help, but I don't have any experience implementing similar functionality.

**Edit:** `:help health-dev-lua` has the essentials of how to implement this functionality. I would still need help with how to get the executables for each configured source and how to check if the executables are executables.

datwaft:
This is a basic implementation of the `:checkhealth` functionality.

Closes #392.

I used the following `Dockerfile` and `init.lua` to test the functionality. Inside the Docker container I just opened `nvim` and executed `:checkhealth null-ls`.

<details>
<summary><code>Dockerfile</code></summary>

```Dockerfile
FROM debian:buster-slim
FROM rust:latest

# Install dependencies
RUN apt-get update && apt-get install -y curl git nodejs npm python3 python3-pip
RUN cargo install stylua
RUN pip3 install black
RUN npm i -g prettier @fsouza/prettierd markdownlint-cli

# Install Neovim
WORKDIR /tmp
RUN curl -LO https://github.com/neovim/neovim/releases/download/v0.6.0/nvim-linux64.tar.gz
RUN tar xzf nvim-linux64.tar.gz -C /opt
ENV PATH /opt/nvim-linux64/bin:$PATH

# Install plugins
WORKDIR /root/.local/share/nvim/site/pack/packer/start
RUN git clone https://github.com/nvim-lua/plenary.nvim
RUN git clone https://github.com/neovim/nvim-lspconfig

# Deploy plugin
COPY . /root/.local/share/nvim/site/pack/packer/start/null-ls.nvim/
COPY test-init.lua /root/.config/nvim/init.lua

# Finish
WORKDIR /root
```

</details>

<details>
<summary><code>init.lua</code></summary>

```lua
require("null-ls").config({
    sources = {
        require("null-ls").builtins.formatting.stylua,
        require("null-ls").builtins.formatting.prettierd,
        require("null-ls").builtins.formatting.markdownlint,
        require("null-ls").builtins.formatting.black,
        require("null-ls").builtins.formatting.sqlformat,
        require("null-ls").builtins.completion.spell,
        require("null-ls").builtins.code_actions.gitsigns,
    },
})

require("lspconfig")["null-ls"].setup({})
```

</details>

## TODO list

- [x] Add `:checkhealth` command support.
- [x] Add support for when command is an string.
- [x] Add support for when command is a function.
- [X] Add support for when command is nil.
- [x] Add support for `prefer_local` and `only_local`.
- [ ] add the flag to check if it should check executable.

I don't know how to support `prefer_local` or `only_local`, I would need guidance with that.

**Note:** do you have any builtin that uses the command as a function?, haven't found any to test it yet.

### DIFF
--- a/lua/null-ls/health.lua
+++ b/lua/null-ls/health.lua
@@ -13,8 +13,8 @@ local messages = {
 
 local function report(source)
     local name = source.name
-    local only_local = source.generator.opts.only_local
-    local command = source.generator.opts.command
+    local only_local = source.generator.opts and source.generator.opts.only_local
+    local command = source.generator.opts and source.generator.opts.command
 
     if type(command) ~= "string" then
         health.report_info(string.format(messages["unable"], name, command))
--- a/lua/null-ls/health.lua
+++ b/lua/null-ls/health.lua
@@ -1,0 +1,43 @@
+local health = require("health")
+local sources = require("null-ls.sources")
+local utils = require("null-ls.utils")
+
+local M = {}
+
+local messages = {
+    ["executable"] = [[%s: the command "%s" is executable.]],
+    ["not-executable"] = [[%s: the command "%s" is not executable.]],
+    ["unable"] = [[%s: cannot verify if the command is an executable.]],
+    ["executable-local"] = [[%s: the command "%s" is not globally executable, but it may be available locally.]],
+}
+
+local function report(source)
+    local name = source.name
+    local only_local = source.generator.opts.only_local
+    local command = source.generator.opts.command
+
+    if type(command) ~= "string" then
+        health.report_info(string.format(messages["unable"], name, command))
+        return
+    end
+
+    if utils.is_executable(command) then
+        health.report_ok(string.format(messages["executable"], name, command))
+        return
+    end
+
+    if only_local then
+        health.report_warn(string.format(messages["executable-local"], name, command))
+        return
+    end
+
+    health.report_error(string.format(messages["not-executable"], name, command))
+end
+
+M.check = function()
+    for _, source in ipairs(sources.get({})) do
+        report(source)
+    end
+end
+
+return M


lua/null-ls/health.lua
```diff
@@ -0,0 +1,29 @@
...
+    health.report_start("null-ls")
+    for _, source in ipairs(sources.get("")) do
+        local name = source.name
+        local command = source.generator.opts.command
+
+        if type(command) == "string" then
+            if utils.is_executable(command) then
+                health.report_ok(name .. ': the command "' .. command .. '" is executable.')
+            else
+                health.report_error(name .. ': the command "' .. command .. '" is not executable.')
```
datwaft, line 17: One enhancement for this could be to use `string.format` instead of raw string concatenation, would be more readable.

What do you think?

datwaft: Maybe using a dictionary to get each message, therefore each message is more easily updated.

jose-elias-alvarez: Yes, I think `string.format` would be a lot cleaner here.

lua/null-ls/health.lua
```diff
@@ -0,0 +1,29 @@
+local health = require("health")
+local sources = require("null-ls.sources")
+local utils = require("null-ls.utils")
+
+local M = {}
+
+M.check = function()
+    health.report_start("null-ls")
+    for _, source in ipairs(sources.get("")) do
```
jose-elias-alvarez, line 9:
```suggestion
    for _, source in ipairs(sources.get({})) do
```

jose-elias-alvarez: I think this makes it clearer that it's an empty query.

jose-elias-alvarez (commented):
I haven't tested this out yet but this seems to be on the right track. One thing I want to clarify is that generators with a command are a special case – there's no requirement to have a command at all, and it would be possible for a user / integration to create an implementation that defines a command under something other than `source.generator.opts.command`, which is just a convention used by [`generator_factory`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/73420db2b58408a60a19024be3c14b0eba8413fe/lua/null-ls/helpers.lua#L103)).

We can't reliably check other cases, so I wonder if it makes sense to only check generators created with `generator_factory`, which covers built-ins as well as custom generators created using [this method](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/73420db2b58408a60a19024be3c14b0eba8413fe/lua/null-ls/helpers.lua#L103). We could do this by adding a flag to those generators (something like `_null_ls_check_command`) and checking that here.

For function commands: It's supported as a user override but not included out-of-the-box. I think we should just ignore cases where `command` is not a string and not report anything. I do think we should make an exception if `only_local = true` and report a warning, something like `$COMMAND is not globally executable (may be available locally)`.

datwaft:
> We could do this by adding a flag to those generators (something like _null_ls_check_command) and checking that here.

That could be interesting, I will think about how to implement this.

> I think we should just ignore cases where command is not a string and not report anything.

My thought process with this is that I would feel confused if I added a source to the list but the source didn't appear on the `:checkhealth` result.

> I do think we should make an exception if only_local = true and report a warning, something like $COMMAND is not globally executable (may be available locally).

Thanks, I will add that to the implementation.

jose-elias-alvarez:
> My thought process with this is that I would feel confused if I added a source to the list but the source didn't appear on the :checkhealth result.

Sure, this makes sense. By the same logic as `only_local`, I guess we can report non-string commands with a warning that the health check was unable to verify whether they're executable.

datwaft:
Just added support for `only_local` and `prefer_local`.

I am not completely happy with the messages yet.

What do you think?

**TODO**: add the flag to check if it should check executable:

> We can't reliably check other cases, so I wonder if it makes sense to only check generators created with generator_factory, which covers built-ins as well as custom generators created using this method. We could do this by adding a flag to those generators (something like _null_ls_check_command) and checking that here.

datwaft:
BTW, on merge will you do the squashing or should I squash my commits before the merge?

lua/null-ls/health.lua
```diff
@@ -0,0 +1,54 @@
+local health = require("health")
+local sources = require("null-ls.sources")
+local utils = require("null-ls.utils")
+
+local M = {}
+
+local messages = {
```
jose-elias-alvarez, line 7: I like the table idea, but I think we could simplify the messages. In my opinion, there's only 3 cases we need to handle:

1. `command` is a string, and we can determine whether it's globally executable
2. `command` is not a string, so we don't know whether it's executable or not
3. `command` is a string and is not globally executable, but `only_local` is true, so this may be okay

lua/null-ls/health.lua
```diff
@@ -0,0 +1,54 @@
...
+    ["only-local"] = '%s: cannot verify if the command "%s" is a local executable.',
+    ["executable-but-prefer-local"] = '%s: the command "%s" is a global executable, but it\'s configured to prefer local executable.',
+    ["not-executable-but-prefer-local"] = '%s: the command "%s" is not a global executable, but it may be available locally.',
+    ["unknown"] = "%s: the command has an unknown type.",
+}
+
+M.check = function()
+    for _, source in ipairs(sources.get({})) do
+        local name = source.name
+        local prefer_local = source.generator.opts.prefer_local == true
```
jose-elias-alvarez, line 22: Can't we just check if this is truthy?

jose-elias-alvarez (commented):
Looks good, I think we should just simplify the messages to handle the relevant cases more cleanly and decide whether we will explicitly limit this to generators created using `generator_factory`. It would be great if you could squash the commits before we merge, but I'm not too fussy about it either way.

lua/null-ls/health.lua
datwaft: What do you think about this:

```lua
local messages = {
    ["executable"] = [[%s: the command "%s" is executable.]],
    ["not-executable"] = [[%s: the command "%s" is not executable.]],
    ["unable"] = [[%s: cannot verify if the command is an executable.]],
    ["executable-local"] = [[%s: the command "%s" is not a global executable, but it's configured to use a local executable.]],
}
```

jose-elias-alvarez: Yep, this looks good.

datwaft:
What do you think about the messages now?

I am not a 100% sure about the message logic, mainly about the `prefer_local` and `only_local`.

lua/null-ls/health.lua
```diff
@@ -0,0 +1,41 @@
...
+local sources = require("null-ls.sources")
+local utils = require("null-ls.utils")
+
+local M = {}
+
+local messages = {
+    ["executable"] = [[%s: the command "%s" is executable.]],
+    ["not-executable"] = [[%s: the command "%s" is not executable.]],
+    ["unable"] = [[%s: cannot verify if the command is an executable.]],
+    ["executable-local"] = [[%s: the command "%s" is not a global executable, but it's configured to use a local executable.]],
```
jose-elias-alvarez, line 11:
```suggestion
    ["executable-local"] = [[%s: the command "%s" is not globally executable, but it may be available locally.]],
```

lua/null-ls/health.lua
```diff
@@ -0,0 +1,41 @@
...
+}
+
+M.check = function()
+    for _, source in ipairs(sources.get({})) do
+        local name = source.name
+        local prefer_local = source.generator.opts.prefer_local
+        local only_local = source.generator.opts.only_local
+        local command = source.generator.opts.command
+
+        if type(command) == "string" then
```
jose-elias-alvarez, line 21: We could simplify this a bunch by using early returns, e.g.:

```lua
if type(command) ~= string then
    health.report_info(string.format(messages["unable"], name, command))
    return
end

local is_executable = utils.is_executable(command)
if is_executable then
    health.report_ok(string.format(messages["executable"], name, command))
    return
end

if only_local then
    health.report_warn(string.format(messages["executable-local"], name, command))
    return
end

health.report_error(string.format(messages["not-executable"], name, command))
```

jose-elias-alvarez: In case it wasn't clear: using early returns would require moving the logic into a separate function that's called from `M.check` (which is a reasonable idea anyways).

datwaft:
After some technical difficulties, I made the changes that you requested, what do you think?

lua/null-ls/health.lua
```diff
@@ -0,0 +1,43 @@
...
+    ["executable"] = [[%s: the command "%s" is executable.]],
+    ["not-executable"] = [[%s: the command "%s" is not executable.]],
+    ["unable"] = [[%s: cannot verify if the command is an executable.]],
+    ["executable-local"] = [[%s: the command "%s" is not globally executable, but it may be available locally.]],
+}
+
+local function report(source)
+    local name = source.name
+    local only_local = source.generator.opts.only_local
+    local command = source.generator.opts.command
```
jose-elias-alvarez, line 17:
```suggestion
    local only_local = source.generator.opts and source.generator.opts.only_local
    local command = source.generator.opts and source.generator.opts.command
```

Since `opts` is just a convention used by built-ins, we need to guard against the possibility that `opts` is nil here.

jose-elias-alvarez:
Other than the requested change this looks good to me! We'll have to see if we run into any issues with custom sources, but we can solve that as we go.

datwaft:
Just merged your changed, everything looks ok.

image

jose-elias-alvarez:
Great, thanks! I'm curious to see how this works with custom sources, but as I said we can improve it as we go. Thanks for working on this!

PR #397, opened 2021-12-03
feat: auto-generated filetypes metadata

### LINKED ISSUE #383: Better sources discoverability

kylo252:
Auto-generate metadata tables for filetype mappings.

Fixes #383


jose-elias-alvarez:
I'll have a chance to try this out next week, but as far as I can tell, the `diagnostics` and `formatters` tables are not generated automatically, right? That would be a downside, since it means we have to keep each source, its documentation, and the relevant table in sync. Ideally we could generate those automatically by scanning the filesystem.

kylo252:
> I'll have a chance to try this out next week, but as far as I can tell, the `diagnostics` and `formatters` tables are not generated automatically, right? That would be a downside, since it means we have to keep each source, its documentation, and the relevant table in sync. Ideally we could generate those automatically by scanning the filesystem.

I did auto-generate them, but it was a bit fiddly since I did it with `ripgrep` and then a vim-macro. any ideas how to grab them otherwise? It's not like I can source the file, either.

jose-elias-alvarez:
If we're going to use a headless Neovim instance anyways, we could modify the script to use [`vim.loop.fs_scandir`](https://github.com/luvit/luv/blob/master/docs.md#uvfs_scandirpath-callback) to get each path in the `builtins` directory, then scan each of those directories for built-ins, load them, and write their info to a file. 

I'd be hesitant to add this in unless we can automate it as much as possible (meaning the script gets all the necessary information on its own and we can run it via a Git hook or a CI action). We already have an issue with built-in documentation falling out of sync, so I've been thinking about automatically generating that, too.

kylo252:
> If we're going to use a headless Neovim instance anyways, we could modify the script to use [`vim.loop.fs_scandir`](https://github.com/luvit/luv/blob/master/docs.md?rgh-link-date=2021-12-05T18%3A11%3A19Z#uvfs_scandirpath-callback) to get each path in the `builtins` directory, then scan each of those directories for built-ins, load them, and write their info to a file.

Any ideas for how to source these files? Can I just set them all up in headless without opening any file and have access to their configs, or what did you have in mind?

kylo252:
> This is a rough sketch of what I had in mind. It's obviously brittle and we shouldn't use it as-is, but it should illustrate the point:

I didn't see your response in time, but I think my implementation is along the same lines 😄 

---

I should probably add the other sources as well even if the filetype attribute is empty, since you want them for docs generation. Any preference for wildcards for filetypes?
```lua
filetypes = {} 
-- vs 
filetypes = {"*"}
```

jose-elias-alvarez:
Let's go with `filetypes = {}` since that's the current style.

kylo252:
> Let's go with `filetypes = {}` since that's the current style.

Done! 

The new code might be a bit harder to read, but it's pretty generic and helps us avoid hard-coding the method names, just like you wanted :)

lua/null-ls/builtins/_meta/filetype_map.lua
```diff
@@ -0,0 +1,204 @@
...
+  html = {
+    formatting = { "rustywind" }
+  },
+  java = {
+    formatting = { "uncrustify" }
+  },
+  javascript = {
+    diagnostics = { "eslint_d" },
+    formatting = { "rustywind" }
+  },
```
kylo252, line 78: Something got messed up here. I need to make sure to not overwrite old values.

jose-elias-alvarez:
This looks great! Do you think there's a good way to run the script automatically? Since it depends on the Neovim runtime, I imagine setting up CI would be tricky, but maybe a Git hook?

kylo252:
> This looks great! Do you think there's a good way to run the script automatically? Since it depends on the Neovim runtime, I imagine setting up CI would be tricky, but maybe a Git hook?

I think the new workflow worked, but I'll try to give it a test just to make sure.

kylo252:
Oh man, I think I give up on the workflow action. I'm not sure if it's a problem with the authentication, or because the PR is not on the main branch. I'll disable the auto-commit part for now.

jose-elias-alvarez:
@kylo252 Now that this is merged, could we try to fix the workflow? Otherwise I can just remove it for now.

kylo252:
> Now that this is merged, could we try to fix the workflow? Otherwise I can just remove it for now.

Sure! and worst case scenario, you run it as a git-hook. I can recommend [pre-commit](https://github.com/pre-commit/pre-commit) since it's very easy to use. I could even set it up for you if you'd like.

PR #399, opened 2021-12-06
feat: enable project-level diagnostics for revive, staticcheck

xenrox:
Change to go linters as discussed.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2059,6 +2059,8 @@ Fast, configurable, extensible, flexible, and beautiful linter for Go.
 
 ##### Usage
 
+- Supports project-level diagnostics.
+
 ```lua
 local sources = { null_ls.builtins.diagnostics.revive }
 ```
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
Great, thank you for this and for your feedback on the PR!

fitrh:
Sorry to comment here, but did you guys know that adding `./...` to the end of the arguments makes `extra_args` for revive unusable? or is this the intention?

fitrh:
I'm sure not, revive uses it for project/package diagnostics, but we can't pass any other arguments after that, e.g. add `-c` to read configuration from current project, as in my workflow define `revive.toml` for each project.

@xenrox did you aware of this ?

Actually this is not a problem as we can just override the `args`, but again, `extra_args` will be unusable

xenrox:
Yes that is a problem for both revive and staticcheck that cannot easily be circumvented I think. As soon as you pass the first file (or in this case `./...`) to one of these linters, they no longer accept any configuration arguments.
@jose-elias-alvarez Do you think this particular behavior/quirk should be documented?

jose-elias-alvarez:
I see, I can't think of a better workaround for this case than to override `args` completely. A PR documenting this quirk would be very appreciated!

xenrox:
Have submitted a first version here: https://github.com/jose-elias-alvarez/null-ls.nvim/pull/888

PR #405, opened 2021-12-08
Update MAIN.md

rockorager:
I was following the MAIN.md doc to create a completion source, the doc shows a lowercase `completion` when all other methods are all caps...the lowercase variety was giving an error so I tried uppercasing it which worked.


jose-elias-alvarez:
Nice catch, thank you!

PR #406, opened 2021-12-09
feat: Allow diagnostics_format to be a function

### LINKED ISSUE #403: Allow specifying diagnostics format as function
### Issues

## Feature

Having a function that takes a `diagnostic` as input and returns a format string would allow more flexibility than the current `diagnostics_format` parameter provides. This could be done by allowing `diagnostics_format` to be a function instead of a string.

For example:

```lua
diagnostics_format = function(diagnostic)
  if diagnostic.code then
    return string.format("[%s] %s (%s)", diagnostic.code, diagnostic.message, diagnostic.source)
  end
  
  return string.format("%s (%s)", diagnostic.message, diagnostic.source)
end
```

This would also more closely match the pattern used by `vim.diagnostic.config` to pass `virtual_text.format`.

## Use Case

For me, the problem this solves is that some linters provide a `code` while others do not. I would like to have formats that differ based on whether or not that code exists. For example, for a linter `my-linter` with code `E123` I might want the following:

```
[E123] An error was detected (my-linter)
```

If the linter does not provide error codes, however, this would produce an unhelpful `[]` at the start of the message:

```
[] An error was detected (my-linter)
```

Whereas I would prefer to omit the brackets when the error code is not available.

```
An error was detected (my-linter)
```

Note that while the example provided was based on the example in `CONFIG.md`, the flexibility to come up with arbitrary patterns based on the presence/absence of an error is desirable:
```
[my-linter: E123] An error was detected
[my-linter] An error was detected
```

## Alternatives

1. This behavior can be achieved today by passing a custom `diagnostics_format` to each built-in diagnostic. However, this requires a configuration on a per-diagnostic basis, rather than being able to set it once globally. 
2. The message format could be extended to support more advanced string templating. [This is how ale tackles the same issue](https://github.com/dense-analysis/ale/blob/master/doc/ale.txt#L1027-L1051). That said, this would likely be more work on this repo for less flexibility to the users. 

### Help

Yes

### Implementation help

_No response_

rliebz:
Resolves #403

### DIFF
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -71,7 +71,7 @@ If you've installed an integration that provides its own sources and aren't
 interested in built-in sources, you don't have to define any sources here. The
 integration will register them independently.
 
-### diagnostics_format (string)
+### diagnostics_format (one of string, function)
 
 Sets the default format used for diagnostics. The plugin will replace the
 following special components with the relevant diagnostic information:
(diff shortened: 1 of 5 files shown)


kylo252 (changes requested):
This is unnecessary, as mentioned in [#403 ](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/403#issuecomment-989661191).

Instead, maybe add an example config or just a link to the Neovim docs.

PR #409, opened 2021-12-09
Add djhtml formatter

luisiacc:
Add [djhtml](https://github.com/rtts/djhtml) as formatting source.

I tried to run the tests but they broke in my face with some error that didn't found a module `plenary.test_harness`, checking out other PRs that add new sources I see they don't add any test so I guess this part of the code doesn't require tests at the time being, if they do let me know.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -443,6 +443,24 @@ local sources = { null_ls.builtins.formatting.dfmt }
 - `command = "dfmt"`
 - `args = {}`
 
+#### [djhtml](https://github.com/rtts/djhtml)
+
+##### About
+
+A pure-Python Django/Jinja template indenter without dependencies.
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.formatting.djhtml }
+```
+
+##### Defaults
+
+- `filetypes = { "django", "jinja.html", "htmldjango" }`
+- `command = "djhtml"`
+- `args = {}`
+
 #### [elm-format](https://github.com/avh4/elm-format)
 
 ##### About
--- a/lua/null-ls/builtins/formatting/djhtml.lua
+++ b/lua/null-ls/builtins/formatting/djhtml.lua
@@ -1,0 +1,14 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    method = FORMATTING,
+    filetypes = { "django", "jinja.html", "htmldjango" },
+    generator_opts = {
+        command = "djhtml",
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lua/null-ls/builtins/formatting/djhtml.lua
```diff
@@ -0,0 +1,16 @@
...
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    method = FORMATTING,
+    filetypes = { "djangohtml", "htmldjango", "jinja" },
+    generator_opts = {
+        command = "djhtml",
+        args = { "-i", "$FILENAME" },
```
kylo252, line 11:
```suggestion
        args = { "--quiet", "--in-place", "$FILENAME" },
```

doc/BUILTINS.md
```diff
@@ -443,6 +443,24 @@ local sources = { null_ls.builtins.formatting.dfmt }
...
+
+```lua
+local sources = { null_ls.builtins.formatting.djhtml }
+```
+
+##### Defaults
+
+- `filetypes = { "djangohtml", "htmldjango", "jinja" }`
+- `command = "djhtml"`
+- `args = { "-i" }`
```
kylo252, line 462:
```suggestion
- `args = { "--quiet", "--in-place", "$FILENAME" }`
```

lua/null-ls/builtins/formatting/djhtml.lua
```diff
@@ -0,0 +1,16 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    method = FORMATTING,
+    filetypes = { "djangohtml", "htmldjango", "jinja" },
```
kylo252, line 8: The default filetype name seems to be `htmldjango`. So I'm not sure about the rest.

kylo252 (changes requested):

lua/null-ls/builtins/formatting/djhtml.lua
luisiacc: I've seen them on some places, there is a jinja plugin that sets the filetype to jinja and have also seem similar behaviour with `djangohtml`, since those are uncommon I guess we can leave it to the user to add them if required.

jose-elias-alvarez:
I saw the discussion at #408 and tried this out locally. Are you sure `djhtml` doesn't accept `stdin`? I installed the latest version using `pip install djhtml` and the following seems to work for me on the example file you provided:

```lua

local FORMATTING = methods.internal.FORMATTING

return h.make_builtin({
    method = FORMATTING,
    filetypes = { "djangohtml", "htmldjango" },
    generator_opts = {
        command = "djhtml",
        args = { "$FILENAME" },
    },
})
```

luisiacc:
> I saw the discussion at #408 and tried this out locally. Are you sure `djhtml` doesn't accept `stdin`? I installed the latest version using `pip install djhtml` and the following seems to work for me on the example file you provided:
> 
> ```lua
> local h = require("null-ls.helpers")
> local methods = require("null-ls.methods")
> 
> local FORMATTING = methods.internal.FORMATTING
> 
> return h.make_builtin({
>     method = FORMATTING,
>     filetypes = { "djangohtml", "htmldjango" },
>     generator_opts = {
>         command = "djhtml",
>         args = { "$FILENAME" },
>     },
>     factory = h.formatter_factory,
> })
> ```
> 
> Edit: [this section](https://github.com/rtts/djhtml#usage) makes it pretty clear that `stdin` should work, so I wonder if there's a version mismatch issue going on here.

I just tried installing to the latest version and it seems to work fine with this config, this was the very first that I tried and it didn't work out, maybe I had an old version that didn't accepted stdin, seems like I also failed to understand this part of the readme `The default is to read from standard in and to write the indented output to standard out.`

jose-elias-alvarez:
Great, glad we figured it out. I'm happy to merge this PR once you've updated it to work with `stdin`.

For reference: the reason your original solution didn't work is that `djhtml $FILENAME` doesn't write changes to the temp file, so what happened is:

1. null-ls creates a temp file with the current file's content
2. `djhtml $FILENAME` reads from the temp file (but doesn't write to it, since the arguments didn't specify that)
3. null-ls reads the content from the unchanged temp file, and nothing happens since there are no changes

jose-elias-alvarez:
@syphar I think adding more filetypes is a good idea, but we don't currently do anything like what you're doing with `args` out-of-the-box (though of course we encourage it as a user customization). If you think it's something a lot of users would like to use, we could mention it in the documentation.

luisiacc:
I've added the other filetypes, I feel the default `tabwidth` (4) is good enough for me.

jose-elias-alvarez:
Great, thanks!

PR #410, opened 2021-12-09
Make shellharden overwrite rather than output to stdout

benjamineskola:
When I enable the shellharden builtin, I encounter a bug whereby any modifications I make to a file are reverted when saving the file.

I found that I could work around this by modifying the parameters given to shellharden when configuring null-ls; this change makes those the default.

### DIFF
--- a/lua/null-ls/builtins/formatting/shellharden.lua
+++ b/lua/null-ls/builtins/formatting/shellharden.lua
@@ -10,8 +10,8 @@ return h.make_builtin({
     },
     generator_opts = {
         command = "shellharden",
-        args = { "--transform", "$FILENAME" },
-        to_stdin = false,
+        args = { "--transform", "" },
+        to_stdin = true,
     },
     factory = h.formatter_factory,
 })


jose-elias-alvarez:
Hmm, from taking a look at the repo / `shellharden --help`, it seems like it should support `stdin` and `stdout` without using temp files. I can see that the current implementation is not using `stdin` properly, but [this issue](https://github.com/anordal/shellharden/issues/17) describes a potential solution. Could you try this diff and see if that solves your issue (and update the PR if it does)?

```diff
diff --git a/lua/null-ls/builtins/formatting/shellharden.lua b/lua/null-ls/builtins/formatting/shellharden.lua
index 2b58108..ca83fc7 100644
--- a/lua/null-ls/builtins/formatting/shellharden.lua
+++ b/lua/null-ls/builtins/formatting/shellharden.lua
@@ -10,8 +10,8 @@ return h.make_builtin({
     },
     generator_opts = {
         command = "shellharden",
-        args = { "--transform", "$FILENAME" },
-        to_stdin = false,
+        args = { "--transform", "" },
+        to_stdin = true
     },
     factory = h.formatter_factory,
 })
```

If you still have issues, I think your workaround is okay.

benjamineskola:
Yes, that appears to behave correctly. Thanks!

jose-elias-alvarez:
Great, could you just update the commit message to reflect the change?

jose-elias-alvarez:
Thanks!

Issue #364, opened 2021-11-19
Documentation on custom null-ls formatters is incomplete/confusing


andrewferrier:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.5.1

### null-ls config

N/A

### Steps to reproduce

N/A

### Expected behavior

I've been trying to figure out how to use null_ls to specify some custom formatters. I have to say I did find the documentation a little confusing. The home README shows an example of how to use a generator to implement a custom diagnostics source (https://github.com/jose-elias-alvarez/null-ls.nvim#parsing-cli-program-output), but didn't contain any reference to how to implement a formatter. It took me a while to find https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/MAIN.md (this doesn't seem to be linked from the main README?), which gave me some clues, but even then this was a bit confusing. The section on formatting (https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/MAIN.md#formatting) appears to be incorrect; it doesn't show that the object should be returned from the `on_output` function, and in fact it appears that has a signature of `function(params, done)` which isn't shown in the code. In the end I had to copy-n-paste an implementation of `on_output` out of the `helpers.lua` file.

All in all, I think some more examples of how to create custom sources for formatting and other types apart from diagnostics would be really welcome.

I would be interested in helping you with this and sending in some pull request(s) if you'd like. One of the things I do find a bit odd is that there seems to be a distinction in `null_ls` between built-in sources and other sources, with slightly different functions for configuring these. I still don't really quite grok why this is.

I know I did open a bug on this repo before and you said you were trying to encourage people to add built-in sources. I agree, and I think that's a good goal, but it's important to make it easy for them to configure their own locally first to try them out before they send them in to you. I have a few I could send you but I'm not sure if I'm doing it the right way as I have nothing to crib from :)

Great project, `null_ls` works really well, and I would look forward to helping you if you're interested.

### Actual behavior

N/A

### Debug log

N/A

### Help

Yes

### Implementation help

[...]

PR #413, opened 2021-12-10
refactor(helpers): simplify internal built-in structure

jose-elias-alvarez:
#251 allowed registering multiple copies of the same built-in with different options, but the code has grown complicated as we've continued to add new options. This refactor moves all of the "special" `with` code into `make_builtin` so that `with` can simply return a new copy created with `make_builtin`. As a result, built-in behavior should be a little more consistent, since there's no real reason why some methods / options should only be available when using `with`. 

The refactor also codifies which options are valid to pass into `with` instead of handling a subset of keys and treating all others as generator options. This might break some exotic uses of `with` to stretch built-ins beyond what they're supposed to do (for example, you can no longer modify spawn options like `to_stdin` or `to_temp_file`). As mentioned in #364, there's still confusion about the distinction between built-in sources (which are only intended for use in the null-ls repo) and generic custom sources, but I think implicitly clarifying what built-ins can do is a step in the right direction. 

Edit: This PR now also modularizes the exports in `helpers.lua`, which should make refactoring / reading the source code in general less painful. As part of this, I've gone ahead and removed the `conditional` helper (unrelated to `with({ condition = ... })`), which was pretty useless and is trivial to replicate.

- [x] Make sure valid options work and have test coverage
- [x] Document valid options
- [x] Fix `generator_factory` tests


kylo252:
Isn't it about time to graduate the factory module as first-class instead of being stuck as a "helper"? 😄 

If nothing else, it would help in navigating the code and make further refactoring efforts easier.

jose-elias-alvarez:
Yep, splitting off `generator_factory` (and possibly `make_builtin`) is probably a good idea, since I agree that navigating in the file is unpleasant.

jose-elias-alvarez:
This should be good to go. I'm pretty confident this will break some things users are doing with built-ins, but the main valid cases are covered for now, so I think the right thing to do is merge it and deal with these cases as they come up (either extending functionality or clarifying why it's not something we want to support).

PR #417, opened 2021-12-11
Display the expanded snippet in the documentation field.

oncomouse:
This builds on the previous vsnip completion implementation to preview expanded snippets.


lua/null-ls/builtins/completion/vsnip.lua
```diff
@@ -11,14 +11,27 @@ return h.make_builtin({
         fn = function(params, done)
             local items = {}
             local snips = vim.fn["vsnip#get_complete_items"](params.bufnr)
-            local targets = vim.tbl_filter(function(item)
-                return string.match(item.word, "^" .. params.word_to_complete)
-            end, snips)
-            for _, item in ipairs(targets) do
+            for _, item in ipairs(snips) do
+                local user_data = vim.fn.json_decode(item.user_data)
```
jose-elias-alvarez, line 15:
```suggestion
                local user_data = vim.json.decode(item.user_data)
```

#389 changes all uses of `vim.fn.json_decode` to `vim.json.decode`, which is faster and available on 0.6.0, so we might as well use it here, too.

jose-elias-alvarez:
This seems like a good change and works perfectly with nvim-cmp, but `vim.lsp.omnifunc` + `completeopt=preview` shows raw Markdown:

Screen Shot 2021-12-11 at 1 17 37 PM

I don't think it's is a big deal, but do you think there's anything we can do to handle this?

oncomouse:
@jose-elias-alvarez Let me see what I can dig up about that. I've had problems with that with other completion sources I've written, so I probably need to check some documentation to figure that out.

Also, I've continued to work on this and I have a more advanced version that returns proper LSP-style snippets that work with `vim-vsnip-integ` and expand at completion. Should I push that to this PR or open another one?

jose-elias-alvarez:
Sure, we can make this into a general vsnip source improvement PR.

PR #419, opened 2021-12-12
feat(completion): add luasnip as a source

saadparwaiz1:


gegoune:
Is that drop in replacement for `saadparwaiz1/cmp_luasnip`?

saadparwaiz1:
@gegoune more or less. The features currently missing are:

- clear cache on autocmds
- refresh cache on autocmds
- expand snippets on confirmation (currently you have to manually expand them after selecting)

I wasn't sure what's the best place to add the first 2 features so will wait on @jose-elias-alvarez. Plus not really sure how we can accomplish the third without some hack maybe @L3MON4D3 has an idea

lua/null-ls/builtins/completion/luasnip.lua
```diff
@@ -0,0 +1,74 @@
...
+                    snip_cache[ft] = ft_items
+                end
+                vim.list_extend(items, snip_cache[ft])
+            end
+            local line_to_cursor = require("luasnip.util.util").get_current_line_to_cursor()
+            done({
+                {
+                    items = vim.tbl_filter(function(item)
+                        return vim.startswith(item.word, params.word_to_complete)
+                            and (item.data.show_condition or item.data.show_condition(line_to_cursor))
```
L3MON4D3, line 66: Is the check here necessary? IIRC `show_condition` defaults to a function returning true, so it should be safe to call right away.

L3MON4D3:
About expanding snippets upon selection, no good idea apart from doing the same as in `cmp_luasnip` (accept-callback) :/
(maybe setting `__index` on some table that has a key that is only read when the item is accepted?)

saadparwaiz1:
does the `CompleteDone` event get triggered after confirming? Maybe we can hook into that and expand if the completed item is a luasnip item?

L3MON4D3:
Hmm, that could be one way, but I'm not sure if it would work with eg. `cmp`, I think they use a floating window to show completion-items, not the default pum.

L3MON4D3:
What could also work: instead of passing the actual text to be inserted in `TextEdit.newText`, we could pass an unique identifier for a given snippet and provide a function that resolves and expands the snippet belonging to the identifier as callbacks for completion engines (in `cmp`s case that would be [this function](https://github.com/hrsh7th/nvim-cmp#i-dont-use-a-snippet-plugin))

saadparwaiz1:
@L3MON4D3 I think we can do this as a follow up PR? because even vsnip right now needs to be manually expanded

jose-elias-alvarez:
> What could also work: instead of passing the actual text to be inserted in `TextEdit.newText`, we could pass an unique identifier for a given snippet and provide a function that resolves and expands the snippet belonging to the identifier as callbacks for completion engines (in `cmp`s case that would be [this function](https://github.com/hrsh7th/nvim-cmp#i-dont-use-a-snippet-plugin))

This seems like a sensible approach, though as @saadparwaiz1 noted we don't currently support expansion, so I agree that we should handle that in another PR. `CompleteDone` is weird because it's triggered whether the user confirms or not, and as @L3MON4D3 mentioned it won't work with plugins that don't use the native `pum`. 

> - clear cache on autocmds
> - refresh cache on autocmds

What are the conditions that should trigger a cache clear / refresh?

saadparwaiz1:
@jose-elias-alvarez thanks for having a look. LuaSnip emits two user autocmds which can be seen [here](https://github.com/saadparwaiz1/cmp_luasnip/blob/7bd2612533db6863381193df83f9934b373b21e1/after/plugin/cmp_luasnip.lua#L7-L8) the commands should either remove the cache completely, or "refresh" it i.e. just remove certain filetypes from the cache

jose-elias-alvarez:
I see. We haven't handled anything like this before, but one approach would be to set up the autocommands when the built-in is loaded (which happens when `null_ls.builtins.completion.luasnip` is indexed). Something like this (untested, but it should illustrate the idea):

```lua
_G._null_ls_luasnip_clear = function()
    -- clear cache
end
_G._null_ls_luasnip_refresh = function()
    -- refresh cache
end

vim.cmd([[
  augroup NullLsLuasnip
    au!
    autocmd User LuasnipCleanup lua _null_ls_luasnip_clear()
    autocmd User LuasnipSnippetsAdded lua _null_ls_luasnip_refresh()
  augroup END
]])
```

We could also avoid using the global table by storing the cache and callbacks in [state](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/state.lua), but that might add too much complexity.

saadparwaiz1:
I think the easiest option right now would be just remove the cache and just generate snippets everytime. Hopefully, the performance impact would be pretty low. What do you think @L3MON4D3?

L3MON4D3:
There should be little performance impact, I agree (only generating docstrings is somewhat costly, but they are cached inside the snippets).

jose-elias-alvarez:
Great, let's go for the simplest option for now then.

saadparwaiz1:
Perfect, will update the PR :)

jose-elias-alvarez:
Thanks! Let me know when this is OK to merge (I don't use Luasnip so it would take me a while to test this locally).

saadparwaiz1:
I have updated the PR and it works for me :)

jose-elias-alvarez:
Thank you!

dsully:
I'm struggling to get this to work. I've added luasnip as a null-ls built-in source, but it's unclear on how to wire it up to nvim-cmp's config. Any help would be appreciated.

jose-elias-alvarez:
@dsully If you are using nvim-cmp, there's really no reason to go through null-ls. You can just use [cmp_luasnip](https://github.com/saadparwaiz1/cmp_luasnip) directly and get a better experience.

dsully:
I see. That part was unclear from this PR / docs.

PR #424, opened 2021-12-13
feat: Send buffer type to cspell

Sloff:
In order to include language specific words as well

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cspell.lua
+++ b/lua/null-ls/builtins/diagnostics/cspell.lua
@@ -11,7 +11,7 @@ return h.make_builtin({
         args = function(params)
             return {
                 "--language-id",
-                vim.api.nvim_buf_get_option(params.bufnr, "filetype"),
+                params.ft,
                 "stdin",
             }
         end,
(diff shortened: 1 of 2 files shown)


Sloff:
For example in Rust, before this change:
image

After change:
image

lua/null-ls/builtins/diagnostics/cspell.lua
```diff
@@ -8,7 +8,13 @@ return h.make_builtin({
     filetypes = {},
     generator_opts = {
         command = "cspell",
-        args = { "stdin" },
+        args = function(params)
+            return {
+                "--language-id",
+                vim.api.nvim_buf_get_option(params.bufnr, "filetype"),
```
jose-elias-alvarez, line 14:
```suggestion
                params.ft,
```

A small optimization. Looks good to me otherwise!

Sloff: Awesome, thanks @jose-elias-alvarez!

jose-elias-alvarez:
Awesome, thanks!

PR #425, opened 2021-12-13
prettier: support solidity

ChristianChiarulli:


jose-elias-alvarez:
Does Prettier work out-of-the-box with Solidity, or does it require [the plugin](https://github.com/prettier-solidity/prettier-plugin-solidity)? So far we've been avoiding adding filetypes as defaults if they require an additional plugin (though some might have slipped through the cracks).

vuki656:
> Does Prettier work out-of-the-box with Solidity, or does it require [the plugin](https://github.com/prettier-solidity/prettier-plugin-solidity)? So far we've been avoiding adding filetypes as defaults if they require an additional plugin (though some might have slipped through the cracks).

Seems like it as noted [here](https://prettier.io/docs/en/plugins.html#community-plugins). Although it would be an interesting feature to support those and let the user know they need to install the plugin.

vuki656:
Just tested it and it works as expected if you add it to the `filetypes` and install the plugin globally. 

You just have to copy all the existing `filetypes` and then add `solidity` to it in the null-ls config which is a little ugly. Maybe add a way to add additional filetypes instead of overriding the default ones?

jose-elias-alvarez:
> Just tested it and it works as expected if you add it to the `filetypes` and install the plugin globally.
> 
> You just have to copy all the existing `filetypes` and then add `solidity` to it in the null-ls config which is a little ugly. Maybe add a way to add additional filetypes instead of overriding the default ones?

Yes, I agree that an `extra_filetypes` option would be good for cases like this where the source supports a large number of filetypes out-of-the-box. I'll open a separate issue to track that, but I think the defaults should only contain filetypes that work without additional plugins, so I'll close this PR.

PR #430, opened 2021-12-14
Add nil-value check to handle_single_file_diagnostics

mahmoudajawad:
This MR contains a commit that adds nil-value check to the method (line#73) as at some point, for a fraction of a second, `vim.diagnostic` is `nil` resulting in the whole sequence of whatever happening in `diagnostics` file failing. This is happening to me at every buffer open, as well as at saving with format. 

It is worth noting that I'm using nvim 0.5.1, and not 0.6.0 (current version in fedora repos), and I'm not sure if this commit would break anything with nvim 0.6.0, but I assume it should be harmless.

### DIFF
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -70,6 +70,9 @@ local postprocess = function(diagnostic, _, generator)
 end
 
 local handle_single_file_diagnostics = function(namespace, diagnostics, bufnr)
+    if vim.diagnostic == nil then
+        return
+    end
     vim.diagnostic.set(namespace, bufnr, diagnostics)
 end
 


jose-elias-alvarez:
This is a harmless change but at the same time, we only explicitly support 0.6.0+, so I don't know if it sets a good precedent to merge a change like this. What is your use case? Are you just not using diagnostics at all?

mahmoudajawad:
@jose-elias-alvarez, I'm not sure what is going on here, and whether diagnostics are running or not, but I'm getting all the warnings and errors I expect from eslint. I'm using following settings (mostly your blog post work):
```viml
require('lspconfig').tsserver.setup({     
    on_attach = function(client, bufnr)     
        client.resolved_capabilities.document_formatting = false     
        client.resolved_capabilities.document_range_formatting = false     
        local ts_utils = require("nvim-lsp-ts-utils")     
        ts_utils.setup({         
            eslint_bin = "eslint_d",     
            eslint_enable_diagnostics = true,     
            eslint_enable_code_actions = true,     
            enable_formatting = true,     
            formatter = "prettier",     
        })                    
        ts_utils.setup_client(client)     
        buf_map(bufnr, "n", "gs", ":TSLspOrganize<CR>")     
        buf_map(bufnr, "n", "gi", ":TSLspRenameFile<CR>")     
        buf_map(bufnr, "n", "go", ":TSLspImportAll<CR>")     
        on_attach(client, bufnr)     
    end,                      
})                               
                                 
require("null-ls").config({})     
require('lspconfig')["null-ls"].setup({ on_attach = on_attach })     
```

jose-elias-alvarez:
It looks like you are following old instructions. You can either update to 0.6.0 and follow the instructions in the README or keep your current 0.5.1 config and use your plugin manager to pin the version to the [0.5.1-compat](https://github.com/jose-elias-alvarez/null-ls.nvim/tree/0.5.1-compat) branch.

mahmoudajawad:
Very well. I'll shift to using the branch until fedora repos update to neovim 0.6. Thanks for your time.

PR #438, opened 2021-12-14
Pass STDERR to check_exit_code

levouh:
Obviously this is not super useful for things like diagnostics, but for something that doesn't run as often (like formatters) it is useful to be able to do something like:

```lua
local function check_exit_code(code, stderr)
  local success = code == 0
  if not success then
    vim.notify({ "Formatting failed", stderr })
  end
  
  return success
end
```

to know _why_ the operation failed. There are other things that could be passed here (e.g. `stdout`) but I only really see a use case for `stderr`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: test
 test:
-	nvim --headless -u test/minimal.vim -c "lua require(\"plenary.test_harness\").test_directory_command('test/spec {minimal_init = \"test/minimal.vim\"}')"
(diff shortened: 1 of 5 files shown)


levouh:
_CI_ is failing due to me (seemingly) needing to run `./scripts/autogen_metadata.sh`; let me know if I actually need to do that @jose-elias-alvarez.

jose-elias-alvarez:
No need to worry about CI for now, but could you split the change to the `Makefile` into a separate commit? Other than that, I like this idea and am happy to merge this.

levouh:
Good call, I've split those two out into separate commits.

jose-elias-alvarez:
Great, thanks!

PR #440, opened 2021-12-14
ci: fix the metadata-update workflow

kylo252:

PR #445, opened 2021-12-15
refactor(loop): move logger inside closure

kylo252:


kylo252:
I thought it's better to move this out of #440.

What do you think about the changes in log level that I made?

lua/null-ls/client.lua
```diff
@@ -105,7 +105,7 @@ end
 
 M.setup_buffer = function(bufnr)
     if not client then
-        log:debug(string.format("unable to set up buffer %d (client not active)", bufnr))
+        log:warn(string.format("unable to set up buffer %d (client not active)", bufnr))
```
jose-elias-alvarez, line 108: No client indicates that the client either failed to start up (which is logged elsewhere) or that the user shut it down (in which case they probably don't want to see this). Was there a specific situation that prompted this change?

lua/null-ls/client.lua
```diff
@@ -125,7 +125,7 @@ end
 
 M.notify_client = function(method, params)
     if not client then
-        log:debug(
+        log:warn(
```
jose-elias-alvarez, line 128: My logic for not doing this is the same as the `setup_buffer` log. Also I believe this situation can happen when the user directly opens a file from the command line.

lua/null-ls/helpers.lua
```diff
@@ -253,7 +253,7 @@ M.generator_factory = function(opts)
 
             -- if dynamic_command returns nil, don't fall back to command
             if not resolved_command then
-                log:debug(string.format("failed to resolve command [%s]; not spawning", command))
+                log:warn(string.format("failed to resolve command [%s]; not spawning", command))
```
jose-elias-alvarez, line 256: This is actually a valid situation when the user doesn't want to run a command if it's not available, such as the [`only_local` helper](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/10a1c36e894a6793e92731d9e8bb3e6baad14e27/lua/null-ls/helpers.lua#L439). Maybe we could change the log message to sound less dire.

lua/null-ls/client.lua
kylo252: Not really, it just looked like a warning to me.

lua/null-ls/helpers.lua
```diff
@@ -253,7 +253,7 @@ M.generator_factory = function(opts)
 
             -- if dynamic_command returns nil, don't fall back to command
             if not resolved_command then
-                log:debug(string.format("failed to resolve command [%s]; not spawning", command))
+                log:warn(string.format("unable to resolve command [%s]", command))
```
jose-elias-alvarez, line 256: I think my last comment wasn't clear, but I think this should still be `debug`, since it doesn't indicate an error.

Other than that this looks good. Does moving logging into the closure fix the error when running the metadata generation script?

kylo252: I didn't know it was causing an error, I just assumed you wanted these changes, so I moved them out to a separate PR 😅 

You can see that the original PR is working fine. This one is just complaining about an out-of-date _meta files.

jose-elias-alvarez:
Wouldn't it be better to avoid the dependency on Plenary in the first place? That was the goal of moving `require` into the closure. Also, wouldn't this change mean that we'd need to clone Plenary, too?

jose-elias-alvarez:
I used a different approach in f6ac566882db8f256f4202cbb1ef9c790d8ba7a3 to fix the logger issue and also included your log change, so I think this is outdated now.

jose-elias-alvarez:
@kylo252 This is pretty strange, but it seems that the CI runner and my local environment disagree on the order in which `prettierd` and `prettier_d_slim` appear in the filesystem. 42edfb3f4aca63039bbbd00033c0f46467ddaea9 is a commit I made after running the script locally, but f8da1595297ed2e5577743fb1e013ee1d0046262 immediately changed it back. Any idea why this might be happening? Could this be some quirk of how `vim.fn.glob` works?

PR #451, opened 2021-12-15
add(source): latexindent formatting source

kunzaatko:
latexindent formatting source for latex files. No issues are attached.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -817,6 +817,24 @@ local sources = { null_ls.builtins.formatting.lua_format }
 - `command = "lua-format"`
 - `args = { "-i" }`
 
+#### [latexindent](https://github.com/Koihik/LuaFormatter)
+
+##### About
+
+A `perl` script for formatting `LaTeX` files that is generally included in mayor `TeX` distributions.
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.formatting.latexindent }
+```
+
+##### Defaults
+
+- `filetypes = { "tex" }`
+- `command = "latexindent"`
+- `args = { "-" }`
+
 #### [markdownlint](https://github.com/igorshubovych/markdownlint-cli)
 
 ##### About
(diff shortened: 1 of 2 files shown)


doc/BUILTINS.md
```diff
@@ -817,6 +817,24 @@ local sources = { null_ls.builtins.formatting.lua_format }
...
+
+```lua
+local sources = { null_ls.builtins.formatting.latexindent }
+```
+
+##### Defaults
+
+- `filetypes = { "tex" }`
+- `command = "latexindent"`
+- `args = { }`
```
jose-elias-alvarez, line 836:
```suggestion
- `args = { "-" }`
```

lua/null-ls/builtins/formatting/latexindent.lua
```diff
@@ -0,0 +1,11 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    method = FORMATTING,
+    filetypes = { "tex" },
+    generator_opts = { command = "latexindent", args = { "-" }, format = "raw", to_stdin = true },
```
jose-elias-alvarez, line 9: Do you specifically want `format = "raw"` here? We generally only use that if we want to handle errors manually. Take a look at [the documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#format) – it may be better to leave this undefined.

kunzaatko:
You are right on both occasions! Thanks!

kunzaatko:
I formatted with a different stylua config. Sorry.

kunzaatko:
I could not figure out, how to squash the commits... Do you have a resource on how to do it, or could you please explain it briefly to me?

jose-elias-alvarez:
I would use `git rebase --interactive`, described [here](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History), to start from your first commit, squash the ones in the middle, then force push to this branch. You could also use something like [lazygit](https://github.com/jesseduffield/lazygit) or a GUI tool to achieve the same result. It's worth learning!

kunzaatko:
Great! Thanks! I read through it. I am still a bit rusty with git. I don't have so much experience with it since I do not have a software background but a mathematics one. But this will come in handy :)

jose-elias-alvarez:
LGTM, thank you for the contribution (and glad you learned something new)!

PR #452, opened 2021-12-16
fix: client formatting disable through resolved_capabilities

### LINKED ISSUE #436: Disable null-ls formatting from `on_attach`
### Issues

### Feature description

Condition: LSP server (e.g. gopls) has formatting capabilities but no formatting source has been specifically enabled for Go filetypes from null-ls, except for `null_ls.builtins.formatting.trim_whitespace` which has `filetypes = {}`. I want to be able to disable null's formatting capabilities, specifically for this source, when this condition is satisfied (if LSP supports formatting but null does not specifically):
```lua
function default_on_attach(client, bufnr)
  -- documentFormatting
  local name = client.name
  print(name)
  local is_null = name == "null-ls"
  local ls_can_format = client.resolved_capabilities.document_formatting
  print("ls_can_format", ls_can_format)
  local null_can_format = require("lsp.null").use_null_formatting()
  print("null_can_format", null_can_format)

  -- prefer null-ls for formatting if available
  if (not is_null and null_can_format) or (is_null and not null_can_format) then
    -- disable formatting
    print("formatting disabled")
    client.resolved_capabilities.document_formatting = false
    client.resolved_capabilities.document_range_formatting = false
  else
    -- use client for formatting
    print("formatting enabled")
    vim.cmd("autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync()")
  end
end
```
Please note that my language servers and null both use the same `on_attach` function above. It seems that disabling formatting for LSP servers doesn't work for null-ls in the same way. My full config can be found [here](https://github.com/axieax/dotconfig). 

### Help

### Implementation help

Not sure if `client.resolved_capabilities.document_formatting = false` and `client.resolved_capabilities.document_range_formatting = false` can be used to target null-ls.

axieax:
Resolves #436 as ease-of-use functionality

### DIFF
--- a/lua/null-ls/client.lua
+++ b/lua/null-ls/client.lua
@@ -27,6 +27,11 @@ end
 local on_init = function(new_client, initialize_result)
     -- null-ls broadcasts all capabilities on launch, so this lets us have finer control
     new_client.supports_method = function(method)
+        -- user tries to disable formatting the lsp way
+        if not new_client.resolved_capabilities.document_formatting then
+            return false
+        end
+
         local internal_method = methods.map[method]
         if internal_method then
             return require("null-ls.generators").can_run(vim.bo.filetype, internal_method)


jose-elias-alvarez:
This is a great change to make our behavior less surprising. Thanks you!

hassec:
Hi @jose-elias-alvarez, sorry for the ping, but I think I just stumbled over this....  

shouldn't setting `client.resolved_capabilities.document_formatting == false` only disable `supports_method` if that's the actual method we are checking?

jose-elias-alvarez:
@hassec You are correct – 0a8bf7b241646dfa7633f0749cead654e6b2fb9d should improve this behavior.

PR #461, opened 2021-12-17
fix: correctly read root_dir for prefer_local/local_only

sileht:
root_dir detection doesn't work with prefer_local/local_only is used

This fixes it.


jose-elias-alvarez:
Thanks for catching this!

PR #462, opened 2021-12-17
fix(builtins): mypy output pattern

sileht:
<string> doesn't work to catch filepath before the first `:`, uses
`[^:]+` instead.


jose-elias-alvarez:
Looks good, thank you!

PR #463, opened 2021-12-17
fix(builtins): mypy replace --command by --shadow-file

sileht:
Using --command $TEXT even if the text looks correctly escaped mypy
sometimes we pass multiple arguments to --command and fail with:

output: usage: mypy [-h] [-v] [-V] [more options; see below]
[-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]
mypy: error: May only specify one of: module/package, files, or command.

This change uses directly --shadow-file $ORIGINAL_FILENAME $FILENAME
$REAL_FILENAME instead of --command $TEXT to avoid this.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -8,17 +8,22 @@ return h.make_builtin({
     filetypes = { "python" },
     generator_opts = {
         command = "mypy",
-        args = {
-            "--hide-error-codes",
-            "--hide-error-context",
-            "--no-color-output",
-            "--show-column-numbers",
-            "--show-error-codes",
-            "--no-error-summary",
-            "--no-pretty",
-            "--command",
-            "$TEXT",
-        },
+        args = function(params)
+            return {
+                "--hide-error-codes",
+                "--hide-error-context",
+                "--no-color-output",
+                "--show-column-numbers",
+                "--show-error-codes",
+                "--no-error-summary",
+                "--no-pretty",
+                "--shadow-file",
+                params.bufname,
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
I can see how sending text content directly could cause issues, but with this change we'd be telling `mypy` to read the file from the disk, meaning that diagnostics won't reflect unsaved changes to the buffer. We could change it to only run on save, but that seems like a regression from current behavior.

The better option (and one that I've mentioned before in #155) seems to be to use `--shadow-file`. I'd prefer to try that before resorting to this.

sileht:
`--shadow-file` version looks better indeed.

lua/null-ls/builtins/diagnostics/mypy.lua
```diff
@@ -16,9 +16,12 @@ return h.make_builtin({
             "--show-error-codes",
             "--no-error-summary",
             "--no-pretty",
-            "--command",
-            "$TEXT",
+            "--shadow-file",
```
jose-elias-alvarez, line 19: Could we tweak this to use a function instead? I'm a little hesitant to add `REAL_FILENAME`, since it's a bit hard to understand and probably not too useful outside of this specific case.We'd just have to use something like this (untested):

```lua
args = function(params)
    return {
        "--hide-error-codes",
        "--hide-error-context",
        "--no-color-output",
        "--show-column-numbers",
        "--show-error-codes",
        "--no-error-summary",
        "--no-pretty",
        "--shadow-file",
        params.bufname,
        params.temp_path,
        params.bufname,
    }
end
```

jose-elias-alvarez:
Great, thanks!

PR #466, opened 2021-12-18
feat: add supported sources in the info panel

kylo252:
- feat: list both supported and active sources in NullLsInfo.
- refactor: organize the logger information
- feat: add _very_ basic highlighting
- feat: add a banner (pretty unnecessary, it's just for fun :smile:)


kylo252:
I tried to leave it pretty basic, just so you can tweak it as you'd like. Here's what it currently looks like

image

jose-elias-alvarez:
Looks great at a glance! Will test this in a few days. I see what you mean about the methods and will think about what we can do there.

lua/null-ls/info.lua
```diff
@@ -65,29 +78,82 @@ end
...
+        local name = methods.readable[method]
+        local active_sources = {}
+        local supported_sources = {}
+
+        for _, source in ipairs(sources.get_available(ft, method) or {}) do
+            table.insert(active_sources, source.name)
+        end
+
+        -- the metadata is indexed by the builtin-names
+        for _, source in ipairs(sources.get_supported(ft, name) or {}) do
```
jose-elias-alvarez, line 97:
```suggestion
        for _, source in ipairs(sources.get_supported(ft, method)) do
```

Wouldn't passing in `method` here let us keep the old readable names? This also will always return a table.

lua/null-ls/sources.lua
```diff
@@ -74,6 +74,15 @@ M.get_available = function(filetype, method)
     return available
 end
 
+M.get_supported = function(filetype, method)
+    local ft_map = require("null-ls.builtins._meta.filetype_map")
+    local supported = ft_map[filetype] or {}
+    if method then
```
jose-elias-alvarez, line 80: Is there a case when we would call this without a method?

lua/null-ls/info.lua
```diff
@@ -65,29 +78,82 @@ end
...
         log:warn("failed to get info: buffer is not attached")
         return
     end
 
+    local create_method_info = function(method)
+        local name = methods.readable[method]
+        local active_sources = {}
+        local supported_sources = {}
+
+        for _, source in ipairs(sources.get_available(ft, method) or {}) do
```
jose-elias-alvarez, line 92:
```suggestion
        for _, source in ipairs(sources.get_available(ft, method)) do
```

This will always return a table.

lua/null-ls/methods.lua
```diff
@@ -47,13 +47,13 @@ local overrides = {
...
 
 local readable_map = {
-    [internal_methods.CODE_ACTION] = "Code actions",
-    [internal_methods.DIAGNOSTICS] = "Diagnostics",
-    [internal_methods.DIAGNOSTICS_ON_SAVE] = "Diagnostics on save",
-    [internal_methods.FORMATTING] = "Formatting",
-    [internal_methods.RANGE_FORMATTING] = "Range formatting",
-    [internal_methods.HOVER] = "Hover",
-    [internal_methods.COMPLETION] = "Completion",
+    [internal_methods.CODE_ACTION] = "code_actions",
```
jose-elias-alvarez, line 50: See my comment on line 97 of `info.lua`. If we use the method and not the readable name, I think we can avoid changing this at all.

jose-elias-alvarez (commented):
I like this! Apart from my comments, I think we could improve the layout by first showing active methods + their sources, then showing inactive methods that have available sources below that (possibly with a divider). I think that would make it easier to parse at a glance.

Also, this PR made me realize that we're not currently handling sources that are active for all filetypes in the metadata script. I think it'll complicate the logic somewhat, but we should try to handle that here, too.

lua/null-ls/info.lua
```diff
@@ -65,29 +78,82 @@ end
...
+        return info_lines
+    end
+
     local lines = {}
 
-    local log_path = c.get().debug and log:get_path()
-        or "not enabled (this is normal; see the README if you need to enable logging)"
-    table.insert(lines, "null-ls log: " .. log_path)
+    local header = {
+        [[   _   _         _  _          _        ]],
```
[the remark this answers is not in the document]
kylo252: Is this good now? I tried looking for another font, but it's been hard, so I basically trimmed it down myself.  We can drop it completely if you'd prefer 😄

lua/null-ls/info.lua
kylo252: I should've explained the problem better, but we index the generated metadata by the builtin-type, e.g. `formatting`
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/2afe5c2e3484fc9e20589aae49e3829e0419596b/lua/null-ls/builtins/_meta/filetype_map.lua#L10-L13

So we have to actually either modify the generator script:
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/2afe5c2e3484fc9e20589aae49e3829e0419596b/scripts/autogen_metadata.lua#L35-L38

or fix/add the mapping.

lua/null-ls/sources.lua
kylo252: It's shorthand for returning all available methods for a filetype.

gegoune:
Just tried it out. Looks good, good job. I am not big fan of ascii art though. It looks rather unprofessional and out of place.

lua/null-ls/methods.lua
kylo252: Updated my comment to better explain the issue.

kylo252:
> I like this! Apart from my comments, I think we could improve the layout by first showing active methods + their sources, then showing inactive methods that have available sources below that (possibly with a divider). I think that would make it easier to parse at a glance.

Yeah, I wasn't sure how to organize the panel. Since it can be presented primarily in two different ways:

#### per-method

The current implementation consolidates related methods, but maybe we can create some kind of checklist for specifics (e.g. range formatting support).

#### per-source

This would give us more roam to add some more specific info about the source. I haven't implemented this, but I think it should actually be nicer looking. It should look something like this

```console
stylua
  methods: formatting | range formatting | on-type formatting
  filetypes: lua
  buffers(?): [12, 56, 100]
  cmd(?): <path-to-stylua>
```

We should only have more detailed info for the active buffer sources, while the rest are more minimal. I'm also not sure how the "supported" sources would be organized here, but probably per-method.

--- 

> Just tried it out. Looks good, good job. 

Thanks! 

> I am not big fan of ascii art though. It looks rather unprofessional and out of place.

I'm mostly thinking of it from an accessibility point-of-view. I basically want some kind of title indicating what this panel is for, and it's fun :)

lua/null-ls/info.lua
jose-elias-alvarez: Alright, I think I get it. I think we can follow the folder structure, then, meaning that diagnostics/diagnostics on save and formatting/formatting on save will be handled in the same way. We could then show more detail for each active source about its specific capabilities.

jose-elias-alvarez:
I like the per-source idea. Basically, I think the purpose of this window should be to show at a glance what null-ls can do with the current buffer, and showing the specific capabilities of active sources fits that goal. 

I personally don't see the need for showing supported sources, but I can see how it could be useful, and I know ALE provides something similar. I think that display can be minimal (and ideally split off from the active source display to avoid confusion):

```
Supported sources:
Diagnostics: luacheck, selene
Formatting: luaformat
```

That avoids the metadata-method mismatch problem, since we can just list them according to their place in the folder structure.

kylo252:
Okay, so I did a total revamp just as discussed, and here's the updated layout

<details>
<summary>Expand screenshot</summary>

image

</details>

Notes: 
- chose to not include wildcard sources in the "supported" section, just to avoid repeating the same info.
- added some basic highlighting, but it's optional if you think it's too much.

jose-elias-alvarez:
Looks excellent – I'll test this out in the next couple of days.

lua/null-ls/info.lua
```diff
@@ -40,65 +42,137 @@ local make_window = function(height_percentage, width_percentage)
...
+            "* path: " .. log:get_path(),
+        }
+        table.insert(highlights, { "Type", info_lines[1] })
+        return info_lines
+    end
+
+    local lines = {}
+
+    local header = {
+        "Null-LS",
```
jose-elias-alvarez, line 138:
```suggestion
        "null-ls",
```

Really nitpicky, but...

lua/null-ls/methods.lua
```diff
@@ -46,16 +46,6 @@ local overrides = {
     },
 }
 
-local readable_map = {
```
jose-elias-alvarez, line 49: I guess it's not a huge difference for now, but are you sure we want to remove the readable names completely? I wonder if in the future we'll have methods that are harder to parse.

jose-elias-alvarez (commented):
Aside from the mother of all nitpicks, this looks great!

lua/null-ls/methods.lua
kylo252: Are you familiar with YAGNI? 😄

lua/null-ls/info.lua
kylo252: I actually wanted to center this, but couldn't figure out the minimal steps for that. There are also a few more tweaks I want to make, but they can be in a different PR.

lua/null-ls/methods.lua
jose-elias-alvarez: Fair point!

jose-elias-alvarez:
This looks excellent and is a big improvement. Thank you for working on this!

PR #477, opened 2021-12-21
fix: Add support for dotted filetypes

rliebz:
Neovim supports setting multiple filetypes through the use of a dot
separator. From `:h 'filetype'`:

    When a dot appears in the value then this separates two filetype
    names.  Example:
        /* vim: set filetype=c.doxygen : */ ~
    This will use the "c" filetype first, then the "doxygen" filetype.
    This works both for filetype plugins and for syntax files.  More than
    one dot may appear.

By respecting this functionality, users can set filetypes such as
`json.openapi` or `yaml.docker-compose` and get out-of-the-box support
for multiple filetype-specific sources in a single file.

Dotted filetypes continue to be supported as a unique filetype for
backward compatibility and cases where sources should only be made
available at the intersection of filetypes.

For files where a source is explicitly configured to be both enabled and
disabled from different filetypes, I opted to give precedence to the
disabled filetype. It's trivial to swap if that behavior is less
desirable. In practice this might never come up.

### DIFF
--- a/lua/null-ls/sources.lua
+++ b/lua/null-ls/sources.lua
@@ -42,26 +42,48 @@ local register_source = function(source)
     registered.names[source.name] = true
 end
 
+local matches_filetype = function(source, filetype)
+    if filetype:find("%.") then
+        local filetypes = vim.split(filetype, ".", { plain = true })
+        table.insert(filetypes, filetype)
+
+        local is_match = source.filetypes["_all"]
+        for _, ft in ipairs(filetypes) do
+            if source.filetypes[ft] == false then
+                return false
+            end
+
+            is_match = is_match or source.filetypes[ft]
+        end
+
+        return is_match
+    end
+
+    return source.filetypes[filetype] or source.filetypes["_all"] and source.filetypes[filetype] == nil
+end
+
+local matches_method = function(source, method)
+    if source.methods[method] then
+        return true
+    end
+
+    if methods.overrides[method] then
+        for m in pairs(methods.overrides[method]) do
+            if source.methods[m] then
+                return true
+            end
+        end
+    end
+
+    return false
+end
+
 M.is_available = function(source, filetype, method)
     if source._disabled or source.generator._failed then
         return false
     end
 
-    local filetype_matches = not filetype
-        or source.filetypes["_all"] and source.filetypes[filetype] == nil
-        or source.filetypes[filetype] == true
-
-    local method_matches = not method or source.methods[method] == true
-    if not method_matches and methods.overrides[method] then
-        for m in pairs(methods.overrides[method]) do
-            if source.methods[m] then
-                method_matches = true
-                break
-            end
-        end
-    end
-
-    return filetype_matches and method_matches
+    return (not filetype or matches_filetype(source, filetype)) and (not method or matches_method(source, method))
 end
 
 M.get_available = function(filetype, method)
--- a/test/spec/sources_spec.lua
+++ b/test/spec/sources_spec.lua
@@ -60,6 +60,28 @@ describe("sources", function()
             assert.truthy(is_available)
         end)
 
+        it("should return true if multiple filetypes are set and one matches", function()
+            local is_available = sources.is_available(mock_source, "foo.lua.bar")
+
+            assert.truthy(is_available)
+        end)
+
+        it("should return true if a dotted filetype is specified", function()
+            mock_source.filetypes["jinja.html"] = true
+
+            local is_available = sources.is_available(mock_source, "jinja.html")
+
+            assert.truthy(is_available)
+        end)
+
+        it("should return false if multiple filetypes are set and one is disabled", function()
+            mock_source.filetypes["bar"] = false
+
+            local is_available = sources.is_available(mock_source, "foo.lua.bar")
+
+            assert.falsy(is_available)
+        end)
+
         it("should return true if filetype includes _all key", function()
             mock_source.filetypes["_all"] = true
 


lua/null-ls/sources.lua
```diff
@@ -48,8 +48,21 @@ M.is_available = function(source, filetype, method)
     end
 
     local filetype_matches = not filetype
-        or source.filetypes["_all"] and source.filetypes[filetype] == nil
-        or source.filetypes[filetype] == true
+
+    if type(filetype) == "string" then
```
jose-elias-alvarez, line 52: `filetype` can only be a string or `nil`, so can we instead return early if it's `nil`?

lua/null-ls/sources.lua
```diff
@@ -48,8 +48,21 @@ M.is_available = function(source, filetype, method)
...
+        local filetypes = vim.split(filetype, ".", { plain = true })
```
jose-elias-alvarez, line 53: I feel like it makes sense to check if the filetype actually contains a dot and return early if it doesn't, too.

lua/null-ls/sources.lua
```diff
@@ -48,8 +48,21 @@ M.is_available = function(source, filetype, method)
...
+        table.insert(filetypes, filetype)
+
+        for _, ft in ipairs(filetypes) do
+            if source.filetypes[ft] == false then
+                return false
+            end
+
+            filetype_matches = filetype_matches
```
jose-elias-alvarez, line 61: Does this need to be in the loop?

jose-elias-alvarez:
This is a good idea, and I can see from the test cases that it's working as-is. If we're going to complicate the filetype detection code, I think it makes sense to split up filetype and method matching into separate functions so that we can handle these cases in order (with early returns for each):

1. `filetype` is `nil`, in which case we return `true`
2. `filetype` is a plain filetype (one that doesn't contain a dot), so we return according to the previous logic
3. `filetype` contains a dot, in which case we split it and check each individual filetype

What do you think?

lua/null-ls/sources.lua
rliebz: Moved to a method and done.

lua/null-ls/sources.lua
rliebz: The goal was to avoid repeating logic for the case where there is one file type and the case where there are multiple file types. So the idea is to convert `"foo.bar"` into `{ "foo", "bar", "foo.bar" }`, but also `"foo"` into `{ "foo" }`. Then for non-dotted filetypes, we just do one iteration through the loop and don't have to special case either one.

That said there was definitely a small performance bug here where `"foo"` would turn into `{"foo", "foo"}`, so that's now fixed!

rliebz:
@jose-elias-alvarez Refactored this, and borrowed the signature shape from `matches_query` for consistency. It seemed more natural to include the `nil` check inside the helper functions, but I can definitely move those out if you'd prefer. 

Thanks for the speedy review, and let me know if you have any other thoughts!

lua/null-ls/sources.lua
```diff
@@ -42,26 +42,51 @@ local register_source = function(source)
...
-M.is_available = function(source, filetype, method)
-    if source._disabled or source.generator._failed then
-        return false
+local matches_filetype = function(source, filetype)
+    if not filetype then
+        return true
+    end
+
+    local filetypes = vim.split(filetype, ".", { plain = true })
+    if string.find(filetype, ".") then
```
jose-elias-alvarez, line 51:
```suggestion
    if string.find(filetype, "%.") then
```

We need to escape this or else it matches any character (so this condition will always be met if `filetype` is not an empty string).

lua/null-ls/sources.lua
```diff
@@ -42,26 +42,51 @@ local register_source = function(source)
...
+        table.insert(filetypes, filetype)
+    end
+
+    is_match = source.filetypes["_all"]
```
jose-elias-alvarez, line 55:
```suggestion
    local is_match = source.filetypes["_all"]
```

lua/null-ls/sources.lua
```diff
@@ -42,26 +42,51 @@ local register_source = function(source)
...
+
+    for _, ft in ipairs(filetypes) do
+        if source.filetypes[ft] == false then
+            return false
+        end
+
+        is_match = is_match or source.filetypes[ft]
```
jose-elias-alvarez, line 62: We can return early here as soon as there is a match, I think.

lua/null-ls/sources.lua
```diff
@@ -42,26 +42,51 @@ local register_source = function(source)
...
-    return filetype_matches and method_matches
+    return false
+end
+
+M.is_available = function(source, filetype, method)
+    if source._disabled or source.generator._failed then
+        return false
+    end
+
+    return matches_filetype(source, filetype) and matches_method(source, method)
```
jose-elias-alvarez, line 89: I think I do prefer moving the `nil` checks for `filetype` and `method` here, since it seems a little odd for `matches_filetype` to return `true` if it's `nil`.

lua/null-ls/sources.lua
```diff
@@ -42,26 +42,51 @@ local register_source = function(source)
     registered.names[source.name] = true
 end
 
-M.is_available = function(source, filetype, method)
-    if source._disabled or source.generator._failed then
-        return false
+local matches_filetype = function(source, filetype)
```
jose-elias-alvarez, line 45: Hmm, I'm still not in love with the structure of this. I see what you mean about combining the logic for both cases, but I think what would help me follow this more easily would be to handle dotted filetypes and single filetypes separately:

```lua
local matches_filetype = function(source, filetype)
    -- multiple filetypes in a single string (e.g. yaml.docker-compose)
    if filetype:find("%.") then
        -- dotted filetype logic goes here and returns
        return is_match
    end

    local is_match = ... -- single filetype logic
    return is_match
end
```

Structuring it like this (and keeping a comment to explain the need for the special case logic) would help me and other contributors understand what's going on, I think.

jose-elias-alvarez (changes requested):

lua/null-ls/sources.lua
rliebz: Done.

lua/null-ls/sources.lua
rliebz: Done.

lua/null-ls/sources.lua
rliebz: Good catch, done.

lua/null-ls/sources.lua
rliebz: We need to loop through every file type if we want disabled filetypes to take precedence over enabled filetypes.

The test case that returning early would break is: 

```
should return false if multiple filetypes are set and one is disabled
```

lua/null-ls/sources.lua
rliebz: Done.

rliebz (commented):
Appreciate the feedback, all comments addressed!

jose-elias-alvarez (approved):

jose-elias-alvarez:
Perfect, thanks!

PR #478, opened 2021-12-21
doc: use formatting_seq_sync() instead of formatting_sync()

zbindenren:
I had different formatters for one filetype  (whitespace and trailing new lines and golang). If I used your autocommand in doc, nvim always asked which lsp source I want to use:

```console
Select a Language Server:
(1) lsp, (2) null-ls:
```

This was pretty annoying. There is a better way: `vim.lsp.buf.formatting_seq_sync()` instead of `vim.lsp.buf.formatting_sync()`.

More info on [stackoverflow](https://stackoverflow.com/questions/67550082/neovim-select-a-language-prompt).

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Please **do not** link to or post your entire Neovim configuration.
 ### How do I format files?
 
 null-ls formatters run when you call `vim.lsp.buf.formatting()` or
-`vim.lsp.buf.formatting_sync()`. If a source supports it, you can run range
+`vim.lsp.buf.formatting_seq_sync()`. If a source supports it, you can run range
 formatting by visually selecting part of the buffer and calling
 `vim.lsp.buf.range_formatting()`.
 
@@ -216,7 +216,7 @@ require("null-ls").setup({
     -- you can reuse a shared lspconfig on_attach callback here
     on_attach = function(client)
         if client.resolved_capabilities.document_formatting then
-            vim.cmd("autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync()")
+            vim.cmd("autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_seq_sync()")
         end
     end,
 })


jose-elias-alvarez:
I'd prefer not to replace the current documentation with this because there are cases where the user wants to ignore the language server's formatting capabilities and just use null-ls instead (most notably `tsserver`), and we already mention [a workaround](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Avoiding-LSP-formatting-conflicts) for that case in the wiki. (I have no idea why it didn't work for the Stack Overflow poster.)

I wouldn't be opposed to mentioning `formatting_seq_sync` as a way to format with both null-ls and an LSP client, though I imagine that case is rarer.

bpinto:
@jose-elias-alvarez would it be possible to use something like this for sequentially formatting source code with two or more null-ls builtin formatters?

jose-elias-alvarez:
@bpinto That's right, I don't think it avoids the undo history issue. Also, neovim/neovim#16206 hints at changes coming to the formatting API, so I'd rather not rely on `formatting_seq_sync`.

zbindenren:
> I'd prefer not to replace the current documentation with this because there are cases where the user wants to ignore the language server's formatting capabilities and just use null-ls instead (most notably `tsserver`), and we already mention [a workaround](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Avoiding-LSP-formatting-conflicts) for that case in the wiki. (I have no idea why it didn't work for the Stack Overflow poster.)
> 
> 
> 
> I wouldn't be opposed to mentioning `formatting_seq_sync` as a way to format with both null-ls and an LSP client, though I imagine that case is rarer (since "good" LSP formatting is generally better and faster than null-ls formatting)

I agree, I would prefer to disable the whitespace formatter for go. But I did't find a way to exclude a formatter for certain file types (i.e. go in this example). As far as I have seen just the opposite is possible with the `source` parameter.

jose-elias-alvarez:
It's possible to disable specific filetypes using the `disabled_filetypes` config key, described [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#filetypes). Does that solve your case? I'm not sure I see the connection with the `formatting_seq_sync()` issue, so maybe I'm misunderstanding something.

jose-elias-alvarez:
Going to close this, since for the reasons I mentioned I'd prefer to keep `formatting_sync` as the recommended approach. Feel free to add a page / section to the wiki about `formatting_seq_sync`.

PR #479, opened 2021-12-21
fix(utils): escape wildcard characters in folder path

staticssleever668:
If a folder has a wildcard in it's path, the plugin gives a `Reverse range in character class` error. I had the same problem in the past with another plugin: https://github.com/airblade/vim-rooter/issues/122. The solution is simple: escape any wildcard characters.

Here is a small snippet to reproduce my error: `folder_name="[u-a]" && mkdir -p "$folder_name" && touch "$folder_name/readme.md" && nvim "$folder_name/readme.md"`.


jose-elias-alvarez:
I see, thanks for the fix!

PR #486, opened 2021-12-25
proposal: add highlight matching for flake8

kozachynskyi:
related issue https://github.com/jose-elias-alvarez/null-ls.nvim/issues/487

### DIFF
--- a/doc/HELPERS.md
+++ b/doc/HELPERS.md
@@ -272,8 +272,6 @@ A function called when the user registers the source. Intended for use with the
 helper `factory` functions described above, but any function that returns a
 valid generator will work.
 
-Note that if defined, `factory` should handle `condition` (described above)
-
 ### filetypes
 
 A list of filetypes for the source, as described in [MAIN](MAIN.md).


lua/null-ls/helpers.lua
```diff
@@ -529,13 +529,20 @@ M.diagnostics = (function()
...
+                        _, end_col = line:find(results, 1, true)
+                        end_col = end_col + 1
+                        if end_col > tonumber(entries["col"]) then
+                            return end_col
+                        end
                     end
 
-                    _, end_col = line:find(quote, 1, true)
-                    return end_col and end_col > tonumber(entries["col"]) and end_col or nil
+                    return nill
```
jose-elias-alvarez, line 545:
```suggestion
                    return nil
```

lua/null-ls/helpers.lua
```diff
@@ -529,13 +529,20 @@ M.diagnostics = (function()
...
                 end_col = function(entries, line)
                     local end_col = entries["end_col"]
-                    local quote = entries["_quote"]
-                    if end_col or not quote or not line then
-                        return end_col
+                    local message = entries["message"]
+                    local code = entries["code"]
+                    
+                    if code == "F841" then
+                        pattern = [[local variable %'(.*)%']]
```
jose-elias-alvarez, line 536:
```suggestion
                        local pattern = [[local variable %'(.*)%']]
```

jose-elias-alvarez (commented):
The issue with this approach is that `from_quote` is supposed to be a general-purpose adapter, so I'd prefer not to include any `flake8`-specific logic there. I think it makes sense to change this into a special-purpose adapter that's part of `flake8.lua`

PR #490, opened 2021-12-26
feat: standardjs fine grained severities

bluz71:
My initial integration of standardjs to null-ls treated all diagnostics emitted as errors.

However, standardjs does distinguish between syntax errors (denoted by "Parsing error") and warnings (usually code-style).

An example of a syntax error:

  /home/fred/foo.js:2:1: Parsing error: Unexpected token

And an example of a code-style warning:

  /home/fred/foo.js:1:35: Strings must use singlequote.

The null-ls integration of standardjs now accounts for errors as against warnings.


jose-elias-alvarez:
Looks great, thank you!

PR #495, opened 2021-12-26
Fix eslint_d code_actions fix range computation

dhleong:
Also switches away from using cursor movement to do the position computation.

The issue repros with multi-line changes, like import reordering. As-is, for an input like:

```typescript
import b;
import a;
```

which should look like:

```typescript
import a;
import b;
```

applying the fix will incorrectly result in:

```typescript
import b;
import a;
import b;
```

### DIFF
--- a/lua/null-ls/builtins/code_actions/eslint_d.lua
+++ b/lua/null-ls/builtins/code_actions/eslint_d.lua
@@ -12,24 +12,13 @@ local get_offset_positions = function(content, start_offset, end_offset)
     start_offset = vim.str_byteindex(to_string, start_offset + 1)
     end_offset = vim.str_byteindex(to_string, end_offset + 1)
 
-    -- save original window position and virtualedit setting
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
The change to the range computation seems fine. Could you verify whether the new position computation method handles multibyte characters (e.g. accented characters or emoji)? You can see jose-elias-alvarez/nvim-lsp-ts-utils#38 for the original motivation. At the time, I wasn't able to find a better solution.

dhleong:
It does seem to! I added some multibyte characters in a comment before the imports in the same repro I used above and the code was still successfully replaced, eg as in:

```typescript
// multibyte: éééé
import B from '@app/modules/B';
import A from '@app/modules/A';
```

jose-elias-alvarez:
Awesome, in that case this is a strict improvement (I was never thrilled with the cursor movement hack). Thanks for the PR!

PR #500, opened 2021-12-27
fix(credo): use callback instead of returning value

giddie:
I thought I was going crazy when I simply couldn't get Credo to show error results after I got ESLint working pretty effortlessly. But I think I accidentally stumbled on some code that was broken by a recent refactor.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/credo.lua
+++ b/lua/null-ls/builtins/diagnostics/credo.lua
@@ -31,7 +31,7 @@ return h.make_builtin({
         format = "raw",
         to_stdin = true,
         from_stderr = true,
-        on_output = function(params)
+        on_output = function(params, done)
             local issues = {}
 
             -- report any unexpected errors, such as partial file attempts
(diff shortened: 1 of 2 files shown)


giddie:
Although... now I'm wondering if this is the correct fix. The intention was to fix an issue that seemingly presented display of the errors that were successfully parsed from credo. I thought the issue was simply that the return mechanism for the results had switched to a callback pattern, but although the above fix does now show the errors as expected, then are not always cleared as they should be. So maybe there is something else going on here.

giddie:
Ah! Never mind - I finally realised that I hadn't accounted for the multiple return sites :)

jose-elias-alvarez:
You're right – the previous change that changed `format` to `raw` necessitated this change, and I'm not sure how it got missed. I went ahead and updated the tests to account for this change, but it looks good to me otherwise. Thanks!

jc00ke:
@giddie you're my hero! I was going crazy as well! I was filling out an issue then had to run an errand, came back and it was fixed. :pray: 

@jose-elias-alvarez thanks for the quick merge!

PR #503, opened 2021-12-28
feat: forward on_exit callback from lsp client

ii14:
Adds on_exit callback.

I'm setting some stuff up in on_init and on_attach that needs to be cleaned up in on_exit.


jose-elias-alvarez:
Thanks!

PR #504, opened 2021-12-28
fix(shellcheck): add bash filetype

ii14:
Shellcheck supports bash.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2080,7 +2080,7 @@ local sources = { null_ls.builtins.diagnostics.shellcheck }
 
 ##### Defaults
 
-- `filetypes = { "sh" }`
+- `filetypes = { "sh", "bash" }`
 - `command = "shellcheck"`
 - `args = { "--format", "json1", "--source-path=$DIRNAME", "--external-sources", "-" }`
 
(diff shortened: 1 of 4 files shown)


kylo252:
This is not necessary, as it's already should be handled in vim by default. Could you please give an example where it hasn't been working for you already?

ii14:
Handled how? For me null-ls is just not attaching to bash buffers unless I explicitly add it to filetypes. If it should just work then I can track down what's causing that, if it's necessary. Maybe it's something with other plugins I use, like for example filetype.nvim.

jose-elias-alvarez:
I'm using filetype.nvim without any additional configuration, and running `:e test.bash | set ft?` shows `filetype=sh`. I wonder what's causing the difference in behavior?

ii14:
Shebang detection in filetype.nvim is causing that. If you create a file without any extension, like `abc`, and put a `#!/bin/bash` shebang on it, it will set the filetype to `bash`. Interestingly, when you add `.bash` extension, the filetype goes back to `sh`.

jose-elias-alvarez:
Very strange – on my end, creating an empty file without an extension and adding the `bash` shebang line still shows `filetype=sh`.

ii14:
Yes, I suppose this is a filetype.nvim issue. Sorry for taking up your time, I forgot that this is a thing with `sh` filetype :)

itaranto:
Consider merging this since as for Neovim 0.9, some bash files are still detected as `bash` file type.

For example, I have [this](https://gitlab.com/itaranto/dotfiles/-/blob/master/sys.home/etc/bash.bashrc.local) file marked as `bash` file type with `# vim:ft=bash`.

IMHO it's not OK that Vim and Neovim treat `sh` and `bash` as the same file type. But I guess that's another discussion.

PR #505, opened 2021-12-28
feat(luacheck): enable project diagnostics

hungpt2702:
This pull request enables project diagnostics for `luacheck`.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/luacheck.lua
+++ b/lua/null-ls/builtins/diagnostics/luacheck.lua
@@ -1,28 +1,27 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 
-local DIAGNOSTICS = methods.internal.DIAGNOSTICS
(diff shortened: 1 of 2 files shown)


kylo252:
Is this not supposed to work interactively? because you've effectively removed reading from stdin in this case.

jose-elias-alvarez:
Right, I think this changes the current behavior of the `luacheck` source in a way that's surprising for current users. Maybe this would be best as a separate source, something like `luacheck_project`?

hungpt2702:
@jose-elias-alvarez I could implement it as a separate source. However, I wonder if it is currently possible to make project diagnostics update on change?

jose-elias-alvarez:
Unless a linter supports something like JSON via `stdin`, I don't think it's really possible, since there's no way for the linter to be aware of Neovim buffer state. If you look at your changes, you're telling `luacheck` to check all the files in the root directory, so it necessarily has to check files as they are on disk (not in Neovim). I also suspect it would cause performance issues on larger projects, so it's probably not a good idea in general.

PR #512, opened 2021-12-30
Fix broken mypy link

lasse16:
Previous link target was not handled as a web adress, but instead linked to [https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/mypy-lang.org](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/mypy-lang.org), which does not exist. I added in the link to the [official mypy github repo](https://github.com/python/mypy).


jose-elias-alvarez:
Thanks!

Issue #509, opened 2021-12-29
Run built-in sources only in the presence of specific file(s)


sQVe:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

The `make_conditional_utils` are using `vim.loop.cmd()` https://github.com/jose-elias-alvarez/null-ls.nvim/blob/e13a1cadf720ecb8a27a022beaf4608778ca8fdc/lua/null-ls/utils.lua#L142

This is really confusing since `null-ls` allows one to configure the `root_dir` and it breaks `root_has_file` on complex setups like monorepos etc. As a user I'm heavily assuming that it's using the output of `root_dir`.

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I tried doing this like it's implemented on:
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/1e131a0b3f52eb812c7c07f5e24aee90c0ee8967/lua/null-ls/client.lua#L66

But I got I got stumped since I don't have access to `fname` in the utility functions. Unsure how one best would implement this.

jose-elias-alvarez:
The resolved `root_dir` is available on the client, and we could theoretically try to get `root_dir` from the client and fall back to `vim.loop.cwd()`, as in [these lines](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/helpers.lua#L260-L261). The issue is that in a typical config (registering a source on null-ls setup), the client won't yet be initialized at the point the sources are registered, so this won't work.

The conditional utils predate root directory configuration and don't quite match the current architecture. Ideally, the condition would be checked when the source first runs, which would give it access to the name of an attached buffer, but that would require significant restructuring. 

I'll try to see if I can think of a solution that doesn't add too much complexity, but ultimately built-ins and their associated options are meant to handle simple cases. Any more complex cases can be handled by manual registration using autocommand events.

jose-elias-alvarez:
It's still a WIP but could you try #413? I tried to restructure things so that `condition` is checked on the source's first run, meaning that it will inherit `root_dir` from the client.

sQVe:
@jose-elias-alvarez For sure. I'll test it out as soon as possible, probably in a couple of hours.

jose-elias-alvarez:
[...]

PR #515, opened 2021-12-30
feat: add fallback dynamic cmd by default

kylo252:
Make use of `dynamic_command` as a fallback when the base command isn't found. This is particularly useful for node_modules, for example: `prettier`.

```lua
generator_opts = {
    command = "prettier",
    args = h.range_formatting_args_factory({ "--stdin-filepath", "$FILENAME" }),
    to_stdin = true,
    dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules
}
```

Related discussion: https://github.com/jose-elias-alvarez/null-ls.nvim/pull/349#pullrequestreview-807707808

### DIFF
--- a/lua/null-ls/helpers/command_resolver.lua
+++ b/lua/null-ls/helpers/command_resolver.lua
@@ -1,0 +1,54 @@
+local log = require("null-ls.logger")
+local s = require("null-ls.state")
+local u = require("null-ls.utils")
+local fmt = string.format
+
+local M = {}
+
+M.generic = function(params, prefix)
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    if resolved then
+        if resolved.command then
+            log:debug(fmt("Using cached value [%s] as the resolved command for [%s]", resolved.command, params.command))
+        end
+        return resolved.command
+    end
+
+    local executable_to_find = prefix and u.path.join(prefix, params.command) or params.command
+    log:debug("attempting to find local executable " .. executable_to_find)
+
+    local root = u.get_root()
+
+    resolved = {}
+
+    u.path.traverse_parents(params.bufname, function(dir)
+        local command = u.path.join(dir, executable_to_find)
+        if u.is_executable(command) then
+            log:trace(fmt("resolved dynamic command for [%s] with cwd=%s", executable_to_find, dir))
+            resolved.command = command
+            resolved.cwd = dir
+            return true
+        end
+
+        -- use cwd as a stopping point to avoid scanning the entire file system
+        if dir == root then
+            return true
+        end
+    end)
+
+    if not resolved.command then
+        log:debug(fmt("Unable to resolve command [%s], skipping further lookups", executable_to_find))
+    end
+
+    s.set_resolved_command(params.bufnr, params.command, { command = resolved.command or false, cwd = resolved.cwd })
+    return resolved.command
+end
+
+M.from_node_modules = function(params)
+    -- try the local one first by default but always fallback on the pre-defined command
+    -- this needs to be done here to avoid constant lookups
+    -- we're checking if the global is executable to avoid spawning an invalid command
+    return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
+end
+
+return M
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -62,8 +62,7 @@ M.has_version = function(ver)
 end
 
 M.is_executable = function(cmd)
-    local is_executable = vim.fn.executable(cmd) > 0
-    if is_executable then
+    if cmd and vim.fn.executable(cmd) == 1 then
         return true
     end
 
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -50,12 +50,12 @@ null_ls.setup({ sources = sources })
 ```
 
 To run built-in sources, the command specified below must be available on your
-`$PATH` and visible to Neovim. For example, to check if `eslint` is available,
+`$PATH` and visible to Neovim. For example, to check if `stylua` is available,
 run the following (Vim, not Lua) command:
 
 ```vim
 " should echo 1 if available (and 0 if not)
-:echo executable("eslint")
+:echo executable("stylua")
 ```
 
 ## Configuration
@@ -661,6 +661,7 @@ local sources = { null_ls.builtins.formatting.eslint }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue" }`
 - `command = "eslint"`
 - `args = { "--fix-dry-run", "--format", "JSON", "--stdin", "--stdin-filename", "$FILENAME" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [eslint_d](https://github.com/mantoni/eslint_d.js)
 
@@ -679,6 +680,7 @@ local sources = { null_ls.builtins.formatting.eslint_d }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue" }`
 - `command = "eslint_d"`
 - `args = { "--fix-to-stdout", "--stdin", "--stdin-filepath", "$FILENAME" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [fish_indent](https://linux.die.net/man/1/fish_indent)
 
@@ -1203,6 +1205,7 @@ local sources = { null_ls.builtins.formatting.prettier }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue", "css", "scss", "less", "html", "json", "yaml", "markdown", "graphql" }`
 - `command = "prettier"`
 - `args = { "--stdin-filepath", "$FILENAME" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [prettier_d_slim](https://github.com/mikew/prettier_d_slim)
 
@@ -1225,6 +1228,7 @@ local sources = { null_ls.builtins.formatting.prettier_d_slim }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue", "css", "scss", "less", "html", "json", "yaml", "markdown", "graphql" }`
 - `command = "prettier_d_slim"`
 - `args = { "--stdin", "--stdin-filepath", "$FILENAME" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [prettierd](https://github.com/fsouza/prettierd)
 
@@ -1245,6 +1249,7 @@ local sources = { null_ls.builtins.formatting.prettierd }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue", "css", "scss", "less", "html", "json", "yaml", "markdown", "graphql" }`
 - `command = "prettierd"`
 - `args = { "$FILENAME" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [prettier-standard](https://github.com/sheerun/prettier-standard)
 
@@ -1262,8 +1267,9 @@ local sources = { null_ls.builtins.formatting.prettier_standard }
 ##### Defaults
 
 - `filetypes = { "javascript", "javascriptreact" }`
-- `command = "prettier-standard"`
+- `command = "prettier_standard"`
 - `args = { "--stdin" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [prismaFmt](https://github.com/prisma/prisma-engines)
 
@@ -1396,6 +1402,7 @@ local sources = { null_ls.builtins.formatting.rustywind }
 - `filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue", "svelte", "html", }`
 - `command = "rustywind"`
 - `args = { "--stdin" }`
+- `dynamic_command = require("null-ls.helpers.command_resolver").from_node_modules`
 
 #### [scalafmt](https://github.com/scalameta/scalafmt)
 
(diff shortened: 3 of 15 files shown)


kylo252:
@jose-elias-alvarez, what do you think of this approach? I just prefer this for readability over `prefer_local` for now, but we can also use that instead.

The modularization is more for cleanup than anything else.

jose-elias-alvarez:
I like this idea and don't have any issue with the changes. Before making this into a default, I'd like to profile how much overhead this adds, both on the first search and with subsequent function calls. If it's opt-in, I think some amount of overhead is acceptable, but if not, we should think about this a bit more (especially since some of these Node sources are already pretty slow).

If the difference is minimal, we can go ahead and add the default to more Node sources, too.

kylo252:
> I like this idea and don't have any issue with the changes. Before making this into a default, I'd like to profile how much overhead this adds, both on the first search and with subsequent function calls. 

Notice how I kept the caching mechanism, however, that is still done per buffer which is mostly fine but could probably be further improved. Maybe we should do the caching per command or per root, or a combination of all three?
 
> If it's opt-in, I think some amount of overhead is acceptable, but if not, we should think about this a bit more (especially since some of these Node sources are already pretty slow). 

Agreed, and this will probably be least noticeable for formatters anyway, especially if it's done during `on_attach`.

jose-elias-alvarez:
A user came up with a solution to a similar problem [here](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/509#issuecomment-1002950065). Ideally, buffers under the same `resolved_cwd` would automatically resolve to the same command, though we'd have to balance that against adding complexity.

kylo252:
> A user came up with a solution to a similar problem [here](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/509#issuecomment-1002950065). Ideally, buffers under the same `resolved_cwd` would automatically resolve to the same command, though we'd have to balance that against adding complexity.

I think my approach should be simpler. I've also added a `cwd` attribute to make things a bit easier, even though it's probably not needed.

I would actually recommend renaming `cwd` to `root_pattern` to make it more consistent across the board, but that would be a breaking change, so I'll leave it to your discretion 😃

jose-elias-alvarez:
I'm not necessarily opposed to making a breaking change, since it's a more advanced option, but I'd be worried about confusion with lspconfig's `root_pattern`, which is different. I think `cwd` also makes it clearer that it's a per-source option.

I briefly looked at the most recent `per_workspace` changes, and I think I'd prefer to keep the original logic for now and improve on it later if we run into performance issues. If we're going to make this a default, I think it's a good idea to minimize the possibility of things going wrong for now.

lua/null-ls/builtins/code_actions/eslint_d.lua
```diff
@@ -160,6 +162,10 @@ return h.make_builtin({
             params.messages = output[1].messages
             return handler(params)
         end,
+        dynamic_command = cmd_resolver.from_node_modules,
+        cwd = function(params)
```
jose-elias-alvarez, line 166: Setting an explicit `cwd` prevents these commands from taking advantage of the resolved `cwd` from the command resolver. That's both slower (since it'll search the filesystem each time) and less accurate (since it won't know about where the command was actually found).

In the future, I'm sure we can improve this mechanism to start searching in likelier locations, but for now I think it's best to leave it off.

lua/null-ls/utils.lua
```diff
@@ -62,8 +62,8 @@ M.has_version = function(ver)
 end
 
 M.is_executable = function(cmd)
-    local is_executable = vim.fn.executable(cmd) > 0
```
jose-elias-alvarez, line 65: I'm curious – was this causing issues?

jose-elias-alvarez (commented):
Apart from my comment I think this looks good. I'm curious to see the profiling results to see if it makes sense to set this as a default.

lua/null-ls/utils.lua
kylo252: It had wanted to re-order the check here: https://github.com/kylo252/null-ls.nvim/blob/6e9cbbebb6de8711c83c9a91603e0e716afe5bb3/lua/null-ls/helpers/command_resolver.lua#L39

Which meant that `cmd` could be `nil` which I'm not sure is actually safe to pass to `executable()`, but it's not really needed otherwise.

lua/null-ls/builtins/code_actions/eslint_d.lua
kylo252: Sounds good! It's always tricky with caching, and I was actually worried if `cwd` would be able to get updated using my approach. I'll remove those for now.

kylo252: I removed the changes from `eslint_d` completely, since it will be handled in #521

kylo252:
> Apart from my comment I think this looks good. I'm curious to see the profiling results to see if it makes sense to set this as a default.

Any ideas for how to go about doing this profiling?

jose-elias-alvarez:
I'd use `vim.loop.now()` / `vim.loop.hrtime()` and capture 10-15 runs under the same environment, then average out the results (not sure if there is something more sophisticated).

kylo252:
> I'd use `vim.loop.now()` / `vim.loop.hrtime()` and capture 10-15 runs under the same environment, then average out the results (not sure if there is something more sophisticated).

oh man, this took waaay longer than I expected, but the results were kinda wild as I was basically getting 300% improvement when running with a `dynamic_cmd`, but I can't say that I'm certain of my testing methodology since `vim.lsp.buf.formatting()` doesn't have a return value. 

So here are the test files that I used along with the results, please do try them and let me know what you think:
https://gist.github.com/kylo252/d2659cba93e4498c6d268ecce0848aa0

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -0,0 +1,56 @@
...
+local u = require("null-ls.utils")
+
+local M = {}
+
+M.generic = function(params, prefix)
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    -- a string means command was resolved on last run
+    -- false means the command already failed to resolve, so don't bother checking again
+    if resolved and (type(resolved.command) == "string" or resolved.command == false) then
+        log:debug(string.format("Using cached value for [%s] at %s", params.command, resolved.command))
```
jose-elias-alvarez, line 12: Perhaps we could change this message when `resolved.command == false` to make it clearer that the command didn't resolve?

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -0,0 +1,56 @@
...
+        -- use cwd as a stopping point to avoid scanning the entire file system
+        if dir == root then
+            return true
+        end
+    end)
+
+    if resolved_command then
+        local is_executable, err_msg = u.is_executable(resolved_command)
+        if not is_executable then
+            log:warn(err_msg)
```
jose-elias-alvarez, line 42: We should throw an error here or handle this case to prevent caching a `resolved_command` that isn't executable.

lua/null-ls/utils.lua
jose-elias-alvarez: `vim.fn.executable(nil)` will throw an error, but perhaps we should handle that case specifically instead?

```lua
if not cmd then
    return false
end
```

jose-elias-alvarez:
The performance results are *really* strange – the difference isn't as dramatic on my setup, but both your script and manual tests confirm that `dynamic_command` is (after the first invocation) faster. I was able to replicate this with `stylua`, too, so doesn't seem to be just a `prettier` quirk. 

I won't pretend to understand how this can happen, but I guess we don't have to worry about performance concerns. I left a few additional comments, but other than what I mentioned I think this should be okay to merge.

kylo252:
- I changed so that we're only assigning `resolved.command` to a valid value during the look up, this simplifies things overall
- I've added a separate log entry to indicate that a command wasn't resolved and won't be looked up again

Hopefully the logic is clear enough. Let me what you think.

lua/null-ls/formatting.lua
```diff
@@ -91,7 +91,9 @@ M.handler = function(method, original_params, handler)
                 return u.make_params(original_params, methods.map[method])
             end,
             callback = function(edits, params)
-                M.apply_edits(edits, params)
+                if edits and #edits > 0 then
```
jose-elias-alvarez, line 94: This is a sensible change but I'd prefer to leave it out for now, since #519 will change this entirely.

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -0,0 +1,53 @@
...
+local u = require("null-ls.utils")
+local fmt = string.format
+
+local M = {}
+
+M.generic = function(params, prefix)
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    -- a string means command was resolved on last run
+    -- false means the command already failed to resolve, so don't bother checking again
+    if resolved and resolved.command then
```
jose-elias-alvarez, line 12: Are we still handling the case where `resolved_command` was explicitly set to `false`? As far as I can tell we are running the scan again on each invocation, but the point of `false` was to indicate that we know the last attempt failed and we don't have to do it again.

lua/null-ls/formatting.lua
kylo252: Without this then we get  `received edits` even when the command was unable to unresolve

```logtalk
[DEBUG Thu 06 Jan 2022 05:19:16 PM CET] ...rt/null-ls.nvim/lua/null-ls/helpers/command_resolver.lua:18: attempting to find local executable node_modules/.bin/prettier
[WARN  Thu 06 Jan 2022 05:19:16 PM CET] ...rt/null-ls.nvim/lua/null-ls/helpers/command_resolver.lua:40: Unable to resolve command [node_modules/.bin/prettier], skipping further lookups
[DEBUG Thu 06 Jan 2022 05:19:16 PM CET] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:266: Using dynamic command for [prettier], got: nil
[DEBUG Thu 06 Jan 2022 05:19:16 PM CET] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:273: unable to resolve command [prettier]
[DEBUG Thu 06 Jan 2022 05:19:16 PM CET] ...ack/packer/start/null-ls.nvim/lua/null-ls/formatting.lua:55: received edits from generators
[DEBUG Thu 06 Jan 2022 05:19:16 PM CET] ...ack/packer/start/null-ls.nvim/lua/null-ls/formatting.lua:74: successfully applied edits
```

But this made me realize that the fallback solution I had is not perfect, because the generator attempted to spawn it without testing if it's available, (which is probably not too bad anyway, since it seems to timeout instead), this also without validating the `#edits`

```logtalk
[DEBUG Thu 06 Jan 2022 05:17:56 PM CET] ...rt/null-ls.nvim/lua/null-ls/helpers/command_resolver.lua:18: attempting to find local executable node_modules/.bin/prettier
[WARN  Thu 06 Jan 2022 05:17:56 PM CET] ...rt/null-ls.nvim/lua/null-ls/helpers/command_resolver.lua:40: Unable to resolve command [node_modules/.bin/prettier], skipping further lookups
[DEBUG Thu 06 Jan 2022 05:17:56 PM CET] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:266: Using dynamic command for [prettier], got: prettier
[DEBUG Thu 06 Jan 2022 05:17:56 PM CET] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:320: spawning command [prettier] at /tmp/test/nvim with args { "--stdin-filepath", "/home/hatsu/.local/share/nvim/lsp_servers/bash/node_modules/bash-language-server/node_modules/cssstyle/node_modules/prettier-linter-helpers/index.js" }
[DEBUG Thu 06 Jan 2022 05:18:01 PM CET] ...site/pack/packer/start/null-ls.nvim/lua/null-ls/loop.lua:138: command [prettier] timed out after 5000 ms
[DEBUG Thu 06 Jan 2022 05:18:01 PM CET] ...ack/packer/start/null-ls.nvim/lua/null-ls/formatting.lua:55: received edits from generators
[DEBUG Thu 06 Jan 2022 05:18:01 PM CET] ...ack/packer/start/null-ls.nvim/lua/null-ls/formatting.lua:74: successfully applied edits
```

lua/null-ls/helpers/command_resolver.lua
kylo252: oops, I completely missed that part since I was focused on the log message. maybe it should be moved into the cache handler instead? or is it fine like this?

lua/null-ls/formatting.lua
jose-elias-alvarez: This behavior is solved in #519, so I'd be happier leaving this out to prevent another rebase.

lua/null-ls/helpers/command_resolver.lua
jose-elias-alvarez: I don't think the cache handler should care about what's actually in the cache, so I think it's fine like this.

kylo252:
> I also think we should mention this in the documentation.

Not sure how much details you want to add. I would also like to really encourage looking into #552

doc/HELPERS.md
```diff
@@ -104,8 +104,8 @@ original command (if set) is available as `params.command`. The callback should
 return a string containing the command to run or `nil`, meaning that no command
 should run.
 
-`dynamic_command` runs every time its parent generator runs and can affect
-performance, so it's best to cache its output when possible.
+The results of `dynamic_command` are cached since it runs every time its parent
```
jose-elias-alvarez, line 107: I don't think this is accurate – the resolvers in this PR wrap `dynamic_command` to cache results and return them immediately, but this doesn't happen by default.

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -0,0 +1,56 @@
...
+        end
+
+        -- use cwd as a stopping point to avoid scanning the entire file system
+        if dir == root then
+            return true
+        end
+    end)
+
+    if not resolved.command then
+        log:warn(fmt("Unable to resolve command [%s], skipping further lookups", executable_to_find))
```
jose-elias-alvarez, line 42: I think we should use `debug` here, since this is a normal situation when using `from_node_modules` and `prefer_local`.

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -0,0 +1,56 @@
...
+local s = require("null-ls.state")
+local u = require("null-ls.utils")
+local fmt = string.format
+
+local M = {}
+
+M.generic = function(params, prefix)
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    -- a string means command was resolved on last run
+    -- false means the command already failed to resolve, so don't bother checking again
```
jose-elias-alvarez, line 11: We can remove or update this comment now since the type check is gone.

jose-elias-alvarez (commented):
A few minor changes left to make, but other than that I think this is ready.

jose-elias-alvarez:
Thanks! This went back-and-forth a bit, so could we just squash the commits into logical units before merging?

jose-elias-alvarez:
Thank you for all the work on this!

Issue #390, opened 2021-12-01
Calling vim.lsp.buf.formatting() results in error


DanielWeidinger:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.6.0

### null-ls config

local lspconfig = require'lspconfig'
local capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())
local null_ls = require("null-ls")

null_ls.config({
    sources = {
        null_ls.builtins.formatting.stylua,
        null_ls.builtins.formatting.prettier_d_slim,
        null_ls.builtins.code_actions.gitsigns,
        -- null_ls.builtins.completion.spell,
    },
})

lspconfig["null-ls"].setup({
    on_attach = on_attach,
    capabilities=capabilities
})

### Steps to reproduce

Run vim.lsp.buf.formatting() on a buffer which is included in the sources

### Expected behavior

Formatting of the buffer

### Actual behavior

Error and no formatting

### Debug log

[WARN  Mi 01 Dez 2021 17:39:20] ...iel/.vim/plugged/null-ls.nvim/lua/null-ls/generators.lua:45: failed to run generator: ...daniel/.vim/plugged/null-ls.nvim/lua/null-ls/helpers.lua:167: command stylua is not executable (make sure it's installed and on your $PATH)
[TRACE Mi 01 Dez 2021 18:47:11] /home/daniel/.vim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:121: received LSP request for method initialize
[TRACE Mi 01 Dez 2021 18:47:11] /home/daniel/.vim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:146: received LSP notification for method initialized
[TRACE Mi 01 Dez 2021 18:47:11] /home/daniel/.vim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:146: received LSP notification for method textDocument/didOpen
[TRACE Mi 01 Dez 2021 18:47:11] ...iel/.vim/plugged/null-ls.nvim/lua/null-ls/generators.lua:9: running generators for method NULL_LS_DIAGNOSTICS_ON_OPEN
[...]

Issue #234, opened 2021-10-04
Sequential formatting creates undo history for intermediate results


disrupted:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

When using multiple formatters (like isort and black for Python), the intermediate formatting results are applied to the buffer creating unwanted history entries in the undo tree.

I found this comment from Jose mentioning the issue with a potential solution, so figured I would open an issue for keeping track of it

> I think it's possible if we use a temporary buffer. [...] I can see that one advantage is to avoid creating 2 separate undo steps. I'll look into it, but I think we can consider it as a future improvement.

https://github.com/jose-elias-alvarez/null-ls.nvim/pull/93#issuecomment-894927671

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

haven't worked with temporary buffers, but it seems like an interesting approach

jose-elias-alvarez:
For reference, the approach I had in mind was:

1. Load the current buffer's content into a new buffer;
2. Sequentially apply the edits to the temporary buffer without diffing them;
3. After all edits have been applied, diff the temporary buffer against the original buffer;
4. Apply the diffed edits to the original buffer; and
5. Unload the temporary buffer.

This would require shuffling around the formatting logic and would also require some work to make the experience seamless (for example, I think we'd have to save and restore the user's previous buffer) but it should be possible.

jose-elias-alvarez:
I played around with this and found that it's more complicated than I expected. Even if everything goes as I described, there's a ton of edge cases / possible bad interactions, so I don't plan on working on this.

PR #519, opened 2021-12-30
refactor(formatting): use temporary buffer

jose-elias-alvarez:
The primary purpose of this PR is to restructure the formatting handler to use a temporary buffer. Edits are applied sequentially to the temporary buffer without diffing, then the real buffer is diffed against the temporary buffer to apply edits in one go. This has the following advantages:

1. This (theoretically) creates only a single undo step, since the real buffer is only edited once (more on this in a bit). Also, only triggering change events once is both cleaner and more accurate.
2. Multiple formatters should be slightly faster, since we only need to diff edits once.
3. We now properly use custom formatting handlers, which means users can now set up async formatting, as described in [this comment](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/390#issuecomment-984850679). This is a huge (perceived) speed boost for slower formatters, though it has disadvantages.
4. We can now handle errors more intelligently without breaking the buffer in between (no more `formatting_sync()` timeouts).
5. Cleaning up the code and avoiding empty handler calls sets us up to take advantage of future improvements to Neovim's formatting API.

The undo history thing is strange. It was first mentioned in #93, got its own issue in #234, and was mentioned as recently as #478. I don't use multiple formatters normally, but running `isort` and `black` on the example in the [isort README](https://github.com/PyCQA/isort) and undoing after formatting correctly restores the buffer to its pre-formatting state... both with this PR and without it (though I could have sworn it didn't work before). The test case added in this PR passes on earlier versions, too, so I'd appreciate feedback from @bpinto, @disrupted, and anyone else who was interested in this feature.

This PR also removes workarounds used to save cursor positions and marks. Since the workarounds were implemented, Neovim switched to using `nvim_buf_set_text` and we switched to using diff formatting, so formatting is overall less destructive (though not perfect). Users who require this behavior can set it up in a custom handler (which is what they'd need to do for LSP formatting, too). 

I'm curious to know about the undo history issue. Either way, I think the improvements here make this PR worth merging after more testing.


bpinto:
@jose-elias-alvarez This is amazing! I am on vacations so I can't test this thoroughly :(

I've applied this branch on my machine and I no longer see a blinking change coming from two incompatible prettier/eslint rules being applied in sequence. So as far as my quick experiment went, this is looking great!

> and undoing after formatting correctly restores the buffer to its pre-formatting state... both with this PR and without it (though I could have sworn it didn't work before).

Yeah, not sure what/when this was fixed, I couldn't reproduce this undo buffer issue either.

jose-elias-alvarez:
@bpinto Great to hear and thank you for taking the time to test it! I'll be busy for the next week or so and won't have time to deal with potential issues from merging this, but let me know if you have any more feedback between now and then.

lua/null-ls/formatting.lua
```diff
@@ -1,110 +1,100 @@
...
+        -- copy content and options to temp buffer
+        local temp_bufnr = api.nvim_create_buf(false, true)
+        api.nvim_buf_set_option(temp_bufnr, "eol", api.nvim_buf_get_option(bufnr, "eol"))
+        api.nvim_buf_set_option(temp_bufnr, "fileformat", api.nvim_buf_get_option(bufnr, "fileformat"))
+        api.nvim_buf_set_lines(temp_bufnr, 0, -1, false, api.nvim_buf_get_lines(bufnr, 0, -1, false))
+
+        local called_handler = false
+        local handler_wrapper = function(...)
+            -- calling handler twice breaks lsp.buf.formatting_sync()
+            if called_handler then
```
bpinto, line 37: Sorry if I'm missing something, but where is `called_handler` set to `true`?

test/spec/e2e_spec.lua
```diff
@@ -499,19 +459,32 @@ describe("e2e", function()
             lsp_wait(0)
 
             lsp.buf.formatting()
-            lsp_wait(0)
+            lsp_wait(100)
```
bpinto, line 462: I'm assuming this is because of the `math.random(50)`. If I may suggest, as I am not passionated about comments, I think we could have a `local per_formatter_timeout = 50` variable set in an external context and reuse this value on all tests. I think this could help with test readability in the future.

test/spec/e2e_spec.lua
```diff
@@ -499,19 +459,32 @@ describe("e2e", function()
...
             sources.register(builtins._test.second_formatter)
             sources.register(builtins._test.first_formatter)
             tu.edit_test_file("test-file.txt")
             lsp_wait(0)
 
             lsp.buf.formatting()
-            lsp_wait(0)
+            lsp_wait(100)
 
             assert.equals(u.buf.content(nil, true), "first\n")
```
bpinto, line 489: I don't understand how this test assertion is correct. Do you mind helping me understand it?

My doubts are the following:

- If each formatter timeout is random, it's not possible to guarantee that the second formatter will be executed after the former, right? So how can we assert the content will be `first\n` ?

- The `second_formatter` has the following response:

```
return done({ { text = params.content[1] == "first" and "sequential" or "second" } })
```

if this formatter runs after the `first_formatter` wouldn't it change the content of the buffer to either `sequential` or `second` ?

lua/null-ls/formatting.lua
jose-elias-alvarez: I think I mistakenly deleted it while refactoring. Thank you for catching this!

test/spec/e2e_spec.lua
jose-elias-alvarez: I like this idea and agree that it's pretty arbitrary without context. The integration tests broadly use two timeouts, one for mock sources and one for real sources, so we could consolidate them.

test/spec/e2e_spec.lua
jose-elias-alvarez: - A core requirement of sequential formatting is that formatters are executed sequentially, since each formatter has to act on the results of the previous formatter. That's what what the first test in the `sequential formatting` block is meant to test.
- We register `first_formatter` after `second_formatter`, and null-ls should run them in registered order, so the content of the buffer should reflect the output of `first_formatter` (in contrast to the first test in the block). A change that might make this clearer would be to have `first_formatter` also check the buffer's content and output something different in this case.

bpinto: 🤦 Thanks! I completely forgot this was sequential and I did not notice that the `second_formatter` had been registered first, that makes a lot of sense now.

bpinto:
I've used it today and this is so smooth. Amazing!

disrupted:
I've also used it yesterday and it looks fantastic! Thanks for working on this feature, @jose-elias-alvarez.

jose-elias-alvarez:
Awesome, glad to hear it's working well (even if the whole undo thing is a bit murky). I'll also continue to test this through the week and plan on merging it at the end of the week if there's no issues.

jose-elias-alvarez:
I'm pretty happy with this and haven't experienced any obvious issues, so I think it's safe to merge. Thanks again to everyone who tried this out!

PR #521, opened 2021-12-31
refactor: deduplicate eslint's config

kylo252:
make use of `builtin.with()` to define any minimal changes
instead of duplicating the entire configuration

```lua
null-ls.builtins.formatting.eslint.with({ command = "eslint_d" })
```


jose-elias-alvarez (commented):
I think this change is fine for `eslint` and `eslint_d` (we can add code actions, too) but I'm not sure we can do the same for the `prettier` family for these reasons:

1. They don't support the same options. Adding `--stdin` to the `prettier` argument list produces a `[warn] Ignored unknown option --stdin` message.  
2. `prettierd` doesn't support range formatting.
3. They don't support the same filetypes. In my experience, `prettierd` technically works but occasionally truncates large Markdown files, making it unusable. `prettier_d_slim` doesn't work at all on non-JavaScript filetypes, and [this PR that makes them work](https://github.com/mikew/prettier_d_slim/pull/11) hasn't been merged since June.

While we're making these changes, we can deprecate` prettier_d_slim`, which is something I've been meaning to do for a while. We can remove it from the documentation and add a deprecation notice to the built-in file. 

Edit: The `prettierd` repo is pretty active, and there is an [open issue](https://github.com/fsouza/prettierd/issues/237) to support the same options as vanilla `prettier`. In the future we may be able to merge them, but I don't think it's a good idea right now.

kylo252:
I only kept the changes to `eslint_d` for now. 

Since deprecating `prettier_d_slim` should probably go in a different PR and requires some docs update, so I leave that to you 😅

lua/null-ls/builtins/code_actions/eslint_d.lua
```diff
@@ -1,165 +1 @@
--- this and builtins/code_actions/eslint.lua should be kept in sync
-local h = require("null-ls.helpers")
-local methods = require("null-ls.methods")
-
-local CODE_ACTION = methods.internal.CODE_ACTION
-
-local api = vim.api
-
-local get_offset_positions = function(content, start_offset, end_offset)
```
kylo252, line 9: For future references: this function has a different implementation than the one in `eslint.lua`
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/d558fc2b763bd93c8f0c368720eb20995bf14980/lua/null-ls/builtins/code_actions/eslint.lua#L9

I kept that one because it seems to rely less on vim-script.

jose-elias-alvarez: That's the right call – a contributor recently refactored this to remove an ugly hack, but it wasn't reflected in the `eslint_d` builtin (which is a great justification for this PR).

jose-elias-alvarez:
To make sure it's clear, we should use `eslint_d` as the basis for code actions, since bbd7aed0ae49244678a82709a0e015a7126abbbc improved the logic. This will also need a rebase because of 031e839ac2a8ca1a4bc4111633a829dcae741c31.

lua/null-ls/builtins/code_actions/eslint_d.lua
kylo252: I will switch them around then.

jose-elias-alvarez:
This seems to be working well. Thank you for the PR!

Issue #520, opened 2021-12-30
feat: user configurable diagnostic source name


JA-Bar:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Ever since 859a3e0b1344e1bb937c1321e491b48dd05a1f24 passing the "name" option to a builtin source no longer changes the diagnostic's source name when displaying diagnostics, so if you have a custom command it looks like this:
```
Diagnostics:
line too long (93 > 90 characters) (/home/user/.local/share/nvim/installables/python/bin/flake8) 
```
instead of:
```
Diagnostics:
line too long (93 > 90 characters) (my_custom_name) 
```

I don't know if this was intended behavior, but it would be nice to have it back as a feature.

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

Merging the user's "name" option with the default generator options solves the issue. 
```diff
diff --git a/lua/null-ls/helpers/make_builtin.lua b/lua/null-ls/helpers/make_builtin.lua
index 0265014..a13db7d 100644
--- a/lua/null-ls/helpers/make_builtin.lua
+++ b/lua/null-ls/helpers/make_builtin.lua
@@ -43,6 +43,7 @@ local function make_builtin(opts)
         cwd = opts.cwd,
         diagnostics_format = opts.diagnostics_format,
         dynamic_command = opts.dynamic_command,
+        name = opts.name,
         runtime_condition = opts.runtime_condition,
         timeout = opts.timeout,
     })
```
But fails the test:
```
/home/user/.local/share/nvim/site/pack/packer/start/null-ls.nvim/test/spec/helpers/make_builtin_spec.lua                               
Fail    ||      make_builtin should return builtin with assigned opts                                                                                  
            ...art/null-ls.nvim/test/spec/helpers/make_builtin_spec.lua:36: Expected objects to be the same.
```
So I don't know if this feature was just thrown out or if it was an oversight of the last refactoring.

jose-elias-alvarez:
The fact that this worked before was an unintended side effect of indiscriminately treating all user options as generator options, but it should be fixed by 295c4b790e2a54106dd1f396e1b1dc18c19a6db8. Can you let me know?

Separately, all built-in sources should probably have a `name` field defined so that overriding `command` doesn't result in this behavior. I'll see if I can put together a script.

JA-Bar:
[...]

PR #524, opened 2021-12-31
chore: add source names to all builtins

JA-Bar:
As discussed in #520, default names were added to every builtin source.

I thought that the builtin source names took priority over the user provided name option, but apparently that's not the case and a user's name option works as expected.


jose-elias-alvarez:
Excellent, thank you!

Issue #487, opened 2021-12-25
end_col matching for flake8


kozachynskyi:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Currently flake8 doesn't detect end_col. If there is an error in variable, it will highlight whole line. One could add matching, that worked for [ale:](https://github.com/dense-analysis/ale/blob/dcec4b3c37880c0564b62b5a8e7ec424c64c8333/ale_linters/python/flake8.vim#L89)

I did some implementation, it works like a charm, however I am bad at lua. If you help me organizing code, I can add filters and write tests, probably ...)

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

Related pull request https://github.com/jose-elias-alvarez/null-ls.nvim/pull/486

PR #527, opened 2022-01-01
feat: flake8 end_col highlight adapter

kozachynskyi:
Closes https://github.com/jose-elias-alvarez/null-ls.nvim/issues/487
In this code I added following:

1. Only first character after start_col is highlighted, unless:
2. Some specific errors, where one can highlight variable names, because they are specified as error source in flake8 message

### DIFF
--- a/lua/null-ls/builtins/diagnostics/flake8.lua
+++ b/lua/null-ls/builtins/diagnostics/flake8.lua
@@ -25,23 +25,23 @@ local custom_end_col = {
             pattern = [[%'(.*)%' imported]]
         end
 
-        if pattern == nil then
+        if not pattern then
             return default_position
-        else
-            local results = message:match(pattern)
-            _, end_col = trimmed_line:find(results, 1, true)
-
-            if end_col == nil then
-                return default_position
-            else
-                end_col = end_col + start_col
-                if end_col > tonumber(start_col) then
-                    return end_col
-                else
-                    return default_position
-                end
-            end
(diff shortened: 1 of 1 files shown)


kozachynskyi:
BEFORE:
pic-window-220101-1232-30
AFTER:
pic-window-220101-1232-42

lua/null-ls/builtins/diagnostics/flake8.lua
```diff
@@ -3,6 +3,48 @@ local methods = require("null-ls.methods")
...
+                return default_position
+            else
+                end_col = end_col + start_col
+                if end_col > tonumber(start_col) then
+                    return end_col
+                else
+                    return default_position
+                end
+            end
+        end
```
jose-elias-alvarez, line 44:
```suggestion
        if not pattern then
            return default_position
        end

        local results = message:match(pattern)
        _, end_col = trimmed_line:find(results, 1, true)

        if not end_col then
            return default_position
        end

        end_col = end_col + start_col
        if end_col > tonumber(start_col) then
            return end_col
        end

        return default_position
```

I think using early returns and `not` vs `== nil` is a little more idiomatic and reduces nesting, which will help if this helper continues to grow to handle more codes.

jose-elias-alvarez (commented):
Other than the suggested refactor this looks good to me!

jose-elias-alvarez:
Sure, thank you!

PR #530, opened 2022-01-02
feat(builtin): Add `gitlint` diagnostic

tymcauley:
Adds `gitlint` diagnostic source, for checking git commit messages: https://jorisroovers.com/gitlint/

Unfortunately [`gitlint` doesn't support receiving data over stdin from sockets](https://github.com/jorisroovers/gitlint/blob/3c017995633f602125f7caaa91e96bb767ca5953/gitlint-core/gitlint/cli.py#L121-L122), so I had to use the temporary-file strategy to make it work.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1915,6 +1915,24 @@ local sources = { null_ls.builtins.diagnostics.flake8 }
 - `command = "flake8"`
 - `args = { "--stdin-display-name", "$FILENAME", "-" }`
 
+#### [gitlint](https://jorisroovers.com/gitlint/)
+
+##### About
+
+Linter for git commit messages
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.diagnostics.gitlint }
+```
+
+##### Defaults
+
+- `filetypes = { "gitcommit" }`
+- `command = "gitlint"`
+- `args = { "--msg-filename", "$FILENAME" }`
+
 #### [pylama](https://github.com/klen/pylama)
 
 ##### About
--- a/lua/null-ls/builtins/diagnostics/gitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/gitlint.lua
@@ -1,0 +1,27 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "gitlint",
+    method = DIAGNOSTICS,
+    filetypes = { "gitcommit" },
+    generator_opts = {
+        command = "gitlint",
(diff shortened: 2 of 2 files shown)


jose-elias-alvarez:
Thank you for the contribution! (This is an aside, but: A while back we had a bug that affected `gitcommit` buffers, and there was some discussion about whether we would ever want to support `gitcommit` sources. I didn't know this linter existed, but I'm glad we continued supporting them.)

tymcauley:
Thanks so much for quickly merging this! Yes, I actually tried to create this feature a few months ago, but then fell victim to [this check](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/514272831c0e58b4c1ed94d26faa69c81447a2a0/lua/null-ls/lspconfig.lua#L21-L25). Once 0.6.0 was released I was very happy to see that I could get back to this :)

David-Else:
This is an awesome feature, thanks!

At the moment if I open a terminal in Neovim and type `git commit` then I can use this linter BUT as it is a Neovim inside a Neovim it gets very confused, slows down and eventually freezes. I don't think this is the rightt way to do it, but I don't want to install any more plugins like vimfugative.

Is there a quick function I could use to open a `gitcommit` buffer and then commit on save? I tried looking at the vimfugative code, but it is gobbledygook to me :) At the moment I have a very simple function to commit, but I can't use `gitlint` then:

```vim
function GitCommit() abort
  let message = input('Enter commit message: ')
  call system("git commit -m '" . message . "'")
endfunction
```

jose-elias-alvarez:
@David-Else I'm not sure about the specific issue of running null-ls in nested terminals, but you could try using [neovim-remote](https://github.com/mhinz/neovim-remote) to avoid nesting in the first place.

tymcauley:
@David-Else it's also possible that using `gitlint` in a different way might work for you. The null-ls `gitlint` diagnostic helps for giving interactive feedback on your commit message while you write it in neovim, but you could also run `gitlint` from the command-line, or install it as [git hook](https://jorisroovers.com/gitlint/#using-gitlint-as-a-commit-msg-hook). If you're fine with amending your commit message after writing it, then you can always just run `gitlint` at the command line, and it will run on your most recent commit by default.

David-Else:
Thanks for the suggestions! I have looked at them all, but unfortunately none are ideal. 

I have decided the best way for now is to have Gnome terminal split into two tabs in the same repo directory:

Terminal tab 1 = Neovim editing files in the repo
Terminal tab 2 = Gnome terminal command prompt

...then just `git commit` from the 2nd pane and let a second instance of Neovim run, that works perfectly and only requires a few more keystrokes.

Added bonus...

Terminal tab 3 = lazygit (check that out if you have not had a go! https://github.com/jesseduffield/lazygit)

PR #532, opened 2022-01-03
feat(builtin): Add `textlint` diagnostic

yuys13:
Hi, thanks for the great software!
I tried to add textlint to null-ls, as I often use textlint when writing.
I'm afraid I wrote the same function in several places, but I couldn't think of a better way. Since the diagnostics are compatible with eslint, I wrote the same function as eslint.
Please review.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2326,6 +2326,24 @@ local sources = { null_ls.builtins.diagnostics.teal }
 - `command = "tl"`
 - `args = { "check", "$FILENAME" }`
 
+#### [textlint](https://github.com/textlint/textlint)
+
+##### About
+
+The pluggable linting tool for text and markdown.
+
+##### Usage
+
+```lua
+local sources = { null_ls.builtins.diagnostics.textlint }
+```
+
+##### Defaults
+
+- `filetypes = {}`
+- `command = "textlint"`
+- `args = { "-f", "json", "--stdin", "--stdin-filename", "$FILENAME" }`
+
 #### [vale](https://docs.errata.ai/vale/about)
 
 ##### About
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Thanks! If you wait until I've had a chance to review and merge #521, we might be able to combine this with ESLint (though it might still be best to keep them separated – I want to think about it a bit more).

jose-elias-alvarez:
We could theoretically merge it with this diff:

```diff
--- Result of linting is compatible for eslint
-local handle_eslint_output = function(params)
-    params.messages = params.output and params.output[1] and params.output[1].messages or {}
-    if params.err then
-        table.insert(params.messages, { message = params.err })
-    end
-
-    local parser = h.diagnostics.from_json({
-        attributes = {
-            severity = "severity",
-        },
-        severities = {
-            h.diagnostics.severities["warning"],
-            h.diagnostics.severities["error"],
-        },
-    })
-
-    return parser({ output = params.messages })
-end
-
 return h.make_builtin({
     name = "textlint",
     method = DIAGNOSTICS,
@@ -33,7 +13,7 @@ return h.make_builtin({
         args = { "-f", "json", "--stdin", "--stdin-filename", "$FILENAME" },
         format = "json_raw",
         check_exit_code = { 0, 1 },
-        on_output = handle_eslint_output,
+        on_output = require("null-ls").builtins.diagnostics.eslint._opts.on_output,
     },
     factory = h.generator_factory,
 })
```

I don't feel strongly about this, but if I had to choose, I'd say I prefer keeping it separate to make sure that future changes on one end don't break the other end. In that case, we could rename the local `handle_eslint_output` helper to `handle_output` and this would be ready to merge.

yuys13:
Thanks for reviewing!

> We could theoretically merge it with this diff:

I also understand that I can use `on_output` around in this case.

>  I'd say I prefer keeping it separate to make sure that future changes on one end don't break the other end. 

Agreed.

> In that case, we could rename the local handle_eslint_output helper to handle_output and this would be ready to merge.

I fixed it!

jose-elias-alvarez:
Thank you!

PR #537, opened 2022-01-03
feat: environment variable injection

### LINKED ISSUE #385: Environment Variables for Run Command
### Issues

### Feature description

Hello @jose-elias-alvarez,

Thank you for expanding this beautiful plugin and for your work on it. It does much more than I would imagine at the initial state of the plugin. 

I would like to kindly ask you that since `null-ls` is checking for whether the command is executable or not, how can I add environment variables to do the command. I have scanned through the code based and already tried the approach to append the environment variables at the beginning of the command but there is no avail since it checks with `vim.executable` to whether this given command is executable or not.

I have a case where I would like to have the environment variable `PRETTIERD_DEFAULT_CONFIG` for `prettierd` to have default configuration because some of the company repositories do not have a `prettierrc` file.

Is there any way that can be implemented as a parameter to the current formatter/linter/code-action provider etc. just like a `environment` entry in the configuration with `prettierd.with`.

### Help

### Implementation help

_No response_

cenk1cenk2:
closes #385

- Adds another option field to pass in environment variables as in dictionary form.

Currently merges operating system variables first to match the default behavior used in the `neovim` repository.

What I could not manage to do because of `null-ls` first checking the command to be executable is extending the environment variables while doing that to match the behavior in the `lsp-installer` breaking change commit [here](https://github.com/williamboman/nvim-lsp-installer/commit/095ab4eb6a02d5fd3ea4b782a0e868e2c65e4427) where it now uses extending the `PATH` variable without using the absolute path of the executable. If you have any advice on that, I guess I can further improve this by injecting environment variables to the initial check of the file being executable, so that this external plugin would also work well.

### DIFF
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -21,8 +21,8 @@ local TIMEOUT_EXIT_CODE = 7451
 local M = {}
 
 M.spawn = function(cmd, args, opts)
-    local handler, input, check_exit_code, timeout, on_stdout_end =
-        opts.handler, opts.input, opts.check_exit_code, opts.timeout, opts.on_stdout_end
+    local handler, input, check_exit_code, timeout, on_stdout_end, env =
+        opts.handler, opts.input, opts.check_exit_code, opts.timeout, opts.on_stdout_end, opts.env
 
     local output, error_output = "", ""
     local handle_stdout = function(err, chunk)
@@ -98,7 +98,11 @@ M.spawn = function(cmd, args, opts)
         done(exit_ok, code == TIMEOUT_EXIT_CODE)
     end
 
-    handle = uv.spawn(vim.fn.exepath(cmd), { args = args, stdio = stdio, cwd = opts.cwd or vim.fn.getcwd() }, on_close)
+    handle = uv.spawn(
+        vim.fn.exepath(cmd),
+        { args = args, env = env, stdio = stdio, cwd = opts.cwd or vim.fn.getcwd() },
+        on_close
+    )
 
     if timeout then
         timer = M.timer(timeout, nil, true, function()
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -16,6 +16,35 @@ local on_error = function(errkind, ...)
     error(string.format("[%q] has failed with: %q", errkind, table.concat(..., ", ")))
 end
 
+---@private
+--- Merges current process env with the given env and returns the result as
+--- a list of "k=v" strings.
+---
+--- <pre>
+--- Example:
+---
+---  in:    { PRODUCTION="false", PATH="/usr/bin/", PORT=123, HOST="0.0.0.0", }
+---  out:   { "PRODUCTION=false", "PATH=/usr/bin/", "PORT=123", "HOST=0.0.0.0", }
+--- </pre>
+---@param env (table) table of environment variable assignments
+---@returns (table) list of `"k=v"` strings
+local function env_merge(env)
+  if env == nil then
+    return env
+  end
+
+  -- Merge.
+  env = vim.tbl_extend('force', vim.fn.environ(), env)
+
+  local final_env = {}
+  for k,v in pairs(env) do
+    assert(type(k) == 'string', 'env must be a dict')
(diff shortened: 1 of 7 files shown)


lua/null-ls/loop.lua
```diff
@@ -16,13 +16,42 @@ local on_error = function(errkind, ...)
...
+end
+
 local TIMEOUT_EXIT_CODE = 7451
 
 local M = {}
 
 M.spawn = function(cmd, args, opts)
-    local handler, input, check_exit_code, timeout, on_stdout_end =
-        opts.handler, opts.input, opts.check_exit_code, opts.timeout, opts.on_stdout_end
+    local handler, input, check_exit_code, timeout, on_stdout_end, env =
```
jose-elias-alvarez, line 53: I prefer to do this conditionally, closer to `uv.spawn`:

```lua
if env then
   env = env_merge(env)
end
```

This would let us remove the `if env == nil` guard from `env_merge`, too.

lua/null-ls/loop.lua
```diff
@@ -16,13 +16,42 @@ local on_error = function(errkind, ...)
...
+--- </pre>
+---@param env (table) table of environment variable assignments
+---@returns (table) list of `"k=v"` strings
+local function env_merge(env)
+    if env == nil then
+        return env
+    end
+
+    -- Merge.
+    env = vim.tbl_extend("force", vim.fn.environ(), env)
```
jose-elias-alvarez, line 37: What do you think about using `uv.os_environ()` instead of the Vimscript function?

lua/null-ls/loop.lua
```diff
@@ -16,13 +16,42 @@ local on_error = function(errkind, ...)
...
+---
+--- <pre>
+--- Example:
+---
+---  in:    { PRODUCTION="false", PATH="/usr/bin/", PORT=123, HOST="0.0.0.0", }
+---  out:   { "PRODUCTION=false", "PATH=/usr/bin/", "PORT=123", "HOST=0.0.0.0", }
+--- </pre>
+---@param env (table) table of environment variable assignments
+---@returns (table) list of `"k=v"` strings
+local function env_merge(env)
```
jose-elias-alvarez, line 31: To make sure I understand why this is necessary, because I haven't messed with this before: `uv.spawn` defaults to the OS environment if `env = nil`. If `env` is non-`nil`, it uses `env` (and only `env`), meaning that other variables get discarded unless we manually merge them like this. Is that correct?

lua/null-ls/loop.lua
```diff
@@ -16,13 +16,42 @@ local on_error = function(errkind, ...)
...
+    if env == nil then
+        return env
+    end
+
+    -- Merge.
+    env = vim.tbl_extend("force", vim.fn.environ(), env)
+
+    local final_env = {}
+    for k, v in pairs(env) do
+        assert(type(k) == "string", "env must be a dict")
```
jose-elias-alvarez, line 41: Is this assertion necessary? Are there cases where the key wouldn't be a string?

jose-elias-alvarez (commented):
This looks great! I left a few comments but they are minor. We also need to add this to the helpers documentation and maybe mention it in the builtin documentation, too. 

As for merging environment variables before checking commands, at first glance it doesn't seem worth the complexity given that it's already fairly easy to modify commands.  and I wouldn't be opposed to adding support in the future if there's demand, since we'd already have access to the `env` table in the correct place, but I think it's okay as is.

lua/null-ls/loop.lua
cenk1cenk2: If I did not miss anything, the default behavior is like that.

lua/null-ls/loop.lua
cenk1cenk2: Added and additional check if checking a whether the table is empty or not.

lua/null-ls/loop.lua
jose-elias-alvarez: Got it, thanks for confirming.

jose-elias-alvarez:
LGTM, thanks for working on this!

cenk1cenk2:
I thank you for your time and for maintaining this beautiful plugin.

Have a good day!

PR #546, opened 2022-01-05
Mention --cache flag in ESLint documentation

sondr3:
This is a nice way to increase the performance of `eslint` as it will only run for the files that has changed (https://eslint.org/docs/user-guide/command-line-interface#--cache). It may however create surprises since this will create a `.eslintcache` file in the project root, so it may be better to add this as a flag instead of as a default? It is however in my mind a good default for the increased performance, but your mileage may vary.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -649,6 +649,8 @@ Fixes problems in your JavaScript code.
 
 - Slow and not suitable for formatting on save. If at all possible, use
   `eslint_d` (described below).
+- It's recommended to add the [`--cache`](https://eslint.org/docs/user-guide/command-line-interface#--cache)
+  flag as an extra argument, see [Expansion](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#arguments).
 
 ##### Usage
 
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
This is interesting, but I agree that it is a bit surprising as a default and would require users to update version control ignore files, which may not be doable in all environments. I also don't think it's necessary for `eslint_d`.

We could theoretically set `--cache-location` to either a temp dir (though that would require special handling for Windows) or something like `vim.fn.stdpath("data")`. My personal preference would be to mention this in the documentation as a recommended customization, but I agree that the default `eslint` experience is painfully slow and would support either one of the approaches.

sondr3:
It's not really explained in the documentation when the `--cache` flag is useful, though you may be right that it doesn't really matter since `null-ls` pipes via `stdin`. I found [this](https://github.com/eslint/eslint/issues/13586) issue and a linked RFC that might help. The docs in `eslint_d` do mention using the `--cache` flag at the botton of the [moar speed](https://github.com/mantoni/eslint_d.js#moar-speed) section. :shrug:

jose-elias-alvarez:
I think the linked issue makes it clear that this won't make a difference for our use case, so I'd prefer not to mention it at all to avoid giving users the expectation that this will help. In the future, if we add a project-wide ESLint diagnostics built-in like like [the `tsc` built-in](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/tsc.lua), this flag would make a big difference, but for now I don't think it's relevant.

PR #551, opened 2022-01-06
fix(rubocop): extract column info from location

aergonaut:
The `start_column` and `last_column` properties are under the `location` property in RuboCop's JSON output, not at the top-level. Previously the builtin for RuboCop was extracting these from the wrong location, so diagnostics were not annotated with the correct begin and end points.

This fixes the builtin so that it sources `start_column` and `last_column` from the correct place.

I have tested this locally by creating a custom source for RuboCop with this change, and the diagnostics have the correct start and end columns.

**Before**

Screen Shot 2022-01-06 at 09 13 45

**After**

Screen Shot 2022-01-06 at 09 12 27


jose-elias-alvarez:
Great, thank you!

PR #562, opened 2022-01-08
fix(logger): respect config.log.enable

sentriz:
hi! it seems the `config.log.enable` option isn't working for me, and I can't find any other references to it

image


jose-elias-alvarez:
Nice catch, thank you!

Issue #563, opened 2022-01-09
Add support for cabal-fmt formatting


ynishi:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Support for cabal-fmt.

cabal-fmt is a formatter tool for cabal haskell package manager.
https://github.com/phadej/cabal-fmt

I setup and works good for me.

### Help

No

PR #564, opened 2022-01-09
feat: cabal-fmt formatting source for cabal files

ynishi:
Resolved: #563 

* "cabal-fmt" is original command name, I changed it to underscore for lib.


jose-elias-alvarez:
Thank you!

PR #568, opened 2022-01-10
Fix: Update qmllint parameters and args

Furkanzmc:
The latest qmllint now uses a configuration file so the users can just use that file to configure the settings. For earlier versions, I think using `with({})` to change how qmllint is run still an option.


jose-elias-alvarez:
So to clarify, we were previously hardcoding the `--no-unqualified-id` rule but are now leaving that up to the config file, right? In that case I think this should be okay, since it's pretty simple to restore the previous behavior.

Furkanzmc:
That's correct.

jose-elias-alvarez:
Great, thanks for confirming and for the PR!

PR #570, opened 2022-01-10
feat: Add builtin for [Alex](https://alexjs.com/)

IamTheFij:
I added a test, but I can't tell if the test is passing. This works when I take the same diagnostic and use it locally, but even if I make the test try to fail intentionally, it still give the same number of `Failed` tests.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/alex.lua
+++ b/lua/null-ls/builtins/diagnostics/alex.lua
@@ -1,0 +1,26 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "alex",
+    method = DIAGNOSTICS,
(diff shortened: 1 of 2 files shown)


IamTheFij:
Incidentally, I noticed you asking for things to be formatted with StyLua and Prettier. If you'd be interested, I can make a PR for running these as pre-commit hooks.

IamTheFij:
Oh, it looks like there is a tool called `Commitizen` which lints commit messages such that they must follow Conventional Commits, but it looks like it may be too strict. eg `(build|ci|docs|feat|fix|perf|refactor|style|test|chore|revert|bump)(\(\S+\))?!?:(\s.*)` would not accommodate the `doc: <summary>` commits that I see in the log.

jose-elias-alvarez:
I went ahead and fixed the test (I think `message` was missing a space, and the positions needed to be strings). Other than that, this looks good. Thank you for the contribution (and for introducing me to a very interesting linter)! I'll check out your other PR, too.

PR #571, opened 2022-01-10
Pre commit

IamTheFij:
Moving pre-commit suggestions to here from https://github.com/jose-elias-alvarez/null-ls.nvim/pull/570#issuecomment-1009412555

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -2,6 +2,7 @@ name: Feature request
 description: Request an enhancement for null-ls
 labels: [enhancement]
 body:
+
   - type: checkboxes
     id: issue-prerequisite
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -2,7 +2,6 @@ name: Feature request
 description: Request an enhancement for null-ls
 labels: [enhancement]
 body:
-
   - type: checkboxes
     id: issue-prerequisite
     attributes:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,11 @@
+---
+repos:
+  - repo: https://github.com/johnnymorganz/stylua
+    # TODO: Use v0.11.4 when released with support for stylua-github
+    rev: master
+    hooks:
+      - id: stylua-github
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.5.1
+    hooks:
+      - id: prettier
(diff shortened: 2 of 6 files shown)


jose-elias-alvarez:
I'm not familiar with this and am not necessarily opposed to it, but is this a common thing for open-source projects? We currently have a CI job to check files with StyLua, which has been working well. It seems to me that the contributors who read and follow instructions are unlikely to ignore a CI failure, so asking for this seems a bit onerous (especially since I imagine many of those contributors are already using StyLua via null-ls).

If we do proceed with this, we'd have to tell Prettier to ignore files in the `test/` directory, since those files are deliberately unformatted and are used in integration tests.

IamTheFij:
A pretty good amount do. One popular one I interact with is Home Assistant. 

It’s a good point though that many contributors will probably clone using the project itself to handle StyLua, and the CI job will prevent unformatted merges. 

Pushing it as a pre-commit hook really just gives a quicker feedback loop.

Given the amount of opportunities for checks that do exist, I can update the wording in the docs to describe it as more of an optional tool for the developers rather than a required step when contributing. Though I’m also totally fine with abandoning this branch. 

I just added Pre Commit support to StyLua myself a few days ago and added it to my own configs and figured it wouldn’t take me long to replicate here.

jose-elias-alvarez:
Sure, I'd be happy to merge this if we can frame it as a suggestion to improve the contributor experience versus a requirement (and if we can solve the Prettier issues I mentioned).

IamTheFij:
Sounds good! I just made the changes. I broke everything out to distinct commits, but this could probably just squash out on merge.

syphar:
One note about this change: 

more advantage of pre-commit can come from calling it directly in `ci.yml` too instead of the linters directly. This merges the linting config for both into just `.pre-commit-config.yaml`. 

Alternatively, calling just `make lint` in this case

jose-elias-alvarez:
I tried this out locally and it's great (I'm especially happy that it plays nice with `lazygit`). 

> One note about this change:
> 
> more advantage of pre-commit can come from calling it directly in `ci.yml` too instead of the linters directly. This merges the linting config for both into just `.pre-commit-config.yaml`.
> 
> Alternatively, calling just `make lint` in this case

Can you elaborate on what you mean by this?

IamTheFij:
I believe that they mean that instead of running the `StyLua` action, you could run `make check` in a shell or the `pre-commit/action` and it will inherit the configuration from `.pre-commit.config.yaml`.

jose-elias-alvarez:
> I believe that they mean that instead of running the `StyLua` action, you could run `make check` in a shell or the `pre-commit/action` and it will inherit the configuration from `.pre-commit.config.yaml`.

Got it. We can keep improving this as we go, I guess, but I think it's a nice addition for those who want to opt in. Thanks!

PR #575, opened 2022-01-11
fix(diagnostics): Fix ansible-lint parsing

ViViDboarder:
There was an issue with parsing severities containing `_`.

This also adds parsing of the error code as well as a test.

Unfortunately, I can't seem to figure out why the test fails.

It works correctly when I run it on an actual file and the match pattern seems to work when executed on the test output.

```sh
echo 'print(string.match([[path/to/file.yaml:2: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect]], [[^[^:]+:(%d+): %[([%w-]+)%] %[([%w_]+)%] (.*)$]]))' | lua
2	risky-file-permissions	VERY_HIGH	File permissions unset or incorrect
```

But when I run the test I get
```
Fail	||	diagnostics ansiblelint should create a diagnostic
            ./test/spec/builtins/diagnostics_spec.lua:804: Expected objects to be the same.
            Passed in:
            (nil)
            Expected:
            (table: 0x0102e82a10) {
              [code] = 'risky-file-permissions'
              [message] = 'File permissions unset or incorrect'
              [row] = '2'
              [severity] = 1 }

            stack traceback:
            	./test/spec/builtins/diagnostics_spec.lua:804: in function <./test/spec/builtins/diagnostics_spec.lua:801>

```

### DIFF
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -910,13 +910,18 @@ describe("diagnostics", function()
                   }
                 ]
             ]]
-            local diagnostic = parser(output, { content = file })
-            assert.same({
-                row = "5",
-                severity = 1,
-                message = "[risky-file-permissions] File permissions unset or incorrect",
-                filename = "playbooks/test-ansible.yaml",
-            }, diagnostic)
+            local diagnostic = parser({ output = vim.json.decode(output), content = file })
+            assert.same(
+                {
+                    {
+                        row = 5,
+                        severity = 1,
+                        message = "[risky-file-permissions] File permissions unset or incorrect",
+                        filename = "playbooks/test-ansible.yaml",
+                    },
+                },
+                diagnostic
+            )
         end)
     end)
 end)
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -869,4 +869,30 @@ describe("diagnostics", function()
             }, diagnostic)
         end)
     end)
+
+    describe("ansiblelint", function()
+        local linter = diagnostics.ansiblelint
+        local parser = linter._opts.on_output
+        local file = {
+            [[---]],
+            [[- name: generate file]],
+            [[  assemble:]],
+            [[    src: "files"]],
+            [[    dest: "dest"]],
+            [[  run_once: true]],
+            [[  become: false]],
+        }
+
+        it("should create a diagnostic", function()
+            local output =
+                [[path/to/file.yaml:2: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect]]
+            local diagnostic = parser(output, { content = file })
+            assert.same({
+                row = "2",
+                severity = 1,
+                message = "File permissions unset or incorrect",
+                code = "risky-file-permissions",
+            }, diagnostic)
+        end)
+    end)
 end)
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -875,23 +875,48 @@ describe("diagnostics", function()
         local parser = linter._opts.on_output
         local file = {
             [[---]],
-            [[- name: generate file]],
-            [[  assemble:]],
-            [[    src: "files"]],
-            [[    dest: "dest"]],
-            [[  run_once: true]],
-            [[  become: false]],
+            [[- name: null-ls]],
+            [[  hosts: all]],
+            [[  tasks:]],
+            [[    - name: This tasks is no good]],
+            [[      assemble:]],
+            [[        src: "files"]],
+            [[        dest: "dest"]],
+            [[      become: false]],
         }
 
         it("should create a diagnostic", function()
-            local output =
-                [[path/to/file.yaml:2: [risky-file-permissions] [VERY_HIGH] File permissions unset or incorrect]]
+            local output = [[
+                [
+                  {
+                    "type": "issue",
+                    "check_name": "[risky-file-permissions] File permissions unset or incorrect",
+                    "categories": [
+                      "unpredictability",
+                      "experimental"
+                    ],
+                    "severity": "blocker",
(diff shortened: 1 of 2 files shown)


ViViDboarder:
Oh, I just realized that there is a JSON output option using `-f codeclimate`. It will be better to replace with that.

Additionally, it appears that even if you pass a single file into `ansible-lint` it will return errors present in roles that are referenced by the file. It would be good to find a way to filter out message for files that don't match the buffer. That is unless there is some way to handle diagnostics across multiple files.

jose-elias-alvarez:
I checked out the PR locally and the test seems to be working fine – did you update it? And yes, we should definitely prefer using JSON if it's an option. 

It's possible for a single source to return diagnostics for multiple files using [the `multiple_files` option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/MAIN.md#multi-file-generators), but given the arguments we use to spawn the linter, doing so might be surprising, so it may be best to filter the diagnostics by using a custom `on_output` (something like what we use for [ESLint](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/eslint.lua)).

ViViDboarder:
Now using JSON, a custom `on_output`, and `multiple_files`. Still, for some reason the test doesn't seem to want to pass for me though.

```
$ env FILE=./test/spec/builtins/diagnostics_spec.lua make test-file
nvim --headless --noplugin -u test/minimal.vim -c "lua require(\"plenary.busted\").run(\"./test/spec/builtins/diagnostics_spec.lua\")"

========================================
Testing: 	./test/spec/builtins/diagnostics_spec.lua
Success	||	diagnostics chktex should create a diagnostic
Success	||	diagnostics credo should create a diagnostic with error severity
Success	||	diagnostics credo should create a diagnostic with warning severity
Success	||	diagnostics credo should create a diagnostic with information severity
Success	||	diagnostics credo should create a diagnostic falling back to hint severity
Success	||	diagnostics credo returns errors as diagnostics
Success	||	diagnostics credo should handle compile warnings preceeding output
Success	||	diagnostics credo should handle messages with incomplete json
Success	||	diagnostics credo should handle messages without json
Success	||	diagnostics luacheck should create a diagnostic
Success	||	diagnostics write-good should create a diagnostic
Success	||	diagnostics markdownlint should create a diagnostic with a column
Success	||	diagnostics markdownlint should create a diagnostic without a column
Success	||	diagnostics mdl should create a diagnostic
Success	||	diagnostics tl check should create a diagnostic (quote field is between quotes)
Success	||	diagnostics tl check should create a diagnostic (quote field is not between quotes)
Success	||	diagnostics tl check should create a diagnostic by using the second pattern
Success	||	diagnostics shellcheck should create a diagnostic with info severity
Success	||	diagnostics shellcheck should create a diagnostic with style severity
Success	||	diagnostics selene should create a diagnostic (quote is between backquotes)
Success	||	diagnostics selene should create a diagnostic (quote is not between backquotes)
Success	||	diagnostics eslint should create a diagnostic with warning severity
Success	||	diagnostics eslint should create a diagnostic with error severity
Success	||	diagnostics standardjs should create a diagnostic with error severity
Success	||	diagnostics standardjs should create a diagnostic with warning severity
Success	||	diagnostics hadolint should create a diagnostic with warning severity
Success	||	diagnostics hadolint should create a diagnostic with info severity
Success	||	diagnostics flake8 should create a diagnostic
Success	||	diagnostics misspell should create a diagnostic
Success	||	diagnostics vint should create a diagnostic with warning severity
Success	||	diagnostics yamllint should create a diagnostic with warning severity
Success	||	diagnostics jsonlint should create a diagnostic
Success	||	diagnostics cue_fmt should create a diagnostic
Fail	||	diagnostics ansiblelint should create a diagnostic
            ./test/spec/builtins/diagnostics_spec.lua:829: Expected objects to be the same.
            Passed in:
            (nil)
            Expected:
            (table: 0x01033e94f8) {
              [filename] = 'playbooks/test-ansible.yaml'
              [message] = '[risky-file-permissions] File permissions unset or incorrect'
              [row] = '5'
              [severity] = 1 }

            stack traceback:
            	./test/spec/builtins/diagnostics_spec.lua:829: in function <./test/spec/builtins/diagnostics_spec.lua:803>


Success: 	33
Failed : 	1
Errors : 	1
========================================
Tests Failed. Exit: 1
make: *** [test-file] Error 1
```

test/spec/builtins/diagnostics_spec.lua
```diff
@@ -784,4 +784,54 @@ describe("diagnostics", function()
...
+                        "begin": 5
+                      }
+                    },
+                    "content": {
+                      "body": "Task/Handler: This tasks is no good"
+                    }
+                  }
+                ]
+            ]]
+            local diagnostic = parser(output, { content = file })
```
jose-elias-alvarez, line 828:
```suggestion
            local diagnostic = parser({ output = vim.json.encode(output), content = file })
```

The issue here is that we changed to a JSON parser, so it expects a single `params` table where `output` is a decoded JSON object.

test/spec/builtins/diagnostics_spec.lua
```diff
@@ -784,4 +784,54 @@ describe("diagnostics", function()
...
+            assert.same({
```
jose-elias-alvarez, line 829: Likewise, sources that use the JSON format are expected to return a table of diagnostics, so the diagnostic should be nested.

jose-elias-alvarez (commented):
I tried this out locally, fixed the tests, and tried to push them here, but I guess they got lost in the void. I left comments on how to fix them, and I think this also needs a rebase now. Other than that, I think this is okay!

test/spec/builtins/diagnostics_spec.lua
ViViDboarder: Thanks!

jose-elias-alvarez:
I think this has to be rebased to fix conflicts in the test file. Other than that I think this is ready to merge.

ViViDboarder:
Rebased. There is an issue with the test for `jsonlint` that is unrelated to the rebase.

jose-elias-alvarez:
Alright, I'll check that out separately. Thanks again!

PR #577, opened 2022-01-12
fix: Fix multifile parsing for mypy

IamTheFij:
Similar to what was happening on #575. `mypy` returns errors found in files referenced by the current one. This adds the filename to the diagnostics result so that they are properly handled.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -29,9 +29,10 @@ return h.make_builtin({
         check_exit_code = function(code)
             return code <= 2
         end,
+        multiple_files = true,
         on_output = h.diagnostics.from_pattern(
-            "[^:]+:(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
-            { "row", "col", "severity", "message", "code" },
+            "([^:])+:(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
+            { "filename", "row", "col", "severity", "message", "code" },
             {
                 severities = {
                     error = h.diagnostics.severities["error"],
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -31,7 +31,7 @@ return h.make_builtin({
         end,
         multiple_files = true,
         on_output = h.diagnostics.from_pattern(
-            "([^:])+:(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
+            "([^:]+):(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
             { "filename", "row", "col", "severity", "message", "code" },
             {
                 severities = {


jose-elias-alvarez:
I won't have a chance to review this or your PRs for a bit but everything looks good at a glance. If possible, could you tell me how I can set up 2 Python files to test out the difference in behavior? The fact that it outputs diagnostics for multiple files at once even when we specify a path seems pretty unusual, so I'd just like to see what the experience is like.

IamTheFij:
It was surprising to me too. Here's a working example: https://github.com/IamTheFij/mypy_multifile_test

jose-elias-alvarez:
I see what you mean, and it plays nicely with `stdin`, so it's definitely a good idea to support this. 

On my end, though, the pattern isn't matching the output, so while I can see the correct output in the debug log, the returned diagnostics are empty. I'm not a regular Python / `mypy` user, and I'm using version 0.931 without any extra config, so is there a reason why this wouldn't work for me?

IamTheFij:
Huh. That’s strange. It “works for me”. I didn’t change the pattern either and only added a capture group for the file name rather than dropping it.

lua/null-ls/builtins/diagnostics/mypy.lua
```diff
@@ -29,9 +29,10 @@ return h.make_builtin({
         check_exit_code = function(code)
             return code <= 2
         end,
+        multiple_files = true,
         on_output = h.diagnostics.from_pattern(
-            "[^:]+:(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
-            { "row", "col", "severity", "message", "code" },
+            "([^:])+:(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
```
jose-elias-alvarez, line 34:
```suggestion
            "([^:]+):(%d+):(%d+): (%a+): (.*)  %[([%a-]+)%]", --
```

It's strange that this works on your end and not on mine, but this change makes it work for me.

jose-elias-alvarez (commented):
So the suggested change seems to fix the pattern issue, but while testing this I ran into a potential issue. In your sample repo, `file.py` is aware of the error in `config.py`, so opening `file.py` correctly generates errors for both files. However, `config.py` is not aware of the fact that it's used in `file.py`, so opening `config.py` wipes out errors in `file.py`.

This is because of how `multiple_files` works: basically, since the source will (almost) never explicitly send *empty* diagnostics for a given file, we have to assume that if the diagnostics aren't there, we should clear existing diagnostics to prevent staleness. 

To me, this seems a bit inconsistent, but overall this is still an improvement over the current behavior. Let me know what you think about this as a user. If you don't think the case I described above is a big deal, I'm happy to merge this.

lua/null-ls/builtins/diagnostics/mypy.lua
IamTheFij: Huh. Yea, that should change. Strange that it worked for me... Changed. Maybe I should add a test for mypy.

jose-elias-alvarez:
Great, thanks. Let me know what you think about the behavior I described in my comment, since I think it affects your other PR, too.

IamTheFij:
Ah. Yea, I think that it’s acceptable behavior and one that I can’t think of a great way to work around. It would be possible to filter results for only the target buffer, but that doesn’t really seem better because you’re definitively dropping messages. 

Not sure what the “best” solution would be. Maybe once I use it more day to day I’ll form an opinion and open a new PR or issue discussion. 

Thanks for creating this and for the help getting these changes merged! I think the API you built is great and makes adapting super easy.

jose-elias-alvarez:
Thanks for the feedback and for the kind words! The only thing I can think of for now some kind of other flag to tell null-ls *not* to wipe out diagnostics if they are empty, but that seems like overkill. Let's see if we can come up with something better, but for now I'm happy with this.

Issue #572, opened 2022-01-11
vim.lsp.buf.formatting_sync: timeout When trying to use prettier.


hrqmonteiro:
Simply doesn't work.

![print](https://user-images.githubusercontent.com/17055027/148868478-3f73b3cf-4887-4ac7-957a-d9bc1a8e8db1.png)
![print2](https://user-images.githubusercontent.com/17055027/148868480-83bdef57-f2b4-4f85-be23-55deb1ee916f.png)
![print3](https://user-images.githubusercontent.com/17055027/148868482-06717eb0-e56e-45f3-be9f-2d3729cf3e83.png)

jose-elias-alvarez:
I added the issue template for a reason and will not provide support to users who can't be bothered to follow it.

iDuuck:
@hrqmonteiro This seems to be an issue with the timeout neovim has implemented for the formatting_sync method. Try increasing it: `vim.lsp.buf.formatting_sync(nil, 2000)` (second param is time in ms)

PR #579, opened 2022-01-12
Add faq for timeout error message to FAQ section in README.md

iDuuck:
Having this problem my self, and seeing this message pop up in the issues (#572) of this project, I decided to create this little FAQ in the FAQs section.

Problem here: In my case, `rubocop` took longer than the default timeout, so I decided to drop the formatting for rubocop, since it takes to long. Other solution would be to just increase the timeout for any `formatting_sync` call, however this slowed down my editor performance massively, so I just dropped formatting support.


jose-elias-alvarez:
Thanks! I agree that adding a snippet explicitly showing how to increase the timeout is a good idea.

iDuuck:
Will add this asap, thanks for the reviews.

iDuuck:
Updated!

jose-elias-alvarez:
Thank you!

PR #583, opened 2022-01-14
feat(builtin): Add protolint diagnostic

Andersbiha:
Adds [protolint](https://github.com/yoheimuta/protolint) diagnostic for `.proto` files.

Since [protolint](https://github.com/yoheimuta/protolint) doesn't support input from `stdin` a temporary file is used instead.

This was initially implemented with `from_json`  but had to be reworked since [protolint](https://github.com/yoheimuta/protolint)  spits out `plaintext` instead of `json` when its parsing invalid `.proto` files.
The final implementation is based on the  [credo](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/credo.lua) diagnostic.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2189,6 +2189,24 @@ local sources = { null_ls.builtins.diagnostics.proselint }
 - `command = "proselint"`
 - `args = { "--json" }`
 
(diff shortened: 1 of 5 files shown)


jose-elias-alvarez:
Thank you for the excellent PR!

PR #591, opened 2022-01-17
feat(builtint): Add protoc-gen-lint diagnostic

Andersbiha:
Adds [protoc-gen-lint](https://github.com/ckaznocha/protoc-gen-lint) diagnostic for `.proto` files.

Since [protoc-gen-lint](https://github.com/ckaznocha/protoc-gen-lint) doesn't support input from `stdin` a temporary file is used instead.

This was initially implemented with `from_pattern` but had to be reworked with an additional matching pattern since [protoc-gen-lint](https://github.com/ckaznocha/protoc-gen-lint) spits out errors in a different format when its parsing invalid `.proto` files.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/protoc_gen_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/protoc_gen_lint.lua
@@ -1,0 +1,31 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "protoc-gen-lint",
+    method = DIAGNOSTICS,
+    filetypes = { "proto" },
+    generator_opts = {
+        command = "protoc",
+        args = { "--lint_out", "$FILENAME", "-I", "/tmp", "$FILENAME" },
+        from_stderr = true,
+        to_temp_file = true,
+        format = "line",
+        check_exit_code = function(code)
+            return code <= 1
+        end,
+        on_output = h.diagnostics.from_patterns({
+            {
+                pattern = [[:(%d+):(%d+): (.*)]],
+                groups = { "row", "col", "message" },
+            },
+            {
+                pattern = [[.*] (.*)]],
+                groups = { "message" },
+            },
+        }),
+    },
+    factory = h.generator_factory,
+})
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2189,6 +2189,24 @@ local sources = { null_ls.builtins.diagnostics.proselint }
 - `command = "proselint"`
 - `args = { "--json" }`
 
+#### [protoc-gen-lint](https://github.com/ckaznocha/protoc-gen-lint)
+
+##### About
+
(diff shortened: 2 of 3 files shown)


lua/null-ls/builtins/diagnostics/protoc_gen_lint.lua
```diff
@@ -0,0 +1,50 @@
...
+        on_output = function(line)
+            local defaultErrorPattern = [[:(%d+):(%d+): (.*)]]
+            local genericErrorPattern = [[.*] (.*)]]
+
+            local match = { line:match(defaultErrorPattern) }
+
+            local issue = {
+                row = 1,
+                col = 0,
+                severity = h.diagnostics.severities.error,
```
jose-elias-alvarez, line 28: If we're not capturing severity, it's best to leave it undefined so that users can use [`fallback_severity`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#fallback_severity-number).

lua/null-ls/builtins/diagnostics/protoc_gen_lint.lua
```diff
@@ -0,0 +1,50 @@
...
+        command = "protoc",
+        args = { "--lint_out", "$FILENAME", "-I", "/tmp", "$FILENAME" },
+        from_stderr = true,
+        to_temp_file = true,
+        format = "line",
+        check_exit_code = function(code)
+            return code <= 1
+        end,
+        on_output = function(line)
+            local defaultErrorPattern = [[:(%d+):(%d+): (.*)]]
```
jose-elias-alvarez, line 20: We also have a [`from_patterns` helper](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/7558ea51d96e7be04bd61d1cc3d3ed892c4b42c3/lua/null-ls/helpers/diagnostics.lua#L151) that will attempt to match against multiple patterns – would that help here?

lua/null-ls/builtins/diagnostics/protoc_gen_lint.lua
```diff
@@ -0,0 +1,50 @@
...
+            local genericErrorPattern = [[.*] (.*)]]
```
jose-elias-alvarez, line 21: Also, while we're making changes, we might as well change this to snake case over camel case, just for consistency.

jose-elias-alvarez (commented):
Thank you for the PR! I left some comments. Let me know if `from_patterns` would be of help here, but if not I think this is okay.

Andersbiha:
I somehow overlooked `from_patterns`. That was exactly what i needed.

Andersbiha:
I removed the changes to the `_meta` files instead of rebasing so the CI can do it's job.

Adding a note would be nice but some comments in the `Makefile` would suffice to begin with.
As as newcomer to both `nvim` and `lua` this was one of the first files i looked at.

Thanks for the a great project as well as your quick responses and merges :+1:  :1st_place_medal:

jose-elias-alvarez:
Sure, I'll add something into the `Makefile`, too. Thanks again for the contributions!

PR #592, opened 2022-01-17
feat(builtin): Add protolint formatter

Andersbiha:
Adds [protolint](https://github.com/yoheimuta/protolint) formatter for `.proto` files.


jose-elias-alvarez:
Thanks! This isn't a big deal but just FYI we have a CI job to update the metadata files, so there's no need to do so manually (though it doesn't hurt).

PR #597, opened 2022-01-18
feat(builtins): set -srcdir in goimports

sentriz:
sometimes goimports has a hard time automatically importing packages that are in your current project, and imports some random 3rd party one with the same name. makes sense since all it's reading from is stdin

from `goimports -h`

```
usage: goimports [flags] [path ...]
...
  -srcdir dir
    	choose imports as if source code is from dir. When operating on a single file, dir may instead be the complete file name.
...
```

I have been using this for a while and it is also making saves go from like 1s+ to milliseconds 

also @abzcoding  it seems you added the original goimports builtin, can you see any drawbacks here? thanks!

### DIFF
--- a/lua/null-ls/builtins/formatting/goimports.lua
+++ b/lua/null-ls/builtins/formatting/goimports.lua
@@ -9,6 +9,7 @@ return h.make_builtin({
     filetypes = { "go" },
     generator_opts = {
         command = "goimports",
+        args = { "-srcdir", "$DIRNAME" },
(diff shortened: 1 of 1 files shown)


abzcoding:
seems good to me 👍

abzcoding (approved):

jose-elias-alvarez:
Excellent, thank you @sentriz for the PR and @abzcoding for the review!

PR #608, opened 2022-01-24
fix: corrected format on save behaviour for fnlfmt and raco fmt

astridlyre:
Because `fnlfmt` and `raco fmt` were formatting with `$FILENAME` in their args, it led to unsaved changes not being reflected in formatting on save. Using stdin seems to solve this.


jose-elias-alvarez:
Interesting, I don't see issues when running `fnlfmt` on save, but I think this is a sensible change. Thanks!

PR #612, opened 2022-01-25
feat(formatting): styler support rmd files

aspeddro:
Currently the `styler` formatter does not format `rmd` files. The `styler::style_text()` function does not support the content of Rmarkdown files.

This PR add support for `rmd` files using `styler::style_file()` function.

### DIFF
--- a/lua/null-ls/builtins/formatting/styler.lua
+++ b/lua/null-ls/builtins/formatting/styler.lua
@@ -9,13 +9,32 @@ return h.make_builtin({
     filetypes = { "r", "rmd" },
     generator_opts = {
         command = "R",
-        args = h.range_formatting_args_factory({
-            "--slave",
-            "--no-restore",
-            "--no-save",
-            "-e",
-            'con=file("stdin");output=styler::style_text(readLines(con));close(con);print(output, colored=FALSE)',
-        }),
+        args = function(params)
+            local default_args = {
+                "--slave",
+                "--no-restore",
+                "--no-save",
+                "-e",
+            }
+            if params.ft == "r" then
+                return vim.list_extend(default_args, {
+                    [[con=file("stdin");output=styler::style_text(readLines(con));close(con);print(output, colored=FALSE)]],
+                })
+            end
+            return vim.list_extend(default_args, {
+                string.format(
+                    [[options(styler.quiet = TRUE)
+                          con = file("stdin")
+                          temp = tempfile("styler",fileext = ".%s")
+                          writeLines(readLines(con), temp)
+                          styler::style_file(temp)
+                          cat(paste0(readLines(temp), collapse = '\n'))
+                          close(con)
+                        ]],
+                    params.ft
+                ),
+            })
+        end,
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/formatting/styler.lua
```diff
@@ -9,13 +9,32 @@ return h.make_builtin({
...
-            'con=file("stdin");output=styler::style_text(readLines(con));close(con);print(output, colored=FALSE)',
-        }),
+        args = function(params)
+            local default_args = {
+                "--slave",
+                "--no-restore",
+                "--no-save",
+                "-e",
+            }
+            if params.ft ~= "r" then
```
jose-elias-alvarez, line 19: This is pretty minor, but I think it's a little cleaner if the special case returns early (e.g. `if params.ft == "r" then`).

lua/null-ls/builtins/formatting/styler.lua
```diff
@@ -9,13 +9,32 @@ return h.make_builtin({
...
     generator_opts = {
         command = "R",
-        args = h.range_formatting_args_factory({
-            "--slave",
-            "--no-restore",
-            "--no-save",
-            "-e",
-            'con=file("stdin");output=styler::style_text(readLines(con));close(con);print(output, colored=FALSE)',
-        }),
+        args = function(params)
```
jose-elias-alvarez, line 12: It looks like the current version of the formatter supports range formatting, but this change loses that functionality. Could you take a look at `range_formatting_args_factory` and see how we can maintain it?

aspeddro: I've tested range formatting and it doesn't work. The only registered method is `methods.internal.FORMATTING`.

jose-elias-alvarez: I see, you're right that it doesn't seem to work in its current state. I know nothing about R or its ecosystem, but could you verify whether the CLI supports defining a specific range? As far as I can tell from [the documentation](https://styler.r-lib.org/dev/reference/index.html), it does not, so the proposed change should be okay.

aspeddro: To perform range formatting, I need the range to quote the content and pass it through the `styler::style_text()` function, the input is a string.

I did some testing, however the range doesn't seem to match the true position in visual mode.

`params.row` != `params.range.row`

image

jose-elias-alvarez: That's pretty strange–we actually [get the range from Neovim itself](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/utils.lua#L127-L129) and simply convert it to 1-indexing. We can look into that separately, but for this PR let's just go ahead and drop range formatting support, since it didn't work to begin with.

jose-elias-alvarez:
This looks good now. Could you just update the documentation?

aspeddro:
Done

jose-elias-alvarez:
Thanks!

Issue #502, opened 2021-12-27
Disable null-ls for specific buffers


jose-elias-alvarez:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

null-ls should make it easy to disable functionality for a specific buffer.

### Help

Yes

### Implementation help

The simplest solution I can think of is to use a buffer-local variable, something like`b:null_ls_disabled`, and check for it in [`client.supports_method`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/client.lua#L29) and [the diagnostics handler](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/diagnostics.lua#L119) to determine whether null-ls should do anything with that buffer. 

This is pretty simple – I'm opening the issue mostly as a reminder to myself, but if anyone happens to see this and wants to take care of it, feel free.

ELLIOTTCABLE:
FWIW, I'd really like to disable *specific sources* for a given buffer. I'm not sure if that merits a separate Issue — in particular, for me, `eslint_d` / `eslint` are often unhelpful in certain circumstances.

jose-elias-alvarez:
In the future, I'd like to restructure null-ls so that sources "attach" to a given buffer instead of being active all the time, which should make that a lot easier, but it's a pretty big change and it'll take some time for me to get around to it. In the meantime, you can toggle the availability of any source on demand using a command / function:

```lua
vim.cmd("command! ToggleEslint lua require('null-ls.sources').toggle('eslint')")
```

jose-elias-alvarez:
I thought about this more and am not sure I see the use case for doing this vs. disabling sources, which will correctly handle things like clearing diagnostics.

MattSPalmer:
Don't know if you'd consider re-opening this, but I have a use case that isn't really covered by disabling sources.

The [`octo.nvim`](https://github.com/pwntester/octo.nvim) plugin loads a diff view that preserves the original filetype, but the root finding always fails on the left hand buffer since it's not in the actual project - the buffer name is `octo://some/plugin/specific/stuff`.

If I could specify a custom `should_attach`, I could filter out `octo://` buffers with a small change. Currently (to the best of my knowledge), I have to go and define or modify a condition on every source to get the same effect.

jose-elias-alvarez:
[...]

PR #613, opened 2022-01-25
feat(config): add optional user-defined should_attach callback

MattSPalmer:
Per the discussion in #502, there are some buffers we always want null-ls to ignore, regardless of filetype or buftype.

Allow the user to hook into the existing `should_attach` function in the client by specifying a callback on the config object.

### DIFF
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -189,13 +189,19 @@ directory.
 
 ### should_attach (function, optional)
 
-Defines a `should_attach` callback to run before null-ls attaches to a buffer.
-The callback receives the `bufnr` as its only argument.
+A user-defined function that controls whether to enable null-ls for a given
+buffer. Receives `bufnr` as its only argument.
+
+In order to cut down potentially expensive calls, null-ls will only call
+`should_attach` after its own internal "should attach" checks pass, so it's not
+guaranteed to run every time.
 
 ```lua
-local should_attach = function(bufnr)
-    return not vim.fn.bufname(bufnr):match("^git://")
-end
+require("null-ls.nvim").setup({
+    should_attach = function(bufnr)
+        return not vim.api.nvim_buf_get_name(bufnr):match("^git://")
+    end,
+})
 ```
 
 ### sources (table, optional)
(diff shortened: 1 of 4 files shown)


doc/CONFIG.md
```diff
@@ -187,6 +187,17 @@ end
 If `root_dir` returns `nil`, the root will resolve to the current working
 directory.
 
+### should_attach (function, optional)
+
+Defines a `should_attach` callback to run before null-ls attaches to a buffer.
+The callback receives the `bufnr` as its only argument.
+
+```lua
+local should_attach = function(bufnr)
```
jose-elias-alvarez, line 196: I think this example is a little hard to put into use without additional context, since it's just a local function. Could we add more context (i.e. an example using `null_ls.setup()`) or leave this out?

jose-elias-alvarez: Also, if we keep the example, I'd prefer to use `vim.api.nvim_buf_get_name(bufnr)` over the Vimscript function.

lua/null-ls/client.lua
```diff
@@ -11,7 +11,9 @@ local lsp = vim.lsp
 local client, id
 
 local should_attach = function(bufnr)
-    if api.nvim_buf_get_option(bufnr, "buftype") ~= "" or api.nvim_buf_get_name(bufnr) == "" then
+    if c.get().should_attach and not c.get().should_attach(bufnr) then
```
jose-elias-alvarez, line 14: I think it makes sense to have this check run after all the other checks, as it's potentially more expensive. It would also help prevent users from running superfluous checks.

jose-elias-alvarez (commented):
Looks good overall, and I appreciate the test coverage!

lua/null-ls/client.lua
```diff
@@ -18,7 +18,7 @@ local should_attach = function(bufnr)
     local ft = api.nvim_buf_get_option(bufnr, "filetype")
     for _, source in ipairs(sources.get_all()) do
         if sources.is_available(source, ft) then
-            return true
+            return not c.get().should_attach or c.get().should_attach(bufnr)
```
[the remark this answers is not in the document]
MattSPalmer: That's fair, my refactor was a tad overzealous.

jose-elias-alvarez:
I think this is all set now. Thanks for working on this!

MattSPalmer:
It's a very useful project, I'm happy to help! Thanks for reviewing it.

PR #619, opened 2022-01-26
fix(client): schedule progress handler call

jose-elias-alvarez:
May fix j-hui/fidget.nvim#9. 

Calling a progress handler inside of an async callback can cause issues if that handler does something that's blocked by textlock. This change schedules the call to prevent the issue.

If a user of fidget.nvim can confirm that this fixes the issue, I'll merge it.

### DIFF
--- a/lua/null-ls/client.lua
+++ b/lua/null-ls/client.lua
@@ -211,17 +211,19 @@ M.send_progress_notification = function(token, opts)
         return
     end
 
-    handler(nil, {
-        token = token,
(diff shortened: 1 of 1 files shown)


bradsherman:
This appears to have fixed this issue for me!

mystilleef:
Fixes the issue for me.

jose-elias-alvarez:
Great, thanks for confirming. I'll go ahead and merge this, then.

Issue #538, opened 2022-01-04
Can't override flake8 errors to warnings


levouh:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.7.0-dev+812-g2f31e7b88 

### null-ls config

See below.

### Steps to reproduce

Per [742739910ef5ef93efcaed32a5b875d7833b42af](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/742739910ef5ef93efcaed32a5b875d7833b42af), warnings printed by the builtin `flake8` diagnostics are highlighted as `error` even when overridden to be `warning`.

```
cat <<EOF >> /tmp/borked.lua
vim.cmd([[
  packadd nvim-lspconfig
  packadd plenary.nvim
  packadd null-ls.nvim
]])

-- No change, just comparing to another source from a different server
require("lspconfig").pyright.setup({})

local n = require("null-ls")
local h = require("null-ls.helpers")

n.setup({
  sources = {
    n.builtins.diagnostics.flake8.with({
      on_output = h.diagnostics.from_pattern(
        [[:(%d+):(%d+): ((%u)%w+) (.*)]],
        { "row", "col", "code", "severity", "message" },
        {
          severities = {
            E = h.diagnostics.severities["warning"], -- Changed to warning!
            W = h.diagnostics.severities["warning"],
            F = h.diagnostics.severities["information"],
            D = h.diagnostics.severities["information"],
            R = h.diagnostics.severities["warning"],
            S = h.diagnostics.severities["warning"],
            I = h.diagnostics.severities["warning"],
            C = h.diagnostics.severities["warning"],
          },
        }
      ),
    }),
  },
})
EOF

mkdir -p /tmp/venvs
python -m venv /tmp/venvs/borked
source /tmp/venvs/borked
python -m pip install flake8
cat <<EOF >> /tmp/borked.py
import os

def foo():
    a=1
EOF

nvim --noplugin -u /tmp/borked.lua /tmp/borked.py
```

### Expected behavior

Overrides should show errors as warnings.

### Actual behavior

Errors are shows as errors, not respecting overrides.

### Debug log

[...]

PR #622, opened 2022-01-28
feat: Expose symbol attribute for pylint

wookayin:
The symbol attribute is used as a human-interpretable code in pylint,
for instance, "bad-indentation" for a error code W0311. (See:
https://pylint.pycqa.org/en/latest/technical_reference/features.html)

This additional metadata can be exported to as diagnostics entry,
so that users may include them in the diagnostics message.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/pylint.lua
+++ b/lua/null-ls/builtins/diagnostics/pylint.lua
@@ -22,6 +22,7 @@ return h.make_builtin({
                 code = "message-id",
                 severity = "type",
                 message = "message",
+                symbol = "symbol",
                 source = "pylint",
             },
             severities = {


wookayin:
This is an example output of `pylint foo.py -f json`:

```
[                                                  
    {                                              
        "type": "convention",                      
        "module": "foo",                             
        "obj": "",                                 
        "line": 1,                                 
        "column": 0,                               
        "endLine": null,                           
        "endColumn": null,
        "path": "foo.py",
        "symbol": "missing-module-docstring",
        "message": "Missing module docstring",
        "message-id": "C0114"
    }
]
```

jose-elias-alvarez:
This seems fine to me. I believe the resulting diagnostic is technically off-spec, according to `:help diagnostic-structure` (it should be under a `user_data` key) but I don't think it makes a big difference. 

What use case do you have in mind for this? I don't think null-ls itself has a good way to do anything useful with this as of now.

wookayin:
The following is the output from `pylint foo.py` (without the `-f json` option), where the diagnostics message includes `code` and `symbol`:
```
foo.py:1:0: C0114: Missing module docstring (missing-module-docstring)
```
but in the current default behavior diagnostics message printed via vim.diagnostics API only includes the message:
```
pylint: Missing module docstring
```
I would like to see the error codes and the symbol as well, so that I can suppress that warnings if appropriate (e.g., `# disable: pylint=missing-module-docstring`).

Also, it might be a little off-topic for this PR, but [diagnostics format string](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#diagnostics_format-string) might also be relevant for this feature. Currently `diagnostics_format` does not support format `#{m}`, `#{s}`, `#{c}`, but we may rather have a more general lua function to customize the format string. I will make another PR for this.

wookayin:
BTW is there any other sources that use non-standard entries other than specified in `:h diagnostic-structure`?  Perhaps `code` itself is already a non-standard entry ...

wookayin:
Per #632, I now think this should be placed under `user_data.null_ls.symbol`. The name `symbol` originates from pylint's  glossary, so this should be the right choice.

wookayin:
Also, since many LSP servers report `string` as code (e.g. pyright: `reportMissingImports`), I think it would be better to allow an option to configure which of `code` or `symbol` would be exposed. Because the field name `symbol` is quite arbitrary and specific to pylint, some other diagnostics plugin will still be not aware of that -- for instance, [trouble.nvim](https://github.com/folke/trouble.nvim). I now feel this approach sounds better as most users typically would prefer such human-readable codes (e.g. `missing-module-docstring`) rather than serial number codes (e.g., `C0114`).

Any thoughts? If you think this would be a better choice, I can have this PR superceded.

jose-elias-alvarez:
A config option seems like a lot (and as described in #538, we don't have a great way for users to configure these kinds of options right now). I think exposing `symbol` under `user_data` and allowing users to configure it by modifying the handler or using the functionality added in #625 is the right call for now.

jose-elias-alvarez:
It looks like the conclusion from #632 / the linked Neovim issue is that we should start putting extra data like this under` user_data.null_ls`. With the functionality added in 51f336638100ba1462e9c1fb0831464e44d02aef, users can then check for the presence of the key and modify the message accordingly. 

We can either expand this PR to include the necessary changes (making sure the diagnostic helpers correctly assign keys and ensuring that diagnostic has a `user_data.null_ls` table to make checks less annoying) or open a separate PR. Let me know if you want to work on it or if I should take over when I have time.

jose-elias-alvarez:
I described in #632 why I think this is okay as-is. Thanks for the contribution!

PR #624, opened 2022-01-28
feat(builtins): add clippy diagnostics

yuys13:
I tried to add clippy for diagnostics.
I'm just starting to learn rust (and lua), but I've been following the [documentation](https://doc.rust-lang.org/rustc/json.html) and trying to implement it.
It's working for me.
Please review.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1876,6 +1876,24 @@ local sources = { null_ls.builtins.diagnostics.chktex }
 - `command = "chktex"`
 - `args = { "-q", "-f%l:%c:%d:%k:%n:%m\n" }`
 
+#### [clippy](https://github.com/rust-lang/rust-clippy)
+
+##### About
+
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
I don't know enough about Rust to properly test / review this, but doesn't the `rust-analyzer` LSP provide support for these diagnostics? [This thread](https://users.rust-lang.org/t/setup-neovim-0-5-for-rust/60183/5) seems to indicate that it does. 

I can see that parsing the output is pretty complicated, and at a minimum we'd need to modify this to only run on save, since it doesn't work with `stdin`. Unless there's a good reason to prefer the CLI version, I'd prefer to simply guide users to the LSP version.

yuys13:
Thanks for the review and for looking into the rust development environment！
I did not know that rust-analyzer supports clippy.
I just tried it and got clippy diagnostic results with rust-analyzer. It also supports code action.
I will close this PR.

PR #625, opened 2022-01-28
Allow functions to specify diagnostics_format

wookayin:
An user-defined function can be used to set the diagnostics message,
which should be more general than the format string.

See the documentation and test specs for example.

### DIFF
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -109,7 +109,7 @@ out. Note that built-in sources can define their own timeout period and that
 users can override the timeout period on a per-source basis, too (see
 [BUILTINS.md](BUILTINS.md)).
 
-### diagnostics_format (string)
+### diagnostics_format (string or function)
 
 Sets the default format used for diagnostics. The plugin will replace the
 following special components with the relevant diagnostic information:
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
So this was requested as a feature in #403, but as mentioned there, the diagnostic API already has a very similar way to configure this. This could still be useful since it's easier to configure on a per-linter level, though. What do you think?

wookayin:
I somehow missed #403 -- thanks for the pointer to the earlier discussion.

I considered this as customization of `message` entry per linter rather than user-level formatting (vim.diagnostic.config), like an adapter for `message`. The vim diagnostic API sounds more general for user needs, but I don't think per-linter configuration will be quite pretty with that.

Overall I think there is no reason to not allow this addition, unless you are going to deprecate and remove `diagnostics_format`. But either way is fine to me. If we decide to deprecate it, using adapter will be another option at the cost of a bit of verbosity.

jose-elias-alvarez:
I agree with your logic. I considered deprecating `diagnostics_format` after the release of the new API, but I found it was too verbose to easily replace, so by the same logic I think this is a nice addition.

One change that I'd like to make is to only use the template string if the user hasn't defined `diagnostics_format` as a function. I think that's a little bit less surprising, since whatever the function returns will be shown as-is.

chaneyzorn:
Since the function can full access the diagnostics, may it be more useful to allow user to customize all diagnostics entries?

wookayin:
That can be done with adapters.

jose-elias-alvarez:
Yes, agreed. FYI I'm happy to merge this if we can change it so that we don't mix template strings and functions.

jose-elias-alvarez:
I added a `diagnostics_postprocess` hook in 51f336638100ba1462e9c1fb0831464e44d02aef that should supersede this, since it allows modifying `message` as well as any other field on the diagnostics table.

PR #627, opened 2022-01-30
feat(init) call try_add after finished handling event

AckslD:


jose-elias-alvarez:
It's strange that you are having this issue. Generally, I think `FileType` events should only fire after the buffer is loaded, and the fact that you're not able to replicate the issue with a minimal config indicates that there is an incompatibility in your config. The change itself should be harmless (though I have to think about potential consequences) but I am hesitant to merge this without more knowledge about what is causing the issue.

AckslD:
@jose-elias-alvarez I understand that. I don't completely understand the full issue but I've managed to narrow it down to how [`nvim-gps`](https://github.com/SmiteshP/nvim-gps) is loaded. I noticed that if I disabled `nvim-gps` in my config the warning went away. So I tried and put `after = 'nvim-treesitter'` in the `packer` config for `nvim-gps` and then I also don't see the warning anymore, even without the change in this branch. So something not right was happening on startup and I guess for some reason these plugins were loaded in the wrong order.

I'd like to understand exactly what is going on but at least things are working for me now without this patch.

PR #631, opened 2022-01-31
Add diagnostics and formatter for Puppet

jovandeginste:
Based on the Rubocop formatter. This uses puppet-lint.

Signed-off-by: Jo Vandeginste <Jo.Vandeginste@kuleuven.be>

### DIFF
--- a/lua/null-ls/builtins/_meta/diagnostics.lua
+++ b/lua/null-ls/builtins/_meta/diagnostics.lua
@@ -91,6 +91,9 @@ return {
   psalm = {
     filetypes = { "php" }
   },
+  puppet_lint = {
+    filetypes = { "puppet", "epuppet" }
+  },
   pydocstyle = {
     filetypes = { "python" }
   },
--- a/lua/null-ls/builtins/_meta/filetype_map.lua
+++ b/lua/null-ls/builtins/_meta/filetype_map.lua
@@ -66,6 +66,10 @@ return {
   elm = {
     formatting = { "elm_format" }
   },
+  epuppet = {
+    diagnostics = { "puppet_lint" },
+    formatting = { "puppet_lint" }
+  },
   erlang = {
     formatting = { "erlfmt" }
   },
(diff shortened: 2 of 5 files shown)


jovandeginste:
I have no experience whatsoever with Lua; please guide me where this PR could improve!

jovandeginste:
(The check seems to fail on a file I didn't touch?)

jose-elias-alvarez:
This seems to be based off a *really* old commit. Everything seems fine, but could you please rebase it from the latest master so that you can verify that it still works as expected (and so that I can test it out)?

jovandeginste:
Nov 21st is "_really_ old"? :rofl: Anyway, I rebased, and updated some things. Also added epuppet (`epp` files). Still, happy to learn where this could improve! Especially the loop is awkward I think. The "output" from `puppet-lint` is something like:

```json
[
  [
    {
      "message": "double quoted string containing no variables",
      "line": 11,
      "column": 6,
      "token": "#<PuppetLint::Lexer::Token:0x0000563212090a20>",
      "kind": "warning",
      "check": "double_quoted_strings",
      "fullpath": "/full/path/to/file.pp",
      "path": "path/to/file.pp",
      "filename": "file.pp",
      "KIND": "WARNING"
    },
    { (other issue },
  ],
  [ (other file) ],
]
```

Note the nested array; every file has a new entry, but we only check one file - should I still account for multiple files?

jovandeginste:
I updated the code - would you check if this is good? I didn't mind the multiple files yet...

jovandeginste:
Appologies - didn't update the actual branch...

jovandeginste:
I tried a few iterations on the multiple files approach, but couldn't get it to work properly; when I open multiple files, it seems to run the linter again when switching to the other buffers (although it should have received all linting issues with the first run).

jovandeginste:
I also see this message: `Client 3 quit with exit code 127 and signal 0`
I assume I'm still missing some detail somewhere; is this maybe coming from null-ls (and can you educate me?), or should I be looking at different parts of LunarVim?

jovandeginste:
If this is good, merge ahead!

jose-elias-alvarez:
Thanks!

PR #636, opened 2022-02-03
feat: Add support for checkmake tool

andrewferrier:
Adds diagnostics support for the checkmake tool (https://github.com/mrtazz/checkmake). This is something I've been using myself as a custom null-ls source for a while, formalized as a null-ls builtin.


andrewferrier:
(The stylua failure above appears to be unrelated to this change).

jose-elias-alvarez:
Yep, the failure is unrelated. This looks good–thanks!

PR #639, opened 2022-02-04
feat: Add xmllint for XML formatting

andrewferrier:
This is another custom source I've been using successfully for some time that I've reworked as a builtin. Although `xmllint` may not be a very well known tool, it's installed by default on many *nix systems as it's part of libxml2, and it works very smoothly for formatting.


jose-elias-alvarez:
LGTM, thanks for the contributions!

PR #642, opened 2022-02-04
[Cspell] Add "_fallback" severity: error -> warning

binhtran432k:
I change severity because cspell diagnostic raise error by default, it confuses me when combined with other LSP.


jose-elias-alvarez:
In isolation, this is a change for the better, but in response to #494, 567daff80cc629302e1ac196f74d0dec8738249a I added a global `fallback_severity` option. Merging this would prevent `cspell` users from overriding the severity without copy-pasting the entire built-in. 

In the future we would like to add the ability to configure this (see #538) but I think leaving the severity undefined is our best solution for now.

PR #644, opened 2022-02-04
fix(diagnostics): statix

basilgood:
I just found out that statix ​​doesn't work.
I have the latest statix ```Version: 0.5.2```
With this fix it works again.


jose-elias-alvarez:
I guess this linter is still under heavy development so changes like this are to be expected. Thanks!

PR #645, opened 2022-02-05
Add actionlint to diagnostics builtins

mattleong:
Add [actionlint](https://github.com/rhysd/actionlint) to diagnostic builtins

### DIFF
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -17,11 +17,8 @@ return h.make_builtin({
             local parser = h.diagnostics.from_json({
                 attributes = {
                     message = "message",
-                    row = "line",
-                    col = "column",
                     source = "actionlint",
                     code = "kind",
-                    file = "filename",
                     severity = 1,
                 },
             })
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -10,21 +10,17 @@ return h.make_builtin({
     generator_opts = {
         command = "actionlint",
         args = { "-no-color", "-format", "{{json .}}", "-" },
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/actionlint.lua
```diff
@@ -0,0 +1,30 @@
...
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "actionlint",
+    method = DIAGNOSTICS,
+    filetypes = { "yaml" },
+    generator_opts = {
+        command = "actionlint",
+        args = { "-no-color", "-format", "{{json .}}", "-" },
+        format = "json_raw",
```
jose-elias-alvarez, line 13: Is it sufficient to use `json` here? We generally only want `json_raw` if we plan on handling errors ourselves.

lua/null-ls/builtins/diagnostics/actionlint.lua
```diff
@@ -0,0 +1,30 @@
...
+        from_stderr = true,
+        to_stdin = true,
+        on_output = function(params)
```
jose-elias-alvarez, line 16: We can do this a bit more directly:

```lua
on_output = h.diagnostics.from_json({
    attributes = {
        message = "message",
        source = "actionlint",
        code = "kind",
        severity = 1,
    },
})
```

jose-elias-alvarez (commented):
Left a couple of comments but this looks useful. Thanks!

lua/null-ls/builtins/diagnostics/actionlint.lua
mattleong: Yup! `json` works just fine

mattleong:
@jose-elias-alvarez Updated 👍

jose-elias-alvarez:
Thanks!

PR #646, opened 2022-02-05
Add support for Yarn v2+ with PnP

stephank:
This PR is probably quite ugly, but I don't do a lot of Lua or Vim scripting normally.

I also couldn't get tests to work locally, but it seems `doc/TESTING.md` is out of date? There's a mention of a relative runtimepath, which I can only find in old versions of `minimal.vim`. I'm not sure where it is trying to find plenary and null-ls right now. I get output like:

```
Error detected while processing /src/contrib/null-ls/test/minimal.vim:
line    5:
E919: Directory not found in 'packpath': "pack/*/opt/plenary.nvim"
line    6:
E919: Directory not found in 'packpath': "pack/*/opt/null-ls.nvim"
line    8:
E5108: Error executing lua [string ":lua"]:1: module 'null-ls.config' not found:
[...]
```

What this PR does is add an extra lookup step to `from_node_modules` for finding executables installed by Yarn (v2+) when it is using Plug'n'Play. Yarn is an alternative package manager for Node.js, and PnP means it doesn't create a `node_modules` directory structure, but instead hooks the Node.js loader to read from zip files. This means local installations of Prettier, for example, live in something like: `.yarn/cache/prettier-npm-2.3.2-4467ec48dc-17ce5784ac.zip`

As a user, you'd do `yarn run prettier ...` to run Prettier. For null-ls, we also need detection, and for that we can use `yarn bin prettier`, which outputs a virtual path like: `/path/to/project/.yarn/cache/prettier-npm-2.3.2-4467ec48dc-17ce5784ac.zip/node_modules/prettier/bin-prettier.js`

Detection is technically enough, and we could do `yarn run prettier ...` from there, but instead doing `node --require .pnp.cjs <prettier virtual path> ...` saves some indirection to speed things up a little. The `--require .pnp.cjs` configures the loader in Node.js so running the virtual path works. Ideally, we'd do `NODE_OPTIONS="--require .pnp.cjs" node <prettier virtual path> ...`, because that'd also apply the loader hook to any subprocess that may be spawned, but I'm not sure what the correct way is to accomplish that in null-ls code.

This is also why I think this PR is ugly. There's a hack in place to have a `dynamic_command` that actually returns multiple args, by allowing it to return a table. But curious to hear what you think. 🙂

edit: also just noticed that there's no way to make `only_local` work with this type of lookup.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -297,6 +297,27 @@ local sources = {
 Another solution is to use the `dynamic_command` option, as described in
 [HELPERS](./HELPERS.md). Note that this option can affect performance.
 
+null-ls includes several command resolvers to handle common cases and cache
+results to prevent repeated lookups.
+
+For example, the following looks for `prettier` in `node_modules/.bin`, then
+tries to find a local Yarn Plug'n'Play install, then tries to find a global
+`prettier` executable:
+
+```lua
+local command_resolver = require("null-ls.helpers.command_resolver")
+
+local sources = {
+    null_ls.builtins.formatting.prettier.with({
+        dynamic_command = function(params)
+            return command_resolver.from_node_modules(params)
+                or command_resolver.from_yarn_pnp(params)
+                or vim.fn.executable(params.command) == 1 and params.command
+        end,
+    }),
+}
+```
+
 ## Conditional sources
 
 ### `condition`
--- a/lua/null-ls/helpers/command_resolver.lua
+++ b/lua/null-ls/helpers/command_resolver.lua
@@ -5,6 +5,23 @@ local fmt = string.format
 
 local M = {}
 
+-- Search for a local install of an executable by searching relative to
+-- startpath, then walking up directories until endpath is reached.
+function search_ancestors_for_command(startpath, endpath, executable_to_find)
+    local resolved
+    u.path.traverse_parents(startpath, function(dir)
+        local command = u.path.join(dir, executable_to_find)
+        if u.is_executable(command) then
+            resolved = { command = command, cwd = dir }
+            return true
+        end
+        if dir == endpath then
+            return true
+        end
+    end)
+    return resolved
+end
+
 M.generic = function(params, prefix)
     local resolved = s.get_resolved_command(params.bufnr, params.command)
     if resolved then
@@ -17,30 +34,15 @@ M.generic = function(params, prefix)
     local executable_to_find = prefix and u.path.join(prefix, params.command) or params.command
     log:debug("attempting to find local executable " .. executable_to_find)
 
-    local root = u.get_root()
-
-    resolved = {}
-
-    u.path.traverse_parents(params.bufname, function(dir)
-        local command = u.path.join(dir, executable_to_find)
-        if u.is_executable(command) then
-            log:trace(fmt("resolved dynamic command for [%s] with cwd=%s", executable_to_find, dir))
-            resolved.command = command
-            resolved.cwd = dir
-            return true
-        end
-
-        -- use cwd as a stopping point to avoid scanning the entire file system
-        if dir == root then
-            return true
-        end
-    end)
-
-    if not resolved.command then
+    resolved = search_ancestors_for_command(params.bufname, u.get_root(), executable_to_find)
+    if resolved then
+        log:trace(fmt("resolved dynamic command for [%s] with cwd=%s", executable_to_find, resolved.cwd))
+    else
         log:debug(fmt("Unable to resolve command [%s], skipping further lookups", executable_to_find))
+        resolved = { command = false }
     end
 
-    s.set_resolved_command(params.bufnr, params.command, { command = resolved.command or false, cwd = resolved.cwd })
+    s.set_resolved_command(params.bufnr, params.command, resolved)
     return resolved.command
 end
 
--- a/lua/null-ls/helpers/command_resolver.lua
+++ b/lua/null-ls/helpers/command_resolver.lua
@@ -53,4 +53,37 @@ M.from_node_modules = function(params)
     return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
 end
 
+M.from_yarn_pnp = function(params)
+    local cache_key = fmt("yarn:%s", params.command)
+    local resolved = s.get_resolved_command(params.bufnr, cache_key)
+    if resolved then
+        if resolved.command then
+            log:debug(fmt("Using cached value [%s] as the resolved command for [%s]", resolved.command, params.command))
+        end
+        return resolved.command
+    end
+
+    -- older yarn versions use `.pnp.js`, so look for both new and old names
+    local root = u.get_root()
+    local pnp_loader = search_ancestors_for_command(params.bufname, root, ".pnp.cjs")
+        or search_ancestors_for_command(params.bufname, root, ".pnp.js")
+    if pnp_loader then
+        local yarn_bin = vim.fn.system(
+            fmt("cd %s && yarn bin %s", vim.fn.shellescape(pnp_loader.cwd), vim.fn.shellescape(params.command))
+        ):gsub("%s+", "")
+        if vim.v.shell_error == 0 then
+            log:trace(fmt("resolved dynamic command for [%s] to Yarn PnP with cwd=%s", params.command, pnp_loader.cwd))
+            resolved = {
+                command = { "node", "--require", pnp_loader.command, yarn_bin },
+                cwd = pnp_loader.cwd,
+            }
+        end
+    end
+
+    resolved = resolved or { command = false }
+
+    s.set_resolved_command(params.bufnr, cache_key, resolved)
+    return resolved.command
+end
+
 return M
--- a/lua/null-ls/helpers/generator_factory.lua
+++ b/lua/null-ls/helpers/generator_factory.lua
@@ -278,7 +278,13 @@ return function(opts)
             local resolved_command
             if dynamic_command then
                 resolved_command = dynamic_command(params)
-                log:debug(string.format("Using dynamic command for [%s], got: %s", params.command, resolved_command))
+                log:debug(
+                    string.format(
+                        "Using dynamic command for [%s], got: %s",
+                        params.command,
+                        vim.inspect(resolved_command)
+                    )
+                )
             else
                 resolved_command = command
             end
@@ -334,8 +340,8 @@ return function(opts)
 
             log:debug(
                 string.format(
-                    "spawning command [%s] at %s with args %s",
-                    resolved_command,
+                    "spawning command %s at %s with args %s",
+                    vim.inspect(resolved_command),
                     resolved_cwd,
                     vim.inspect(resolved_args)
                 )
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -132,6 +132,17 @@ M.spawn = function(cmd, args, opts)
         parsed_env = env_merge(env)
     end
 
+    if type(cmd) == "table" then
+        local concat_args = {}
+        for i = 2, #cmd do
+            concat_args[#concat_args + 1] = cmd[i]
+        end
+        for _, arg in ipairs(args) do
+            concat_args[#concat_args + 1] = arg
+        end
+        cmd, args = cmd[1], concat_args
+    end
+
     handle = uv.spawn(
         vim.fn.exepath(cmd),
         { args = args, env = parsed_env, stdio = stdio, cwd = opts.cwd or vim.fn.getcwd() },
--- a/test/spec/loop_spec.lua
+++ b/test/spec/loop_spec.lua
@@ -147,6 +147,13 @@ describe("loop", function()
             assert.stub(uv.read_start).was_called(2)
         end)
 
+        it("should allow a table for cmd", function()
+            loop.spawn({ "cat", "-b" }, { "-n" }, mock_opts)
+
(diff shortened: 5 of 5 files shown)


jose-elias-alvarez:
Yep, the testing document is pretty outdated. To my knowledge, everything works with a standard Packer setup, but it may not work otherwise. Out of curiosity, do tests work if you revert to the `rtp` modifications in 1fadae38e9e43cbd20ab238e18c2d6f099607e74?

For the PR: I'm not sure about this approach. Fundamentally, I think allowing `resolved_command` to include extra information like `ext_command` is difficult to use without knowing about implementation details. An approach I like more is to allow `command` to be a table, where the first element is treated as the command and the others are treated as arguments (this is how [Neovim handles](https://github.com/neovim/neovim/blob/cee944e3ee548621776fde7ecd1bf65ae0609a06/runtime/lua/vim/lsp.lua#L196) the commands to spawn LSP servers). Making this generally possible would be useful beyond this PR, too. 

Also, as much as possible, I'd like for this to avoid messing with the existing `node_modules` logic. As far as I understand it, this resolution mechanism doesn't really have to do with `node_modules`, so it's logical for it to be independent. Since the `node_modules` logic is also enabled by default, not breaking it is a priority. For this PR, I think it makes sense to start with a separate Yarn PnP command resolver that users can opt into, then we can think about how to use multiple resolvers and enable both by default if everything is stable.

stephank:
I took another pass at this, and feel like it's already a lot better. Thanks for the feedback, and please feel free to nitpick on further review. I'm making a lot of assumptions on style and winging it when it comes to best practices. 🙃

- I made a separate commit for the change that allows `dynamic_command` to return a table.
- I moved the Yarn stuff to a separate resolver function `from_node_modules_or_yarn_pnp`.
- I added a `nodejs` resolver function that switches based on a new config option `enable_yarn_pnp_resolver` (default false).

> Yep, the testing document is pretty outdated. To my knowledge, everything works with a standard Packer setup, but it may not work otherwise. Out of curiosity, do tests work if you revert to the `rtp` modifications in [1fadae3](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/1fadae38e9e43cbd20ab238e18c2d6f099607e74)?

Ah, that does work better, but any test that tries to use `test.utils` still fails:

```
========================================
Testing: 	/src/contrib/null-ls/test/spec/client_spec.lua
========================================
FAILED TO LOAD FILE
/src/contrib/null-ls/test/spec/client_spec.lua:6: module 'test.utils' not found:
	no field package.preload['test.utils']
	no file '/nix/store/8zf2h7y18w45p4fqli7vv7pk64257rpz-luajit-2.1.0-2021-10-27-env/share/lua/5.1/test/utils.lua'
	no file '/nix/store/8zf2h7y18w45p4fqli7vv7pk64257rpz-luajit-2.1.0-2021-10-27-env/share/lua/5.1/test/utils/init.lua'
	no file '/nix/store/8zf2h7y18w45p4fqli7vv7pk64257rpz-luajit-2.1.0-2021-10-27-env/lib/lua/5.1/test/utils.so'
	no file '/nix/store/8zf2h7y18w45p4fqli7vv7pk64257rpz-luajit-2.1.0-2021-10-27-env/lib/lua/5.1/test.so'
========================================
```

I understand this variant a bit better; `set rtp` + `runtime` makes sense to me. But I don't understand how, in the current git head, `packadd` is supposed to pick up my null-ls working copy? I also don't follow how `require('test.utils')` can work in *any* situation if vim only looks in `lua/` directories.

FWIW, I'm not on Packer, but use Nix on macOS. I believe that prepares a pack directory and injects it at the start of `runtimepath`, so doesn't really use a plugin manager.

stephank:
Makes sense! Thanks again for the feedback. I removed the cache and global setting changes, and that does make the diff a lot cleaner.

I think I had the wrong idea of at what level cache was supposed operate, so really didn't need the earlier cache changes. I used namespacing to separate `generic` and `from_yarn_cache_only` caches, in case calls to those functions are mixed in interesting ways. Does that make sense?

jose-elias-alvarez:
Excellent, this looks much cleaner. I'll test this out and leave a review this weekend.

stephank:
Sorry, I haven't been able to get around to updating this yet, but will soon.

The `.pnp.js` filename is used by older versions of Yarn. I'll see if I can add a check for it.

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -51,4 +51,52 @@ M.from_node_modules = function(params)
     return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
 end
 
+M.from_yarn_pnp_only = function(params)
```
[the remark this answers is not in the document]
stephank: I added a `search_ancestors_for_command(startpath, endpath, executable_to_find)`. Hope that's what you had in mind. I'm not sure how buffer number was supposed to come into play?

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -51,4 +51,52 @@ M.from_node_modules = function(params)
...
+        end
+
+        -- use cwd as a stopping point to avoid scanning the entire file system
+        if dir == root then
+            return true
+        end
+    end)
+
+    if yarn_dir then
+        local yarn_bin = vim.fn.system(
```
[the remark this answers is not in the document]
stephank: Ah, didn't realize it would merge stdout/stderr. I now check `vim.v.shell_error == 0`, which should handle both 'no binary' (1) and 'no yarn' (127) cases. Though the latter will cause a wasteful lookup, but we could chalk that up to user error while we don't enable it by default.

lua/null-ls/helpers/generator_factory.lua
```diff
@@ -332,15 +342,25 @@ return function(opts)
             opts._last_args = resolved_args
             opts._last_cwd = resolved_cwd
 
+            -- concatenate resolved_command and resolved_args
```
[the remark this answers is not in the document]
stephank: I did move this logic, because that was a simple change, but it'll take broader changes to make it work for e.g. `command`. What I could find:

- [`is_executable` check](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/3059dc761b27c4c836f88879c460cd7a989caa47/lua/null-ls/helpers/generator_factory.lua#L178)
- [cache functions expect strings](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/3059dc761b27c4c836f88879c460cd7a989caa47/lua/null-ls/state.lua#L45-L65)
- [`dynamic_command` (and cache) expect strings](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/3059dc761b27c4c836f88879c460cd7a989caa47/lua/null-ls/helpers/generator_factory.lua#L276-L284)
- [Logging](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/3059dc761b27c4c836f88879c460cd7a989caa47/lua/null-ls/helpers/generator_factory.lua#L288), might be some more here and there
- Doesn't look like `command` is otherwise part of a `vim.validate` validation.

But this is all specific to the `command` option. It should be usable in other contexts.

stephank:
> This is fantastic as it allows to use eslint in yarn pnp setups. The only issue that I ran into is that this PR requires a `.pnp.cjs`. In my yarn 2 setups I do however have a `.pnp.js` file. Not sure how I can get yarn to generate a commonjs version instead. Could we handle both?

@dastein1 I added the extra check, so this should now work. :)

> Also, I made some changes in https://github.com/jose-elias-alvarez/null-ls.nvim/commit/2c8c906a5aecd784d78345f1913c63132bf9d1bb that should hopefully make it possible to run tests with a non-standard (i.e. not Packer) plugin setup. Let me know if that works, since I think adding coverage for the changes outside of the command_resolver module will be important.

@jose-elias-alvarez Tests now work like a charm for me! 👍

If you want, I can take a look at adding selene + tests to the GitHub Actions workflow, in a separate PR?

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -51,4 +53,35 @@ M.from_node_modules = function(params)
     return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
 end
 
+M.from_yarn_pnp = function(params)
+    local cache_key = fmt("yarn:%s", params.command)
+    local resolved = s.get_resolved_command(params.bufnr, cache_key)
```
jose-elias-alvarez, line 58: Perhaps we could also extract this bit of logic out, since we're repeating it in `generic` and `from_yarn_pnp`?

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -51,4 +53,35 @@ M.from_node_modules = function(params)
...
+    if resolved then
+        if resolved.command then
+            log:debug(fmt("Using cached value [%s] as the resolved command for [%s]", resolved.command, params.command))
+        end
+        return resolved.command
+    end
+
+    resolved = {}
```
jose-elias-alvarez, line 66: This is nitpicky, but I like the changes you made to the flow in the `generic` resolver and think they would improve readability here, too. Something like this:

```diff
diff --git a/lua/null-ls/helpers/command_resolver.lua b/lua/null-ls/helpers/command_resolver.lua
index 4dcb869..5e4da14 100644
--- a/lua/null-ls/helpers/command_resolver.lua
+++ b/lua/null-ls/helpers/command_resolver.lua
@@ -63,8 +63,6 @@ M.from_yarn_pnp = function(params)
         return resolved.command
     end

-    resolved = {}
-
     -- older yarn versions use `.pnp.js`, so look for both new and old names
     local root = u.get_root()
     local pnp_loader = search_ancestors_for_command(params.bufname, root, ".pnp.cjs")
@@ -75,12 +73,16 @@ M.from_yarn_pnp = function(params)
         ):gsub("%s+", "")
         if vim.v.shell_error == 0 then
             log:trace(fmt("resolved dynamic command for [%s] to Yarn PnP with cwd=%s", params.command, pnp_loader.cwd))
-            resolved.command = { "node", "--require", pnp_loader.command, yarn_bin }
-            resolved.cwd = pnp_loader.cwd
+            resolved = {
+                command = { "node", "--require", pnp_loader.command, yarn_bin },
+                cwd = pnp_loader.cwd,
+            }
         end
     end

-    s.set_resolved_command(params.bufnr, cache_key, { command = resolved.command or false, cwd = resolved.cwd })
+    resolved = resolved or { command = false }
+
+    s.set_resolved_command(params.bufnr, cache_key, resolved)
     return resolved.command
 end
```

lua/null-ls/loop.lua
```diff
@@ -132,6 +132,17 @@ M.spawn = function(cmd, args, opts)
         parsed_env = env_merge(env)
     end
 
+    if type(cmd) == "table" then
```
jose-elias-alvarez, line 135: This looks good–can we add a test?

jose-elias-alvarez (commented):
This looks excellent! I left a few comments but I think this is almost ready to merge. We don't have an ideal place for this in the documentation, but I think [the section about local executables](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#using-local-executables) might be best.

jose-elias-alvarez:
> If you want, I can take a look at adding selene + tests to the GitHub Actions workflow, in a separate PR?

Sure, this would be a big help if you want to put it together.

lua/null-ls/helpers/command_resolver.lua
stephank: Hmm, we always need the if-statement to short-circuit, so the only thing to abstract is the logging. Or back to the idea of wrapping these in a higher-order function?

lua/null-ls/helpers/command_resolver.lua
stephank: I rebase with this change 👍

lua/null-ls/loop.lua
stephank: Done

stephank:
I expanded on the `dynamic_command` section in "local executables", but not sure if you'd rather have something separately explain `command_resolver` instead?

doc/BUILTINS.md
```diff
@@ -295,7 +295,24 @@ local sources = {
 ```
 
 Another solution is to use the `dynamic_command` option, as described in
-[HELPERS](./HELPERS.md). Note that this option can affect performance.
+[HELPERS](./HELPERS.md). Note that this option can affect performance. For
+example, the following looks for `prettier` in `node_modules/.bin`, then tries
+to find a local Yarn Plug'n'Play install, then tries to find a global
+`prettier` executable:
```
jose-elias-alvarez, line 301:
```suggestion
[HELPERS](./HELPERS.md). Note that this option can affect performance. 

null-ls includes several command resolvers to handle common cases and cache results to prevent repeated lookups. 

For example, the following looks for `prettier` in `node_modules/.bin`, then tries
to find a local Yarn Plug'n'Play install, then tries to find a global
`prettier` executable:
```

Agreed that this is a little better split up.

lua/null-ls/helpers/command_resolver.lua
jose-elias-alvarez: I see what you mean. Higher-order is probably the way to go in the long term when we have more resolvers, but for now I think this is okay.

doc/BUILTINS.md
stephank: Applied this change 👍

jose-elias-alvarez:
Excellent, I think this is good to go. Thank you for your hard work and patience!

Issue #621, opened 2022-01-28
Adapters for attributes other than severities (to support customized message format)


wookayin:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

I wish diagnostics generators could allow custom attribute adapters other than the `severity` attribute. I need this to customize warning/error message formats.

For example, this is what we can see form the `pylint` command line:

```
foo.py:1:0: C0114: Missing module docstring (missing-module-docstring)
```

This is the translated `diagnostic` message:
```
{                                           
  bufnr = 1,                                
  code = "C0114",
  col = 0,
  end_col = 0,
  end_lnum = 1,
  lnum = 0,
  message = "Missing module docstring",
  namespace = 39,
  row = 1,
  severity = 3,
  source = "pylint",
}
```

However, some diagnostics builders like `from_json`, `from_pattern` cannot understand `adapters = {...}`, such as:
```diff
 on_output = h.diagnostics.from_json({
     attributes = {
         row = "line",
         col = "column",
         code = "message-id",
         severity = "type",
         message = "message",
         symbol = "symbol",
         source = "pylint",
     },
     severities = {
         convention = h.diagnostics.severities["information"],
         refactor = h.diagnostics.severities["information"],
     },
     offsets = {
         col = 1,
     },
+    adapters = {
+        message = function(entries, _)
+           return entries["code"] .. ": " .. entries["message"]
+        end,
+    },
 })
```

The adapter function will not even get executed, because the `make_attr_adapters` function in `null-ls/helpers/diagnostics` hardcodes adapters for severities only. Why don't we have it rather general?

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

-

wookayin:
Never mind, I figured out that `adapters` are expected to a list (which shouldn't be documented or gone through a sanity check?).

The following works.
```
    adapters = {
        {
               message = function(entries, _)
                  return entries["code"] .. ": " .. entries["message"]
               end,
        }
    },
```

wookayin:
I suggest

wookayin:
Actually a recommended way to customize the diagnostics message format string is: https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#diagnostics_format-string.

PR #650, opened 2022-02-06
feat(diagnostics): fallback_severity option for per built-in

chaneyzorn:
This patch makes `fallback_severity` seem to `diagnostics_format`.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -185,6 +185,22 @@ See [CONFIG](CONFIG.md) to learn about the structure of `diagnostics_format`.
 Note that specifying `diagnostics_format` for a built-in will override your
 global `diagnostics_format` for that source.
 
+### Fallback severity
+
+You can define the severity by setting `fallback_severity` when the diagnostic
+source does not explicitly define a severity.
+
+```lua
+local sources = {
+    null_ls.builtins.diagnostics.cspell.with({
+        fallback_severity = vim.diagnostic.severity.HINT,
+    }),
+}
+```
+
+Note that specifying `fallback_severity` for a built-in will override your
+global `fallback_severity` specified in [CONFIG](CONFIG.md) for that source.
+
 ### Diagnostics performance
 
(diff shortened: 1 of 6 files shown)


chaneyzorn:
[trouble.nvim](https://github.com/folke/trouble.nvim) renders diagnostics sorted by diagnostics.severity. The global `fallback_severity` sets to `vim.diagnostic.severity.ERROR`, which makes the unspecified-severity diagnostics range before other explicit-severity diagnostics.

This patch allows user to specified fallback-severity for single built-in diagnostics, such as `cspell`:

image


fallback_severity for per built-in might seem a little strange, but I can't find a easy way to customize the built-in diagnostics severity. 

Such as: https://github.com/jose-elias-alvarez/null-ls.nvim/issues/621 https://github.com/jose-elias-alvarez/null-ls.nvim/issues/538 https://github.com/jose-elias-alvarez/null-ls.nvim/pull/625

Overriding `on_output` is a heavy way.

May we need a `postprocess` (lua/null-ls/diagnostics.lua) hook for user? But It may need more works.

jose-elias-alvarez:
This is a reasonable change, but one reason I'm hesitant to merge it is because the idea of a `fallback_severity` is opaque. Basically, this would only be used if the built-in does not specifically define a severity, but that's not clear without looking at the built-in's source.

I wouldn't be opposed to adding a user hook, since it would effectively solve this and other issues by making it possible to customize anything about the diagnostics. I don't think it should be too hard to add, either–we'd just need to add a new generator option and check for it in `postprocess`.

chaneyzorn:
I think so~

Since this involves the design of the API, I think it's better to leave it to the author to take it over ;D

The PR can be closed at any time, please let me know if I can be of any help.

jose-elias-alvarez:
I added a general `diagnostics_postprocess` hook in 51f336638100ba1462e9c1fb0831464e44d02aef that should handle this and other cases. Thank you for the PR and for your input!

PR #662, opened 2022-02-09
Change prettier formatting call to fix OSX issue

victorheid:
On OSX calling it with --stdin-filepath was causing a null return. Removing the argument seems to resolve the issue.

### DIFF
--- a/lua/null-ls/builtins/formatting/prettier.lua
+++ b/lua/null-ls/builtins/formatting/prettier.lua
@@ -25,7 +25,7 @@ return h.make_builtin({
     },
     generator_opts = {
         command = "prettier",
-        args = h.range_formatting_args_factory({ "--stdin-filepath", "$FILENAME" }),
+        args = h.range_formatting_args_factory({ "$FILENAME" }),
         to_stdin = true,
         dynamic_command = cmd_resolver.from_node_modules,
     },


jose-elias-alvarez:
I don't have this issue on OSX with Prettier 2.5.1. Are you sure it's not a version issue? If you need to run a lower version, you should modify the built-in's arguments locally instead.

victor-ivre:
I do have the 2.5.1 indeed. Both on my Win and Mac - it was running fine up to recently. I will do some further investigation later on.

One thing that I notice is that in both OSX and Windows, when running the prettier cli, I get the following

```sh
$ prettier --stdin-filepath test-file.ts
$ 
```
```sh
$ prettier --stdin-filepath test-file.ts
$ const test = "Hello there"; 
```

which suggests that the --stdin-filepath option is returning null. Which is what I got from the debug log.

jose-elias-alvarez:
@victor-ivre This is expected–what about this?

```
cat test-file.ts | prettier --stdin-filepath test-file.ts
```

null-ls is using `stdin`, so not piping anything in will result in empty output.

victorheid:
I do have the 2.5.1 indeed. Both on my Win and Mac - it was running fine up to recently. I will do some further investigation later on.

One thing that I notice is that in both OSX and Windows, when running the prettier cli, I get the following

```sh
$ prettier --stdin-filepath test-file.ts
$ 
```
```sh
$ prettier test-file.ts
$ const test = "Hello there"; 
```

which suggests that the --stdin-filepath option is returning null. Which is what I got from the debug log.

(deleted previous comment as I used the wrong account, sorry)

jose-elias-alvarez:
@victorheid This is expected–what about this?

```
cat test-file.ts | prettier --stdin-filepath test-file.ts
```

Using the --stdin-filepath flag tells prettier to get input from stdin, so not piping anything in will result in empty output.

(moved comment so discussion makes sense)

victorheid:
@jose-elias-alvarez The above does produce the same results both in Win and OSX, printing the fixed file in the terminal. 

I will have to do some more debugging later to understand why its behaving like that on my Air when calling `lua vim.lsp.buf.formatting()`, probably removing all the other plugins to isolate any side effect.

victorheid:
Closing this PR for now. Today when testing again everything was working fine, something else might have caused the issue. 

As I can't replicate, I will close this and keep an eye out if it occurs again.

Thanks @jose-elias-alvarez, appreciate the time put on the plugin.

Mr-istov:
@victorheid Ah okay, thank you. I'll try that out, if it doesn't fix it I'll have to debug a little more.

victorheid:
@Mr-istov did that work? I thought I had the same experience today, but the cause was a dangling comma on the .prettierrc.json file, so might check that - not sure if that's what happened to me the first time, but worth checking in your side if this is still an issue

Mr-istov:
Hey @victorheid, sorry I managed to fix it but I forgot to update you here.
It was another issue at my end, because I wanted to use `prettier` installed from my project dependencies (not a global install, since I don't install it globally), I was reading the documentation and saw that this can be configured by adding the `prefer_local` config, and it could be defined as a path to the modules or a boolean, if `True` it will search it and load it dynamically. 
So at first I configured it `prefer_local = true` so I won't have to pass in the path and this was when it started to time out.
Then I changed it to be `prefer_local = 'node_modules/.bin'` and this fixed the issue on my end.
Not sure why it was timing out when it was `prefer_local = true`, when in the logs it was saying that it found `prettier` and that it loaded it dynamically.

PR #670, opened 2022-02-10
Add has_file utility

knpwrs:
This utility makes it easier to work in polyglot monorepos. For instance, I may have a `.prettierrc` file I want to check for in a sub-directory rather than the root directory.

An alternative name for this function could be `cwd_has_file`. This is more convenient than `vim.loop.fs_stat(filename)` because it allows for tables like the `root_has_file` function, enabling the following:

```lua
local function prettier_condition(utils)
  -- See https://prettier.io/docs/en/configuration.html
  local files = {
    '.prettierrc',
    '.prettierrc.json',
    '.prettierrc.yml',
    '.prettierrc.yaml',
    '.prettierrc.json5',
    '.prettierrc.js',
    '.prettierrc.cjs',
    '.prettier.config.js',
    '.prettier.config.cjs',
    '.prettierrc.toml',
  }

  return utils.has_file(files) or utils.root_has_file(files)
end

local sources = {
  nls.builtins.formatting.prettierd.with({
    condition = prettier_condition,
  })
}

nls.setup({
  sources = sources,
})
```

### DIFF
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -146,7 +146,8 @@ M.make_conditional_utils = function()
         has_file = function(...)
             local patterns = vim.tbl_flatten({ ... })
             for _, name in ipairs(patterns) do
-                if M.path.exists(name) then
+                local full_path = vim.loop.fs_realpath(name)
+                if full_path and M.path.exists(full_path) then
                     return true
                 end
             end
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -143,6 +143,15 @@ M.make_conditional_utils = function()
     local root = M.get_root()
 
     return {
+        has_file = function(...)
+            local patterns = vim.tbl_flatten({ ... })
+            for _, name in ipairs(patterns) do
+                if M.path.exists(name) then
+                    return true
+                end
+            end
+            return false
+        end,
         root_has_file = function(...)
             local patterns = vim.tbl_flatten({ ... })
             for _, name in ipairs(patterns) do
--- a/test/spec/utils_spec.lua
+++ b/test/spec/utils_spec.lua
@@ -324,8 +324,27 @@ describe("utils", function()
         local utils = u.make_conditional_utils()
 
         it("should return object containing utils", function()
+            assert.truthy(type(utils.has_file) == "function")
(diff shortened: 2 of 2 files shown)


lua/null-ls/utils.lua
```diff
@@ -143,6 +143,15 @@ M.make_conditional_utils = function()
     local root = M.get_root()
 
     return {
+        has_file = function(...)
+            local patterns = vim.tbl_flatten({ ... })
+            for _, name in ipairs(patterns) do
+                if M.path.exists(name) then
```
jose-elias-alvarez, line 149: Could we use [`vim.loop.fs_realpath(name)`](https://github.com/luvit/luv/blob/master/docs.md#uvfs_realpathpath-callback) to ensure that we are checking the existence of the file at its absolute path?

jose-elias-alvarez (commented):
One small change, but I think this is a useful addition–thanks!

lua/null-ls/utils.lua
knpwrs: Done in 03f5420, but now the tests fail. I'm fairly weak with lua / vimscript / vimscript expressed as lua. Feel free to push any changes you like.

knpwrs: In addition to the tests failing I see an error whenever I open neovim now.

jose-elias-alvarez: Try this and let me know if it works:

```diff
diff --git a/lua/null-ls/utils.lua b/lua/null-ls/utils.lua
index 0262fd1..f804410 100644
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -146,7 +146,8 @@ M.make_conditional_utils = function()
         has_file = function(...)
             local patterns = vim.tbl_flatten({ ... })
             for _, name in ipairs(patterns) do
-                if M.path.exists(vim.loop.fs_realpath(name)) then
+                local full_path = vim.loop.fs_realpath(name)
+                if full_path and M.path.exists(full_path) then
                     return true
                 end
             end
```

knpwrs: Looks like that works! I pushed it in 421c66a.

jose-elias-alvarez:
Great, I think this is okay now. Thanks for the PR!

PR #671, opened 2022-02-10
feat(info): enable info when not attached

### LINKED ISSUE #664: Improve behavior of :NullLsInfo
### Discussed in https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/663

<sup>Originally posted by **filipekiss** February  9, 2022</sup>
First of all, thank for all the hard work you put in this plugin. It really saves a lot of time when trying to configure non-LSP sources.

I was first setting up the plugin and I did a simple call to `require('null-ls').setup()` and things worked as expected (as in I got no errors).

I understand that, by default, NullLs doesn't attach to the buffers, which makes sense, but, at the same time, it renders the command a little bit useless.

So I was thinking that maybe the command `:NullLsInfo` could be improved in a way that, by default, if no sources are configured for the current buffer, it would detect the filetype and show me the available sources from the builtins.

That way, I can open a file for a language and run `:NullLsInfo` and I'll have a list of all available sources that I can use, making discoverability for builtins a lot better.

I haven't looked at the source code yet so I'm not sure how difficult it would be to implement something like this, but I think it would greatly improve the overall experience of using and configuring NullLs, especially for new comers.

Once again, thanks for all  your hard work and let me know if I can help in any way :)</div>

@filipekiss I converted this into an issue. Implementing this actually wouldn't be hard, but there are probably some edge cases I'm not seeing right now. I'll work on it when I have time (or anyone else who happens to see this and wants to work on it can pick it up).

filipekiss:
Let me know if anything needs to be changed! 

CleanShot 2022-02-10 at 19 47 19

This is how it looks when not attached.

Closes #664

### DIFF
--- a/lua/null-ls/info.lua
+++ b/lua/null-ls/info.lua
@@ -160,6 +160,14 @@ M.show_window = function()
         end
     end
 
+    if not is_attached then
+        local info_lines = { "* Note: currently buffer has no sources attached" }
+        table.insert(highlights, { "Type", info_lines[1] })
+
+        lines = vim.list_extend(lines, indent_lines({ "" }))
+        lines = vim.list_extend(lines, indent_lines(info_lines))
+    end
+
     local win_bufnr, win_id = make_window(0.8, 0.7)
 
     api.nvim_buf_set_lines(win_bufnr, 0, -1, true, lines)
--- a/lua/null-ls/info.lua
+++ b/lua/null-ls/info.lua
@@ -161,7 +161,7 @@ M.show_window = function()
     end
 
     if not is_attached then
-        local info_lines = { "* Note: currently buffer has no sources attached" }
+        local info_lines = { "* Note: current buffer has no sources attached" }
         table.insert(highlights, { "Type", info_lines[1] })
 
         lines = vim.list_extend(lines, indent_lines({ "" }))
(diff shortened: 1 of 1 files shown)


lua/null-ls/info.lua
```diff
@@ -140,22 +140,24 @@ M.show_window = function()
...
+    local logger_info = nil
+    if is_attached then
+        sources_info = create_active_sources_info(filetype)
+        logger_info = create_logging_info()
+    end
 
     -- stylua: ignore
-    for _, section in ipairs({
-        { "" },
+    for _, section in pairs({
```
jose-elias-alvarez, line 151: In testing, this seems to work, but to my knowledge, `pairs` doesn't guarantee order, so it might be safer to build up the table as we go and continue using `ipairs`.

jose-elias-alvarez (commented):
In addition to the suggested change, I think it might be a good idea to add a message clearly specifying that the current buffer isn't attached.

lua/null-ls/info.lua
filipekiss: The issue with `pairs` not guarantying order is when either the table is changed or there are mixed keys of integers and strings, for example. 

In this case, since the table is being implicitly indexed by Lua itself, there should be no issues with using `pairs`

filipekiss:
> In addition to the suggested change, I think it might be a good idea to add a message clearly specifying that the current buffer isn't attached.

CleanShot 2022-02-11 at 09 16 58

I've added a message at the bottom of the window. Let me know if you have any suggestions.

jose-elias-alvarez:
Thank you! I think this is a nice improvement.

filipekiss:
Glad I could contribute! :D

Issue #679, opened 2022-02-12
Pylama diagnostics getting "stuck" and not being rerun on save


triorph:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.7.0-dev+1063-ga7321e37a

### Operating System

Arch (via virtualbox)

### Minimal config

-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
	local path_sep = on_windows and "\\" or "/"
	local result = table.concat({ ... }, path_sep)
	return result
end

vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir
if on_windows then
	temp_dir = vim.loop.os_getenv("TEMP")
else
	temp_dir = "/tmp"
end

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
	local null_ls = require("null-ls")
	-- add only what you need to reproduce your issue
	null_ls.setup({
		sources = {
			require("null-ls").builtins.formatting.black,
			require("null-ls").builtins.diagnostics.pylama.with({
				extra_args = { "--linters=print,mccabe,pycodestyle,pyflakes", "--ignore=E501,W0612,W605,E231,E203" },
			}),
		},
		on_attach = function(client)
			if client.resolved_capabilities.document_formatting then
				vim.cmd("autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync(nil, 5000)")
			end
		end,
		debug = true,
	})
end

local function load_plugins()
	-- only add other plugins if they are necessary to reproduce the issue
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				"tami5/lspsaga.nvim",
				config = function()
					require("lspsaga").init_lsp_saga()
				end,
			},
			{
				"jose-elias-alvarez/null-ls.nvim",
[...]

PR #680, opened 2022-02-13
Handle pylama giving non-json outputs when the Python will not compile

triorph:
This is for issue: https://github.com/jose-elias-alvarez/null-ls.nvim/issues/679

This is my first time doing a pull request for someone on github, and I'm also unfamiliar with using lua. Let me know if you have any problems with my process here.

I'm also not super sure on the approach I've taken to "solve" the issue here, but I suspect its "good enough". Let me know if you'd like me to try a different approach.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/pylama.lua
+++ b/lua/null-ls/builtins/diagnostics/pylama.lua
@@ -9,11 +9,12 @@ return h.make_builtin({
     generator_opts = {
         command = "pylama",
         to_stdin = true,
-        from_stderr = true,
+        from_stderr = false,
+        ignore_stderr = true,
         args = { "--from-stdin", "$FILENAME", "-f", "json" },
         format = "json_raw",
         check_exit_code = function(code)
-            return code < 1
+            return code <= 1
         end,
         on_output = h.diagnostics.from_json({
             attributes = {
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -642,6 +642,41 @@ describe("diagnostics", function()
         end)
     end)
 
+    describe("pylama", function()
+        local linter = diagnostics.pylama
+        local parser = linter._opts.on_output
+        local exit_func = linter._opts.check_exit_code
+
+        it("should create a diagnostic with error severity", function()
+            local output = vim.json.decode([[
+                  [{
+                    "lnum": 3,
+                    "col": 1,
+                    "etype": "E",
+                    "message": "block comment should start with '# '",
+                    "number": "E265",
+                    "source": "run-clang-tidy.py"
+                  }]
+            ]])
+            local diagnostic = parser({ output = output })
+            assert.same({
+                {
+                    row = 3,
+                    col = 1,
+                    severity = 1,
+                    code = "E265",
+                    message = "block comment should start with '# '",
+                    source = "run-clang-tidy.py",
+                },
+            }, diagnostic)
+        end)
+        it("should count exit code of 1 as success", function()
+            assert.is.True(exit_func(1))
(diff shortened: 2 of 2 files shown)


lua/null-ls/builtins/diagnostics/pylama.lua
```diff
@@ -15,26 +15,32 @@ return h.make_builtin({
...
-                I = h.diagnostics.severities["warning"],
-                C = h.diagnostics.severities["warning"],
-                T = h.diagnostics.severities["warning"],
-                F = h.diagnostics.severities["information"],
-                D = h.diagnostics.severities["information"],
-                R = h.diagnostics.severities["information"],
-            },
-        }),
+        on_output = function(params)
+            if params.err == "failed to decode json: Expected value but found invalid token at character 1" then
```
jose-elias-alvarez, line 19:
```suggestion
            if params.err:find("failed to decode json") then
```
I think it's a little better to match against a general JSON error to avoid this breaking again if the error format changes.

lua/null-ls/builtins/diagnostics/pylama.lua
```diff
@@ -15,26 +15,32 @@ return h.make_builtin({
...
+                params.err = nil
+                params.output = { {} }
```
jose-elias-alvarez, line 21:
```suggestion
                return {}
```
We should be able to just return an empty able here instead of passing it through the parser.

jose-elias-alvarez (commented):
I think this approach is good overall but I left a couple of comments. Thanks!

lua/null-ls/builtins/diagnostics/pylama.lua
triorph: Thanks. I wasn't sure how to do this so went with the more specific case. This is much better.

lua/null-ls/builtins/diagnostics/pylama.lua
triorph: Don't we want to return params with output = {}?

triorph: Nope {} is correct. My bad.

triorph:
On a different approach: 
When running pylama from the terminal, the json goes to stdout and the non-json python errors go to stderr. However inside whatever calls python in null-ls, it all goes to stderr and I'm not quite sure why. If we could resolve that, then from_stderr=false, ignore_stderr=true would be a better solution.

jose-elias-alvarez:
> On a different approach: When running pylama from the terminal, the json goes to stdout and the non-json python errors go to stderr. However inside whatever calls pylama in null-ls, it all goes to stderr and I'm not quite sure why. If we could resolve that, then from_stderr=false, ignore_stderr=true would be a better solution.

I looked into this a bit, and the issue seems to be that `pylama` exits with code `1` if the file contains linting errors. Changing the built-in's `check_exit_code` callback to accept `1` fixes the issue and makes it possible to implement the fix you suggested, which I think is much cleaner.

triorph:
> > On a different approach: When running pylama from the terminal, the json goes to stdout and the non-json python errors go to stderr. However inside whatever calls pylama in null-ls, it all goes to stderr and I'm not quite sure why. If we could resolve that, then from_stderr=false, ignore_stderr=true would be a better solution.
> 
> I looked into this a bit, and the issue seems to be that `pylama` exits with code `1` if the file contains linting errors. Changing the built-in's `check_exit_code` callback to accept `1` fixes the issue and makes it possible to implement the fix you suggested, which I think is much cleaner.

Thanks for your help! I appreciate you taking the time to help me learn this. I've pushed what i think is the correct fix now.

jose-elias-alvarez:
No worries and thanks for the bug report and PR!

PR #682, opened 2022-02-13
add djlint support

shaeinst:
add [djlint](https://github.com/Riverside-Healthcare/djLint) to support django,jinja formatting.


jose-elias-alvarez:
Thank you for the contribution!

PR #685, opened 2022-02-14
Add support for builtin astyle.

academo:
This is a simple file formatter that works well for Arduino files.

I also added filetypes `"c", "cpp", "cs", "java" ` as this formatter also supports them.

I tested this locally and it works well.


jose-elias-alvarez:
LGTM, thank you!

PR #687, opened 2022-02-14
fix auto insert of 'log' to current source file

shaeinst:
Formatting with djList also inserts "0 files were updated."  on current source file.

> The quiet option just skips printing the diff, it doesn't suppress all output.
> Using stdin will surpress all output.

more: [Riverside-Healthcare/djLint/issues/184](https://github.com/Riverside-Healthcare/djLint/issues/184)


jose-elias-alvarez:
Thanks!

PR #688, opened 2022-02-15
Allow exit 1 for statix action

meck:
Never version uses 1 to indicate possible fixes, fix should be backward compatible.


jose-elias-alvarez:
Thank you!

PR #690, opened 2022-02-16
feat(rescript/formatting): add extension and node_modules resolver

aspeddro:


jose-elias-alvarez:
Seems like a good change, thanks!

PR #703, opened 2022-02-23
refactor(formatting)!: make formatting API LSP-independent

### LINKED ISSUE #697: Provide function that exposes formatting without hooking into LSP
### Issues

### Feature description

I get a *ton* of questions about how to disable formatting for language servers in favor of only using null-ls. The current solution is quite hacky (see: https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Avoiding-LSP-formatting-conflicts) as it effectively just tells neovim to ignore formatting requests in buf_request_all.

I'm not sure why this functionality cannot just be directly exposed, there is rarely an instance when a user would want to format with both null-ls and a language server (which is what buf request all effectively does). Then users can map a key to call the `null-ls` formatting directly. I don't see the value in shimming this into the protocol.

### Help

No

jose-elias-alvarez:
Closes #697. This is a potential solution to the issue of formatting conflicts and represents a potential divergent path for the plugin as a whole, but as it's a significant breaking change and nothing is immediately broken, the plan for now is to wait and see. See below for further discussion.

As mentioned in the linked issue:

> The initial logic behind hooking into LSP formatting was to allow users to have a "unified" setup that lets them, say, use the same keybindings and commands for LSP formatting and null-ls formatting without having to worry about which one is currently active.

However, unlike other LSP features, it's quite rare to want to combine null-ls formatting and formatting from actual language servers. The headaches caused by conflicts outweigh the benefits of unification, and so the goal of this PR is to refactor the formatting API to make it independent from the LSP formatting family.

In practice, this has the following implications:

1. The following functions no longer interact with null-ls:

```lua
vim.lsp.buf.formatting_sync()
vim.lsp.buf.formatting()
vim.lsp.buf.range_formatting()
```

2. Instead, users can use the following functions, which should work as drop-in replacements:

```lua
require("null-ls").formatting_sync()
-- for convenience, these will resolve and use the relevant LSP formatting handler
require("null-ls").formatting()
require("null-ls").range_formatting()
```

3. To make it easier to prioritize null-ls formatting but fall back to LSP formatting, which seems to be the most common use case, I've added an option which should make the transition easier (most users will only have to replace a single line in their configs):

```lua
local my_common_on_attach = function(client)
    if client.resolved_capabilities.document_formatting then
        vim.cmd([[
            augroup LspFormatting
                autocmd! * <buffer>
                autocmd BufWritePre <buffer> lua require("null-ls").formatting_sync({ fallback_to_lsp = true })
            augroup END
            ]])
    end
end
```

4. The new functions return `false` if null-ls is unable to format the current buffer, meaning users who want more control can chain null-ls formatting and LSP formatting (something many users are already accustomed to from plugins like nvim-cmp):

```lua

_G.formatting_sync = function() 
    -- pretend this is doing something more interesting
    return require("null-ls").formatting_sync() or vim.lsp.buf.formatting_sync()
end

local my_common_on_attach = function(client)
    if client.resolved_capabilities.document_formatting then
        vim.cmd([[
            augroup LspFormatting
                autocmd! * <buffer>
                autocmd BufWritePre <buffer> lua formatting_sync()
            augroup END
            ]])
    end
end
```

5. More advanced use cases (e.g. async formatting) can still be handled by directly sending requests to the client using null-ls specific methods:

```lua
_G.formatting = function()
    local bufnr = vim.api.nvim_get_current_buf()
    local client = require("null-ls.client").get_client()
    if not client then
        return
    end

    local params = vim.lsp.util.make_formatting_params()
    client.request(require("null-ls").methods.FORMATTING, params, function(err, res)
        if err then
            -- handle error
        elseif res then
            -- handle response
        end
    end, bufnr)
end
```

This is a breaking change for many null-ls users, but as mentioned, most users should only have to change a single line in their configs, with the advantage of removing code that disables server capabilities. I also see this as a potential step towards making null-ls more LSP-independent, which in the absence of other changes is one way to ensure the plugin's continued longevity. 

- [ ] Add tests
- [x] Update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,14 +202,10 @@ plugin before sending anything upstream.
 
 ### How do I format files?
 
-null-ls formatters run when you call `vim.lsp.buf.formatting()` or
-`vim.lsp.buf.formatting_sync()`. If a source supports it, you can run range
-formatting by visually selecting part of the buffer and calling
-`vim.lsp.buf.range_formatting()`.
-
-### How do I stop Neovim from asking me which server I want to use for formatting?
-
-See [this wiki page](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Avoiding-LSP-formatting-conflicts).
+null-ls formatters run when you call `:lua require("null-ls").formatting()` or
+`:lua require("null-ls").formatting_sync()`. If a source supports it, you can run
+range formatting by visually selecting part of the buffer and calling
+`:lua require("null-ls").range_formatting()`.
 
 ### How do I format files on save?
 
@@ -217,13 +213,12 @@ See the following snippet:
 
 ```lua
 require("null-ls").setup({
-    -- you can reuse a shared lspconfig on_attach callback here
     on_attach = function(client)
         if client.resolved_capabilities.document_formatting then
             vim.cmd([[
             augroup LspFormatting
                 autocmd! * <buffer>
-                autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync()
+                autocmd BufWritePre <buffer> lua require("null-ls").formatting_sync()
             augroup END
             ]])
         end
@@ -231,10 +226,12 @@ require("null-ls").setup({
 })
 ```
 
-You can also set up async formatting, as described on [this wiki
-page](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Async-formatting).
-Please read the Caveats section there to understand the meaning of (and
-drawbacks of) async formatting.
+If you want to try formatting with null-ls and fall back to LSP formatting, add
+the following option:
+
+```lua
+require("null-ls").formatting_sync({ fallback_to_lsp = true })
+```
 
 ### How do I view project-level diagnostics?
 
@@ -259,15 +256,8 @@ option after you've collected the information you're looking for.
 
 ### Does it work with (other plugin)?
 
-In most cases, yes. null-ls tries to act like an actual LSP server as much as
-possible, so it should work seamlessly with most LSP-related plugins. If you run
-into problems, please try to determine which plugin is causing them and open an
-issue.
-
-[This wiki
-page](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Compatibility-with-other-plugins)
-mentions plugins that require specific configuration options / tweaks to work
-with null-ls.
+In most cases, yes. If you run into problems, please try to determine which
+plugin is causing them and open an issue.
 
 ### How does it work?
 
@@ -282,18 +272,6 @@ More testing is necessary, but since null-ls uses pure Lua and runs entirely in
 memory without any external processes, in most cases it should run faster than
 similar solutions. If you notice that performance is worse with null-ls than
 with an alternative, please open an issue!
-
-### I am seeing a `vim.lsp.buf.formatting_sync: timeout` error message
-
-This issue occurs when a formatter takes longer than the default timeout value
-of the `formatting_sync` function. This is an automatic mechanism and controlled
-by Neovim. You might want to increase the timeout in your `formatting_sync`
-call:
-
-```lua
--- increase timeout to 2 seconds
-vim.lsp.buf.formatting_sync(nil, 2000)
-```
 
 ## Tests
 
--- a/doc/FORMATTING.md
+++ b/doc/FORMATTING.md
@@ -1,0 +1,94 @@
+# Formatting
+
+The null-ls formatting API is analogous to Neovim's LSP formatting API and
+provides the following methods, exposed under the main `require("null-ls")`
+module.
+
+```lua
+local null_ls = require("null-ls")
+
+null-ls.formatting(options)
+null-ls.formatting_sync(options, timeout_ms)
+null-ls.range_formatting(options, start_pos, end_pos)
+```
+
+## `null-ls.formatting(options)`
+
+Analogous to `vim.lsp.buf.formatting`. Accepts the following arguments:
+
+- `options` (table, optional): See `:help vim.lsp.buf.formatting`. Note that as
+  of now, null-ls does not recognize these options.
+
+  null-ls accepts an extra option, `options.fallback_to_lsp` (boolean), which will
+  automatically call `vim.lsp.buf.formatting(options)` if null-ls formatting is
+  unavailable.
+
+## `null-ls.formatting_sync(options, timeout_ms)`
+
+Analogous to `vim.lsp.buf.formatting`. Accepts the following arguments:
+
+- `options` (table, optional): See `:help vim.lsp.buf.formatting_sync`. Note
+  that as of now, null-ls does not recognize these options.
+
+  null-ls accepts an extra option, `options.fallback_to_lsp` (boolean), which
+  will automatically call `vim.lsp.buf.formatting_sync(options, timeout_ms)` if
+  null-ls formatting is unavailable.
+
+- `timeout_ms` (number, optional): See `:help vim.lsp.buf.formatting_sync`. Note
+  that as of now, null-ls does not recognize this option (timeouts are
+  source-specific, as described in [SOURCES](./SOURCES.md).
+
+## `null-ls.range_formatting(options, start_pos, end_pos)`
+
+Analogous to `vim.lsp.buf.range_formatting`. Accepts the following arguments:
+
+- `options` (table, optional): See `:help vim.lsp.buf.range_formatting`. Note
+  that as of now, null-ls does not recognize these options.
+
+  null-ls accepts an extra option, `options.fallback_to_lsp` (boolean), which
+  will automatically call `vim.lsp.buf.range_formatting(options, start_pos, end_pos)` if null-ls formatting is unavailable.
+
+- `start_pos` (number, optional): See `:help vim.lsp.buf.range_formatting`.
+  Defaults to the start of the last visual selection.
+- `end_pos` (number, optional): See `:help vim.lsp.buf.range_formatting`.
+  Defaults to the end of the last visual selection.
+
+## Custom formatting
+
+### Chaining
+
+The API methods described above return `false` if null-ls is unable to format
+the current buffer. This makes it possible to fall back to another formatting
+source if null-ls is unavailable, as in this example:
+
+```lua
+local custom_formatting = function()
+    if not require("null-ls").formatting_sync() then
+      -- fallback
+    end
+end
+```
+
+### Manual requests
+
+For advanced use cases, you can manually send requests to the null-ls client:
+
+```lua
+local manual_formatting = function()
+    local bufnr = vim.api.nvim_get_current_buf()
+    local client = require("null-ls.client").get_client()
+    if not client then
+        return
+    end
+
+    local params = vim.lsp.util.make_formatting_params()
+    -- note that the method must correspond to an internal formatting method
+    client.request(require("null-ls").methods.FORMATTING, params, function(err, res)
+        if err then
+            -- handle error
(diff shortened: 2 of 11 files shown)


jose-elias-alvarez:
A related question that came up while working on this: Does it make sense to maintain the idea of a client that receives requests / notifications? It definitely simplifies the refactor, and I feel like it may continue to be a useful abstraction, but it's worth thinking about.

Furkanzmc:
Recently, Neovim introduced `vim.lsp.formatexpr()` so we can set that to `formatexpr` option. This was a really nice change as I like seeing plugins leveraging Vim built-ins rather than creating their own thing. Now that formatting is being decoupled, I think this will break.

> Does it make sense to maintain the idea of a client that receives requests / notifications?

I don't fully understand the need for decoupling the formatting from LSP, null-ls is essentially an LSP server and making it kind of an LSP server and kind of a something like ALE is a bit confusing to me. I would think Neovim would make it easier to turn on/off capabilities of LSP servers so you don't have to provide a hacky solution.

 I love your point about using a unified interface for all this, and as a user, I don't see a good reason to decouple this and now requiring seperate mappings, and this unification idea is what drew me to null-ls in the first place.
For example, what If I'm using clangd's own formatting, and I want to use null-ls for Python? Now I have to have different configuration for each. In my case, I'm only using null-ls for formatting, but I would count this out just because I'm not using it.

Also, technically, all of the features of null-ls can be implemented without having to provide an LSP interface. Now with this PR, you are changing it so that users will have to change their mappings. But do you **really** need to support LSP interface for code actions? I don't have great insight into this, but I doubt that you really need LSP for this or the other features of null-ls.

But if the decision to remove formatting from LSP is set in stone, then I don't think it makes sense to keep the client capability for formatting. But then, why bother keeping a separate formatting implementation? Why not defer that to another plugin? It would make the maintenance of null-ls much simpler. Since it's decoupled from LSP, there's not really a strong reason to keep it in the same source code.

> I also see this as a first step towards making null-ls more LSP-independent, which I consider necessary to the plugin's continued development and longevity.

Did you document your thoughts on this in some issue? I'm very interested to understand why you think that way.

jose-elias-alvarez:
@Furkanzmc There's a lot to break down here, and I'm happy to address specific concerns, but to give you an overview, I don't think that the current approach of pretending to be an LSP server is sustainable. 

First, the current approach only works because [we override `rpc.start`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/ae1edec262c11964d45188b56af19135c5e38c89/lua/null-ls/rpc.lua#L59) to create a client without a corresponding server, but this is a hack that could break without warning. The chances of getting official support for serverless client creation are low, and the only alternative is to actually spawn a server (in fact, null-ls used to spawn a second instance of Neovim as a dummy server, but that had its own share of problems).

Apart from the `rpc.start` hack, the current approach relies on other internal details, so things can and do break. This means we're constantly playing catch-up and can never reach a truly "stable" state. I've kept up with things so far, but as you can imagine, a situation where *any* Neovim change can break the plugin–not just announced breaking changes–is not ideal. 

One path forward is to follow the `vim.diagnostic` model and work on official APIs that allow for integration of LSP and non-LSP sources (I opened a [Neovim issue](https://github.com/neovim/neovim/issues/17511) to gauge interest on this). null-ls would then be something of a library for users to easily create sources (especially using CLI tools, which I think is our strength) as well as a repository for shared sources. Even in this case, I would want null-ls to remain "LSP-like" so that users can continue to use the same mental model for configuration and share hooks like `on_attach` and `on_exit`.

To comment on formatting: As you note, the nice thing about null-ls (and the reason I wrote it in the first place) is that it lets you combine LSP and non-LSP sources into a single unified setup / display. Formatting is different because most users don't actually want to combine LSP and non-LSP formatting–they just want one or the other–so hooking into the LSP infrastructure creates conflicts with minimal advantages. You're right that formatting could be deferred to another plugin, but I think the availability of built-ins, the ease of defining new ones, and niceties like diff formatting and sequential formatting still justify its inclusion (though anyone who wants to use the implementation as a base for a separate plugin is completely free to do so). 

Nothing is set in stone, and I'm curious to see how the Neovim maintainers react to my issue, but my goal is for null-ls to become a stable, accepted part of the ecosystem, and so I feel that changes like this are at least a step in the right direction. I'm happy to continue discussing and am always open to other solutions.

mjlbach:
> Recently, Neovim introduced vim.lsp.formatexpr() so we can set that to formatexpr option. This was a really nice change as I like seeing plugins leveraging Vim built-ins rather than creating their own thing. Now that formatting is being decoupled, I think this will break.

It should be fairly trivial to write a formatexpr wrapper that first calls null-ls, then calls the built-in client's formatexpr.

> I don't fully understand the need for decoupling the formatting from LSP, null-ls is essentially an LSP server and making it kind of an LSP server and kind of a something like ALE is a bit confusing to me.

I think Jose's post addresses this, but in my view null-ls isn't a language server because it is a single, in-process instance with a shared state.

> For example, what If I'm using clangd's own formatting, and I want to use null-ls for Python? Now I have to have different configuration for each. In my case, I'm only using null-ls for formatting, but I would count this out just because I'm not using it.

I think this is addressed in the original PR description no? Your keybinding first calls null-ls, then calls an LSP request if that fails? I think this gives users more control. Unless you are running a single server, you likely have an opinion on which formatter (of multiple servers, and potentially null-ls) you would like called, so you need to prioritize this somehow.

Option 1: We introduce an API for disabling capabilities on specific servers.

Option 2 (my preferred/this PR): a user maps the keybinding that calls the formatters in the appropriate order.

To clarify, https://github.com/neovim/neovim/issues/15847 is *only*, in my view, about overriding the input/output stream handles, so it won't help null-ls in any way.

Furkanzmc:
> @Furkanzmc There's a lot to break down here, and I'm happy to address specific concerns, but to
> give you an overview, I'm not sure that the current approach of pretending to be an LSP server is
> sustainable (in the absence of changes that are out of my control).
> 
> First, the current approach only works because [we override
> `rpc.start`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/ae1edec262c11964d45188b56af19135c5e38c89/lua/null-ls/rpc.lua#L59)
> to create a client without a corresponding server, but this is a hack that could break without
> warning. The chances of getting official support for serverless client creation are low, and the
> only alternative is to actually spawn a server (in fact, null-ls used to spawn a second instance
> of Neovim as a dummy server, but that had its own share of problems).
> 
> Apart from the `rpc.start` hack, the current approach relies on other internal details, so things
> can and do occasionally break, which means it's difficult to reach a truly "stable" state. As you
> can imagine, a situation where _any_ Neovim change can theoretically break the plugin is not
> ideal, though the situation has improved and this is something we can continue to improve on our
> end.

@jose-elias-alvarez I now realize I didn't have enough insight to have an opinion on this topic. My
opinions were solely end user focused. There's great strength in null-ls because it can mimic LSP
capabilities. IMHO, I value the `SP` part of `LSP` more than the `L` part. It's a service protocol,
and the language side is just because of the general way it's used. Reading your explanation now, I
can see how frustrating it is to keep maintaining this project. Seeing how fast Neovim evolves, you
would basically be fixing regressions left and right. Thanks for explaining it.

> One potential path forward is to follow the `vim.diagnostic` model and work on official APIs that
> allow for integration of LSP and non-LSP sources (I opened
> [neovim/neovim#17511](https://github.com/neovim/neovim/issues/17511) to gauge interest on this).
> null-ls would then be something of a library for users to easily create sources (especially using
> CLI tools, which I think is our strength) as well as a repository for shared sources. Even in this
> case, I would want null-ls to remain "LSP-like" so that users can continue to use the same mental
> model for configuration and share hooks like `on_attach` and `on_exit`.

This is a good idea but I don't know what other purpose this would have other than supporting
null-ls, maybe there won't be much interest from the Neovim developers.

One of the (many) good things about null-ls is how easy it is to configure and use the same language
you use to configure your IDE (Every time I issue my custom LSP commands, I thank you in my mind. So
thanks again for a wonderful project!). Prior to this, I was using
[efm-langserver](https://github.com/mattn/efm-langserver), however the problem with that was the
lack of proper configuration. It initially didn't support Lua at all and later the feature was added
but it was just a representation of the YAML content, nothing sophisticated. Maybe this changed
later on, but it wasn't enough for my needs.

Now that I understand this better, I also support your idea of turning null-ls into a library. But I
still think having formatting be different than the rest of the functionality is confusing. This is
an open source project, so you obviously have all the right to do whatever you like but maybe
letting this implementation be in a beta stage for a while and then slowly start migrating the rest
of it might be worth it. I would happily help with this since I use null-ls daily.

What I understand from what you said is:

1. The library: This one provides all the interface for all the LSP machinery (formatting, code
actions, completion etc.) The library would essentially be a library like you said, and the users would
use null-ls to issue whatever actions they would want in a very similar approach to what you
suggested in your PR description. In this implementation, there would be nothing about LSP
implementation and you wouldn't need to conform to the LSP specification. You would be free to make
whatever API choice you'd like.

2. The LSP frontend [Optinal, my opinion]: This module would enable all the niceties of what null-ls
provides today. It would be the "LSP" that calls into the library. Maybe this could even be used as
a nice transition so that not that many people are affected. Using this, we wouldn't need to change
any code in our config and would still keep using null-ls like an LSP. I think the maintenance of
this should mostly fall on the shoulders of the community though. I would be happy to be one of the
volunteers for this. Since there's already a working system, it'd be a shame to throw away all that
code. This would also make it so the completion sources can keep working as usual as well.

With all this in place, one could even create an executable that calls into the library and have it
used in other editors. I've been thinking about doing something similar for a while now. I'd be very
interested in providing team specific tools using a common executable that would work with any
editor that supports LSP, and one that I can extend using Lua functions. So basically like
efm-langserver but with proper scripting support (I know you can technically write a Python script
and call it with efm-langserver which is what I did before but it doesn't feel natural and accessing
Vim becomes very difficult.).

> ... but my goal is for null-ls to become a stable, accepted part of the ecosystem, and so I feel
> that changes like this are at least a step in the right direction in a way that's within our
> control.

You are right. After I submitted the code completion PR things got pretty busy for me and I was not
that active here. But I would very much like to see null-ls being in the forefront of editor action
configuration akin to nvim-lspconfig.

-----

@mjlbach I think my issue is with the separation of formatting from the rest of the things. It's not
really a technical reason, but probably a philosophical one. Again though, I'm just expressing my
opinion but I have not contributed enough to the project to feel like my opinion has any weight. I
don't want to cause an trouble here. :)

> I think Jose's post addresses this, but in my view null-ls isn't a language server because it is a
> single, in-process instance with a shared state.

I briefly touched on this in my answer to Jose. I think null-ls is great as a service provider.

jose-elias-alvarez:
A counterargument: I recently ran across [lsp-format.nvim](https://github.com/lukas-reineke/lsp-format.nvim), an excellent plugin that significantly simplifies the process of setting up and running async formatting. In my limited testing, null-ls works out-of-the-box with this plugin, which is the whole reason null-ls exists: to take advantage of the LSP ecosystem without requiring other sources to handle null-ls as a special case. Merging this would require writing and maintaining a special null-ls integration, which isn't ideal from their end or ours.

mjlbach:
I don't understand the point of that plugin, the only reason I tell people to use formatting_sync is explicitly because of the `:wq` issue which that plugin does not solve (my PR to core does, btw https://github.com/neovim/neovim/pull/16206). In any case, duplicating that functionality is quite trivial.

jose-elias-alvarez:
The goal of this PR was to reduce conflicts between null-ls formatting and actual LSP formatting. I think `vim.lsp.buf.format` is a much nicer solution to that problem, so I'm going to close this PR.

PR #704, opened 2022-02-23
feat: run selene and tests in CI

stephank:
I was thinking about ways to get up-to-date versions of stylua, selene and neovim installed in CI, and could only think of Homebrew and Nix. I'm most familiar with Nix, so that's what I implemented here. Hope it's not too much voodoo. 🙃

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -37,6 +37,7 @@ jobs:
       - uses: rhysd/action-setup-vim@v1
         with:
           neovim: true
+          version: nightly
       - run: make test
         working-directory: null-ls.nvim
         timeout-minutes: 1
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -12,3 +12,31 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --color always --check .
+
+  selene:
+    name: Lint
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: NTBBloodbath/selene-action@v1.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          args: --color always ./lua
+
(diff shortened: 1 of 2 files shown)


polyzen:
If it'd help, selene and stylua could be added to the Arch Linux repos (currently they're in the AUR).

jose-elias-alvarez:
I'm a little hesitant to introduce Nix just because I've never used it before. I've seen other repos use [this GitHub action](https://github.com/rhysd/action-setup-vim) to set up Neovim–maybe we could give that a shot? Otherwise I'd have to try this out and make sure I'd be comfortable maintaining it in the future.

stephank:
That's okay, I adjusted it to use actions. I found [an action for Selene](https://github.com/NTBBloodbath/selene-action) before, but thought it was no longer updated. But it seems to work fine.

> If it'd help, selene and stylua could be added to the Arch Linux repos (currently they're in the AUR).

I think the only way to use those on GitHub Actions is via the `archlinux` Docker images? It looks like it'd be similar in download size to what I was doing with Nix, but less magic.

It's all preference, I guess. I think a benefit of the Nix or actions based approaches is that we could run tests on macOS. Haven't tried yet, because I don't think it's that important. (Saying this as a macOS user.)

jose-elias-alvarez:
I went ahead and pushed a commit to specify the Neovim version we want and everything seems to be working well. Selene took 4+ minutes to complete, which seems strange–is there anything we're doing wrong? Otherwise I'm happy to merge this.

stephank:
It uses the GitHub API to find release assets, so I guess there was a hiccup there, but hard to tell why exactly.

It should normally also prefer cache, but it's not doing that here. Possibly that's disabled for PR builds.

My best guess is these issues should be sporadic.

jose-elias-alvarez:
Got it, 4 minutes isn't too burdensome anyways, and hopefully it's sporadic. Thanks again for the contribution!

PR #707, opened 2022-02-24
wip: add xo diagnostics

mvllow:
So far unable to get this to work properly. Help/feedback is much appreciated. I'll continue playing with this and hopefully get something working here 😌

### DIFF
--- a/lua/null-ls/builtins/diagnostics/xo.lua
+++ b/lua/null-ls/builtins/diagnostics/xo.lua
@@ -1,0 +1,44 @@
+local h = require("null-ls.helpers")
+local cmd_resolver = require("null-ls.helpers.command_resolver")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local handle_xo_output = function(params)
+    params.messages = params.output and params.output[1] and params.output[1].messages or {}
+    if params.err then
+        table.insert(params.messages, { message = params.err })
+    end
+
+    local parser = h.diagnostics.from_json({
+        attributes = {
+            severity = "severity",
+        },
+        severities = {
+            h.diagnostics.severities["warning"],
+            h.diagnostics.severities["error"],
+        },
+    })
+
+    return parser({ output = params.messages })
+end
+
+return h.make_builtin({
+    name = "xo",
+    method = DIAGNOSTICS,
+    filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact" },
+    generator_opts = {
+        command = "xo",
+        args = { "--stdin", "--stdin-filename", "$FILENAME" },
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
These changes seem to make things work on my end:

```diff
diff --git a/lua/null-ls/builtins/diagnostics/xo.lua b/lua/null-ls/builtins/diagnostics/xo.lua
index 6e849ea..0903615 100644
--- a/lua/null-ls/builtins/diagnostics/xo.lua
+++ b/lua/null-ls/builtins/diagnostics/xo.lua
@@ -29,9 +29,9 @@ return h.make_builtin({
     filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact" },
     generator_opts = {
         command = "xo",
-        args = { "--stdin", "--stdin-filename", "$FILENAME" },
+        args = { "--reporter", "json", "--stdin", "--stdin-filename", "$FILENAME" },
         to_stdin = true,
-        -- format = "json_raw",
+        format = "json",
         check_exit_code = function(code)
             return code <= 1
         end,
```

I think that should work as a starting point. `xo` seems to be more different from ESLint than I expected, so I think it's a good idea to test this before we merge it. If we find that it's ultimately the same, we could set it up as a variant of `eslint`, kind of like what we do for [`eslint_d`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/eslint_d.lua).

I also see that `xo` provides rule fixes, so we could investigate whether to add code actions as well, like [ESLint](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/code_actions/eslint.lua). Again, if everything lines up with vanilla `eslint`, it would be simple to set this up as a variant.

mvllow:
Thanks for the help 😌 So now we have:

- [x] Diagnostics
- [x] Code actions (using "ESLint" verbiage, may want to change some to say XO?)
- [ ] Formatting (doesn't appear to be working with my initial attempt

I agree that extending ESLint should be fine. I'm curious why ESLint diagnostics/code actions use `format: "json_raw"` whereas formatting uses `format: "json"`. Other than that, we could simply overwrite filetypes, command, and args.

mvllow:
I'll look into the `--fix-dry-run` and try to extend eslint to reduce duplication. Appreciate the insight 😌

mvllow:
Removed formatting as `--fix` is destructive. All tests in `make test` pass locally. Please let me know if there's anything else that needs tested/completed.

jose-elias-alvarez:
Docs are still done manually, ideally in alphabetical order.

jose-elias-alvarez:
Excellent, glad we were able to get this solved with the minimal amount of code. Thanks for the contribution!

PR #708, opened 2022-02-25
Uncrustify argument split breaks extra_args

Toesmasher:
The space between "-l" and the language causes any extra_args to not be accepted, simply splitting it up solves the issue.


jose-elias-alvarez:
Was this not an issue in the end?

Toesmasher:
It was not an issue. I had a devil of a time passing arguments which looked completely fine in the debug log. What I thought was my fix turned out to be another change I made when calling setup. In other words, I'm a sillyface. :-)

Issue #711, opened 2022-02-25
add support for range formatting in autopep8


JoseConseco:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

autopep8 support --line-range form to parameter. I wish null ls would add support for range format for pep8.

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I copied code form `yapf` and converted it for autopep:

```lua
local h = require("null-ls.helpers")
local methods = require("null-ls.methods")

local FORMATTING = methods.internal.FORMATTING
local RANGE_FORMATTING = methods.internal.RANGE_FORMATTING

local function range_formatting_args_factory(base_args, start_arg)
    vim.validate({
        base_args = { base_args, "table" },
        start_arg = { start_arg, "string" },
    })

    return function(params)
        local args = vim.deepcopy(base_args)
        if params.method == FORMATTING then
            return args
        end

        local range = params.range
        table.insert(args, start_arg)
        table.insert(args, range.row .. " " .. range.end_row) -- range of lines to reformat, one-based
        return args
    end
end

return h.make_builtin({
    name = "autopep8",
    method = {FORMATTING, RANGE_FORMATTING},
    filetypes = { "python" },
    generator_opts = {
        command = "autopep8",
        args = range_formatting_args_factory({
            "-",
        }, "--line-range"),
        to_stdin = true,
    },
    factory = h.formatter_factory,
})
```
It work ok for non-range formatting. But it does nothing on ranged format. The problem seems to be  wrong `row_start` and 'row_end`  parameters - they are both set to 1. Debug log:

```

[TRACE pią, 25 lut 2022, 15:38:56] ...te/pack/packer/start/null-ls.nvim/lua/null-ls/client.lua:92: starting null-ls client
[TRACE pią, 25 lut 2022, 15:38:56] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:118: received LSP request for method initialize
[TRACE pią, 25 lut 2022, 15:38:56] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:143: received LSP notification for method initialized
[TRACE pią, 25 lut 2022, 15:38:56] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:143: received LSP notification for method textDocument/didOpen
[...]

PR #712, opened 2022-02-25
added support for range format in autopep8

JoseConseco:
added support for range format in autopep8 as mentioned in [this issue](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/711)


jose-elias-alvarez:
Thanks! Could we add a mention to the documentation that this supports range formatting?

JoseConseco:
@jose-elias-alvarez  ok, added docs info about range support.

jose-elias-alvarez:
Thanks!

PR #719, opened 2022-02-27
Include FromLinter in golangci-lint's source

shihanng:
The diagnostics from golangci-lint can be emitted from multiple linters. Including the value of FromLinter can help user identifies which linter emitted the specific message. Prior this change, a diagnostic message looks like:

```
don't use `init` function golangci-lint [15, 1]
```

With this change it will be:

```
don't use `init` function golangci-lint:gochecknoinits [15, 1]
```

Please let me know what you think or if there is anything I can do to improve this PR. Thank you.


jose-elias-alvarez:
This seems like a nice improvement, and I can see that the VS Code extension shows the same information (albeit in a slightly different way). Thanks!

PR #725, opened 2022-03-01
feat: add support for PHPMD

Geraint:


lua/null-ls/builtins/diagnostics/phpmd.lua
```diff
@@ -0,0 +1,48 @@
...
+    filetypes = { "php" },
+    generator_opts = {
+        command = "phpmd",
+        args = {
+            "--ignore-violations-on-exit",
+            "-", -- process stdin
+            "json",
+            -- 'phpmd.xml',
+        },
+        format = "json_raw",
```
jose-elias-alvarez, line 16: One small change: we should use `json` over `json_raw` if we don't plan on handling errors manually in `on_output` (which as far as I can tell is not the case).

jose-elias-alvarez:
Great, thanks!

PR #729, opened 2022-03-03
fix: prevent errors from killing PHPMD Language Server

Geraint:
When writing code, you will naturally have syntax errors (before you have finished writing a line of code, for example).

Without this change, such errors will cause an error in null-ls:

```
lua/null-ls/helpers/generator_factory.lua:229: error in generator output
```

The PHPMD language server will stop working:

- any existing diagnostic messages will remain on the screen, even if you fix the relevant problem
- no new diagnostic messages will be shown

I believe this is because `phpmd` is exiting with a non-zero status code when it encounters syntax errors.  This change stops it from doing that.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -2425,7 +2425,7 @@ Note that `extra_args` is required, and allows you so specify the
 
 - `filetypes = { "php" }`
 - `command = "phpmd"`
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Thanks for the fix!

Issue #734, opened 2022-03-06
Support diagnostics for bazel build / bzl files with buildifier linting


kgreenek:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

[Bazel](https://bazel.build/) is a popular build system.

[Buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier) is a popular tool for automatically formatting bazel build files and starlark files. It also supports linting and can generate the output warnings in a structured json format.

### Help

Yes

PR #735, opened 2022-03-06
Add buildifier linter diagnostics

kgreenek:
Closes https://github.com/jose-elias-alvarez/null-ls.nvim/issues/734

### DIFF
--- a/lua/null-ls/builtins/diagnostics/buildifier.lua
+++ b/lua/null-ls/builtins/diagnostics/buildifier.lua
@@ -1,0 +1,95 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local function rtrim(s)
+    local n = #s
+    while n > 0 and s:find("^%s", n) do
+        n = n - 1
+    end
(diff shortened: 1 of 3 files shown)


lua/null-ls/builtins/diagnostics/buildifier.lua
```diff
@@ -0,0 +1,95 @@
...
+    return diagnostics
+end
+
+local function parse_error(line)
+    local pattern = [[.-:(%d+):(%d+): (.*)]]
+    local results = { line:match(pattern) }
+    local row = tonumber(results[1])
+    local col = tonumber(results[2])
+    -- Remove trailing newline in the error message.
+    local message = rtrim(results[3])
```
jose-elias-alvarez, line 53: Would `vim.trim` achieve the same result here?

jose-elias-alvarez (commented):
Maybe one small optimization we can make but other than that this looks good.

lua/null-ls/builtins/diagnostics/buildifier.lua
kgreenek: Nice! Looks like it does. I just didn't know about that function :)

jose-elias-alvarez:
Great, thanks!

PR #740, opened 2022-03-06
ci: automatically generate built-in documentation

### LINKED ISSUE #552: [RFC] Parsable metadata for auto-generated docs and interop
### Issues

### Feature description

As mentioned in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/397, 
> We already have an issue with built-in documentation falling out of sync, so I've been thinking about automatically generating that, too.

I think we'd need some sort of manifest, containing the rest of the hard-coded information that is mentioned in [BUILTINS](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md), such as: description, URL, supported features(?)

Another interesting part would be exposing how that source can be installed, for example: `npm install markdownlint-cli`. This is pretty useful to have interop with https://github.com/williamboman/nvim-lsp-installer which offers a APIs that should be able to handle most common installation methods, @williamboman 👋 

### Help

### Implementation help

I'm not sure what the best presentation of that manifest should look like. The simplest solution is probably a basic json/lua table with desired entries, à la NoSQL.

What do you guys think?

jose-elias-alvarez:
Partially closes #552.

This is an early attempt at adding metadata to built-ins and using that metadata to automatically generate documentation. The goal is to have each built-in be a self-contained source of truth so that contributors can more easily add / update them without worrying about keeping documentation up-to-date. 

As part of this change, BUILTINS.md is now split into two files: a manually written BUILTIN_CONFIG.md file that describes setting up and configuring built-ins, and an automatically generated BUILTINS.md file containing only a list of built-ins and their metadata, parsed into a Markdown document that more or less resembles the old, manually written version.

For now, I've only started to add metadata to one built-in as a test case. Converting metadata is a pretty gigantic task in and of itself, and I'm sure going through the process will expose new use cases / edge cases to handle.

Note: I haven't thought much about the interop aspect discussed in #552. I'm happy to consider other approaches and update this PR as needed, but I consider the documentation issue significant enough to prioritize this for the moment.

- [ ] Make sure "derivative" built-ins are correctly handled
- [x] Signal built-in methods more accurately (diagnostics on save, range formatting)
- [ ] Handle other use cases / test cases from the current file


jose-elias-alvarez:
@kylo252 Forgot to tag you on this but let me know if you have any feedback. I figure with the existing metadata, we can reference built-ins by name and load the metadata from the file, and we can add more as we go, but let me know if you think this approach might cause trouble in the future.

kylo252:
@jose-elias-alvarez, this is still inline with the interop discussion, great work! 🚀 

All that's missing is a section similar to what @williamboman suggested here https://github.com/jose-elias-alvarez/null-ls.nvim/issues/552#issuecomment-1008323291.

jose-elias-alvarez:
I think this is good to go, though since the process was mostly manual I'm sure there were some typos / errors (still better than the manual documentation).

PR #742, opened 2022-03-10
Add rstcheck diagnostics

ObserverOfTime:
I've configured it to run on the entire workspace but it can work just as well on a single file.


jose-elias-alvarez:
Great, this looks good too, thanks again!

Issue #691, opened 2022-02-17
Eslint_d error -  File ignored by default


lucasmoura-boticario:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.6.1

### Operating System

MacOSX  12.0.1

### Minimal config

```
local null_ls = require('null-ls')

null_ls.setup({
    debug = true,
    sources = {
      null_ls.builtins.diagnostics.eslint_d.with({
      extra_args = {"--stdin-filename", ".eslintrc.js"}
      }), -- eslint or eslint_d
      null_ls.builtins.formatting.eslint -- prettier, eslint, eslint_d, or prettierd
    },
})

null_ls.setup()

```

### Steps to reproduce

Diagnostic log

```
[TRACE Qui 17 Fev 16:58:34 2022] ...ocal/share/nvim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:118: received LSP request for method shutdown
[TRACE Qui 17 Fev 16:58:34 2022] ...ocal/share/nvim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:143: received LSP notification for method exit
[TRACE Qui 17 Fev 16:58:41 2022] ...l/share/nvim/plugged/null-ls.nvim/lua/null-ls/client.lua:92: starting null-ls client
[TRACE Qui 17 Fev 16:58:41 2022] ...ocal/share/nvim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:118: received LSP request for method initialize
[TRACE Qui 17 Fev 16:58:41 2022] ...ocal/share/nvim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:143: received LSP notification for method initialized
[TRACE Qui 17 Fev 16:58:41 2022] ...ocal/share/nvim/plugged/null-ls.nvim/lua/null-ls/rpc.lua:143: received LSP notification for method textDocument/didOpen
[TRACE Qui 17 Fev 16:58:41 2022] ...are/nvim/plugged/null-ls.nvim/lua/null-ls/generators.lua:21: running generators for method NULL_LS_DIAGNOSTICS_ON_OPEN
[...]

Issue #722, opened 2022-02-28
`black` not respecting the local `pyproject.toml`


scallaway:
Basically as per the title, I'm unable to format my files with `null_ls_module.builtins.formatting.black` based on the `pyproject.toml` file that is located in my project root directory.

I am, however, able to call `:! black %` and have it format respecting the `pyproject.toml` file.

The only other piece of information on this I could find (since there was nothing in this repo's issues or discussions) was this reddit post here; https://www.reddit.com/r/neovim/comments/s2o8ww/format_and_sort_on_save_with_lsp_and_nullls/ - this seems to be basically the same problem.

---

On a side note, I have a feeling that the same thing is happening with `isort`, since it reports that I have imports that are out of order, even _after_ I call `black` manually.

Is this because of the location in which the processes are spawned? And so since it's not in the directory it in theory _should_ be, it's not able to find the config file and is reverting back to the defaults?

---

I don't have the time at the moment to test with a minimal config, but I wanted to at least open this point to get any other voices on the matter.

karljs:
This might be a duplicate of #700 where `black` just isn't running as expected at least some of the time, which I was able to reproduce there.

TortoiseHive:
Same here

jose-elias-alvarez:
I'm closing this issue because it lacks the information I need to investigate and act on it. Feel free to open a new issue with reproduction instructions if you're still having this issue.

Jackevansevo:
For me this issue is fixed by the suggestion in this thread https://github.com/jose-elias-alvarez/null-ls.nvim/issues/848#issuecomment-1119040548


- https://github.com/jose-elias-alvarez/null-ls.nvim/issues/848

I.e. 


```lua
    require('null-ls').builtins.diagnostics.black.with {
      cwd = function(params)
        return vim.fn.fnamemodify(params.bufname, ':h')
      end,
    },
```

scallaway:
> For me this issue is fixed by the suggestion in this thread [#848 (comment)](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/848#issuecomment-1119040548)

Can confirm that this does indeed fix my issue!

PR #745, opened 2022-03-10
feat: add root_markers to help detect config files

### LINKED ISSUE #538: Can't override flake8 errors to warnings

kylo252:
# Description

Add `root_resolver.lua` to help with set `cwd` from _root-markers_. This helps with detecting common configuration files, e.g. `pyproject.toml`.
Additionally, this should be easily configurable since `root_markers.FOO` is global, and `cwd` is still overridable as well.

Fixes #538, #722

### DIFF
--- a/lua/null-ls/helpers/command_resolver.lua
+++ b/lua/null-ls/helpers/command_resolver.lua
@@ -1,6 +1,7 @@
 local log = require("null-ls.logger")
 local s = require("null-ls.state")
 local u = require("null-ls.utils")
+local root_resolver = require("null-ls.helpers.root_resolver")
 local fmt = string.format
 
 local M = {}
@@ -50,7 +51,15 @@ M.from_node_modules = function(params)
     -- try the local one first by default but always fallback on the pre-defined command
     -- this needs to be done here to avoid constant lookups
     -- we're checking if the global is executable to avoid spawning an invalid command
-    return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
+    local relative_cmd = M.generic(params, u.path.join("node_modules", ".bin"))
+    local command = relative_cmd or u.is_executable(params.command) and params.command
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    if not resolved.cwd then
+        -- attempt to set the cwd from other node markers
+        resolved.cwd = root_resolver.from_node_markers
+        s.set_resolved_command(params.bufnr, params.command, resolved)
+    end
+    return command
 end
 
 M.from_yarn_pnp = function(params)
@@ -81,6 +90,10 @@ M.from_yarn_pnp = function(params)
     end
 
     resolved = resolved or { command = false }
+    if not resolved.cwd then
+        -- attempt to set the cwd from other node markers
+        resolved.cwd = root_resolver.from_node_markers
+    end
 
     s.set_resolved_command(params.bufnr, cache_key, resolved)
     return resolved.command
--- a/lua/null-ls/helpers/root_resolver.lua
+++ b/lua/null-ls/helpers/root_resolver.lua
@@ -26,7 +26,7 @@ M.from_python_markers = function(params)
     if cwd and cwd ~= "" then
         log:debug(fmt("Resolved python project path: [%s]", cwd))
     else
-        log:debug("Unable to resolved python project path")
+        log:debug("Unable to resolve python project path")
     end
     return cwd
 end
@@ -34,7 +34,11 @@ end
 M.from_node_markers = function(params)
     local cwd = u.root_pattern(M.root_markers.node)(params.bufname)
     log:trace(fmt("Using node markers: [%s]", M.root_markers.node))
-    log:debug(fmt("Resolved node project path: [%s]", cwd or ""))
+    if cwd and cwd ~= "" then
+        log:debug(fmt("Resolved node project path: [%s]", cwd))
+    else
+        log:debug("Unable to resolve node project path")
+    end
     return cwd
 end
 
--- a/lua/null-ls/helpers/root_resolver.lua
+++ b/lua/null-ls/helpers/root_resolver.lua
@@ -1,0 +1,41 @@
+local log = require("null-ls.logger")
+local u = require("null-ls.utils")
+local fmt = string.format
+
+local M = {}
+
+M.root_markers = {
+    python = {
+        "pyproject.toml",
+        "setup.cfg",
+        "tox.ini",
+        ".flake8",
+    },
+    node = {
+        ".eslintrc*",
+        "package.json",
+        "tsconfig.json",
+        "jsconfig.json",
+        ".prettierrc*",
+    },
+}
+
+M.from_python_markers = function(params)
+    local cwd = u.root_pattern(M.root_markers.python)(params.bufname)
+    log:trace(fmt("Using python markers: [%s]", M.root_markers.python))
+    if cwd and cwd ~= "" then
+        log:debug(fmt("Resolved python project path: [%s]", cwd))
+    else
+        log:debug("Unable to resolved python project path")
+    end
+    return cwd
+end
+
+M.from_node_markers = function(params)
+    local cwd = u.root_pattern(M.root_markers.node)(params.bufname)
+    log:trace(fmt("Using node markers: [%s]", M.root_markers.node))
+    log:debug(fmt("Resolved node project path: [%s]", cwd or ""))
+    return cwd
+end
+
+return M
--- a/lua/null-ls/builtins/diagnostics/flake8.lua
+++ b/lua/null-ls/builtins/diagnostics/flake8.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local root_resolver = require("null-ls.helpers.root_resolver")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
@@ -62,6 +63,7 @@ return h.make_builtin({
         to_stdin = true,
         from_stderr = true,
         args = { "--format", "default", "--stdin-display-name", "$FILENAME", "-" },
+        cwd = root_resolver.from_python_markers,
         format = "line",
         check_exit_code = function(code)
             return code == 0 or code == 255
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local root_resolver = require("null-ls.helpers.root_resolver")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
@@ -29,6 +30,7 @@ benefits of dynamic (or "duck") typing and static typing.]],
                 params.bufname,
             }
         end,
+        cwd = root_resolver.from_python_markers,
         to_temp_file = true,
         format = "line",
         check_exit_code = function(code)
--- a/lua/null-ls/builtins/diagnostics/pydocstyle.lua
+++ b/lua/null-ls/builtins/diagnostics/pydocstyle.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local root_resolver = require("null-ls.helpers.root_resolver")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
@@ -18,6 +19,7 @@ return h.make_builtin({
         command = "pydocstyle",
         name = "pydocstyle",
         args = { "$FILENAME" },
+        cwd = root_resolver.from_python_markers,
         to_stdin = false,
         to_temp_file = true,
         format = "raw",
--- a/lua/null-ls/builtins/diagnostics/pylama.lua
+++ b/lua/null-ls/builtins/diagnostics/pylama.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local root_resolver = require("null-ls.helpers.root_resolver")
 
 return h.make_builtin({
     name = "pylama",
@@ -16,6 +17,7 @@ return h.make_builtin({
         from_stderr = false,
         ignore_stderr = true,
         args = { "--from-stdin", "$FILENAME", "-f", "json" },
+        cwd = root_resolver.from_python_markers,
         format = "json_raw",
         check_exit_code = function(code)
             return code <= 1
--- a/lua/null-ls/builtins/diagnostics/pylint.lua
+++ b/lua/null-ls/builtins/diagnostics/pylint.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local root_resolver = require("null-ls.helpers.root_resolver")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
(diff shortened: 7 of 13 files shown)


kylo252:
This might potentially fix #691 as well, but I wasn't sure if we should add this when `dynamic_command` is already defined.

@jose-elias-alvarez, what do you think?

lua/null-ls/helpers/root_resolver.lua
```diff
@@ -0,0 +1,37 @@
...
+        "tsconfig.json",
+        "jsconfig.json",
+        ".prettierrc*",
+    },
+}
+
+M.from_python_markers = function(params)
+    local cwd = u.root_pattern(M.root_markers.python)(params.bufname)
+    log:trace(fmt("Using python markers: [%s]", M.root_markers.python))
+    log:debug(fmt("Resolved python project path: [%s]", cwd or ""))
```
jose-elias-alvarez, line 26: I feel like this log could be a little clearer when it fails to resolve.

lua/null-ls/helpers/root_resolver.lua
```diff
@@ -0,0 +1,37 @@
+local log = require("null-ls.logger")
+local u = require("null-ls.utils")
+local fmt = string.format
+
+local M = {}
+
+M.root_markers = {
```
jose-elias-alvarez, line 7: Just a thought, but I wonder if we could add a helper to add more root markers? Users could override `cwd` independently but it might be a nice addition if we can do it cleanly.

jose-elias-alvarez (commented):
I think this is a nice addition that could solve some issues with minimal user intervention. 

For #691: I noticed this recently, but the current `node_modules` resolver caches a value for `cwd` based on the path to resolved executable, but the `cwd` isn't actually passed to the command. I wonder if solving that would solve the issue.

lua/null-ls/helpers/root_resolver.lua
kylo252: Just in case this wasn't clear, you can already change these since they're global

```lua
require("null-ls.helpers.root_resolver").root_markers.python = {"README.md"}
```

But I guess you meant having an API for it? maybe adding it in the generic config section would work? since these are set by language (at least that's my intention).

---

Re #691: the problem might be from people using global installations, so maybe we need to enforce this regardless? I believe even pnp might mess with that for mono repos https://github.com/jose-elias-alvarez/null-ls.nvim/blob/a1804de23ce354c982aa08c57d3ed89aad8a15a9/lua/null-ls/helpers/command_resolver.lua#L68-L69

lua/null-ls/helpers/root_resolver.lua
kylo252: I assume you mean that it should be different if `cmd = nil`, correct?

jose-elias-alvarez: Yep, when `cwd = nil` I think the message could be a little clearer.

lua/null-ls/helpers/root_resolver.lua
jose-elias-alvarez: I thought about this more, and I think it's already pretty straightforward to add / override patterns, so I don't think we need a dedicated API. I'm still not sure about what's causing #691, so we can deal with it separately, too.

kylo252: @jose-elias-alvarez, sorry for the delay! what do you think about [`fb1ce9f` (#745)](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/745/commits/fb1ce9f4d8d3826788c6430173d268d866c50614)?

jose-elias-alvarez: I think this change makes sense, but at the moment, I believe the resolved `cwd` is actually not getting picked up anywhere. d76b41429cba092673db0051d24ed80f9a1649f2 added some basic handling to allow `command` to be a table, so perhaps it's enough to just do this [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/4cbdc9ff36cb7d527ea49abf2e3fe4100e1a4e35/lua/null-ls/helpers/generator_factory.lua#L298)?

```diff
diff --git a/lua/null-ls/helpers/generator_factory.lua b/lua/null-ls/helpers/generator_factory.lua
index 7337816..5b89839 100644
--- a/lua/null-ls/helpers/generator_factory.lua
+++ b/lua/null-ls/helpers/generator_factory.lua
@@ -275,9 +275,9 @@ return function(opts)

             params.command = command

-            local resolved_command
+            local resolved_command, resolved_cwd
             if dynamic_command then
-                resolved_command = dynamic_command(params)
+                resolved_command, resolved_cwd = dynamic_command(params)
                 log:debug(
                     string.format(
                         "Using dynamic command for [%s], got: %s",
@@ -295,7 +295,10 @@ return function(opts)
                 return done()
             end

-            local resolved_cwd = cwd and cwd(params) or root
+            -- cwd should fall back to root
+            if not resolved_cwd then
+                resolved_cwd = cwd and cwd(params) or root
+            end
             params.cwd = resolved_cwd

             local spawn_opts = {
```

The command resolver function can then return a tuple. Just an idea, and I'm happy to break it out into another PR, but I just wanted to point out that I don't think the change will have an effect until this is resolved.

kylo252: > I think this change makes sense, but at the moment, I believe the resolved `cwd` is actually not getting picked up anywhere. [d76b414](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/d76b41429cba092673db0051d24ed80f9a1649f2) added some basic handling to allow `command` to be a table, so perhaps it's enough to just do this [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/4cbdc9ff36cb7d527ea49abf2e3fe4100e1a4e35/lua/null-ls/helpers/generator_factory.lua#L298)?

I was counting on it being read from here
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/041601cb03daa8982c5af6edc6641f4b97e9d6b5/lua/null-ls/helpers/make_builtin.lua#L87-L91

so it should be set before calling the generator factory, but maybe that's not the case?

jose-elias-alvarez: I think this only applies to `prefer_local` / `only_local`, meaning it won't apply to the built-ins using the `from_node_modules` command resolver. Another option is to also override `cwd` for these, though that adds a little bit of boilerplate (since I imagine we will always want this behavior).

lua/null-ls/helpers/root_resolver.lua
```diff
@@ -0,0 +1,45 @@
...
+    node = {
+        ".eslintrc*",
+        "package.json",
+        "tsconfig.json",
+        "jsconfig.json",
+        ".prettierrc*",
+    },
+}
+
+M.from_python_markers = function(params)
```
jose-elias-alvarez, line 23: If we use this as-is, it won't get cached, right? If we're going to use this pattern we most likely want to create a generic `cwd` wrapper that caches to prevent lookups.

lua/null-ls/helpers/command_resolver.lua
```diff
@@ -50,7 +51,15 @@ M.from_node_modules = function(params)
     -- try the local one first by default but always fallback on the pre-defined command
     -- this needs to be done here to avoid constant lookups
     -- we're checking if the global is executable to avoid spawning an invalid command
-    return M.generic(params, u.path.join("node_modules", ".bin")) or u.is_executable(params.command) and params.command
+    local relative_cmd = M.generic(params, u.path.join("node_modules", ".bin"))
+    local command = relative_cmd or u.is_executable(params.command) and params.command
+    local resolved = s.get_resolved_command(params.bufnr, params.command)
+    if not resolved.cwd then
+        -- attempt to set the cwd from other node markers
+        resolved.cwd = root_resolver.from_node_markers
```
jose-elias-alvarez, line 59: I mentioned this under `from_python_markers`, but I think this also needs to be wrapped to prevent running a lookup on each run. I also just noticed that this is a reference to a function, so if we do pick this up as the 2nd return value from `dynamic_command`, we need to handle it.

jose-elias-alvarez (commented):
After this PR, I want to spend a little time cleaning up this API, since I think it's suffering from my lack of planning. For now, I think as long as we are caching the resolved `cwd` output and are picking up the `from_node_modules` `cwd` *somewhere*, we're good.

jose-elias-alvarez:
I added a helper in 490a09c0ccfc8da22f6840d6a03398ef5d1b3035 to simplify the process of caching return values that won't change over a buffer's lifetime, which might simplify this PR and make it easier for us to implement better defaults. Let me know if there's anything I can do to help with this, since I think this is a good way to improve our defaults.

zdog234:
Apologies if this is something that most neovim users would know, but could we add something to the documentation explaining how one would use this? (btw, this would be a really awesome feature :) )

jose-elias-alvarez:
> Apologies if this is something that most neovim users would know, but could we add something to the documentation explaining how one would use this? (btw, this would be a really awesome feature :) )

This is already possible to do using the [`cwd` config option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md#spawn-directory) + the cache helper added in 490a09c0ccfc8da22f6840d6a03398ef5d1b3035, though you'll have to wire it together yourself. I think this PR is still valuable in that it makes it easier to provide sensible defaults, which is a major goal of this project, but the core functionality is already there.

kylo252:
> > Apologies if this is something that most neovim users would know, but could we add something to the documentation explaining how one would use this? (btw, this would be a really awesome feature :) )
> 
> This is already possible to do using the [`cwd` config option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md?rgh-link-date=2022-08-15T15%3A46%3A00Z#spawn-directory) + the cache helper added in [490a09c](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/490a09c0ccfc8da22f6840d6a03398ef5d1b3035), though you'll have to wire it together yourself. 

It's been pretty rocky trying to get `cwd` to behave in a coherent manner, imo.

Here's a simple use-case: 
- some projects/plugins use `luacheck` and some of them don't
- if I'm switching between multiple directories, then `luacheck` should only spawn if the project uses it

the problem is that `cwd` isn't validating when changing the directory completely, instead you'd always need `runtime_condition`, but here's the confusing part: _it's actually enough to set a `runtime_condition` without even altering `cwd`.._ 🤯 

try it for yourself

```lua
local sources = {
  null_ls.builtins.diagnostics.luacheck.with {
    command = "luacheck",
    extra_args = {},
    filetypes = { "lua" },

    -- force luacheck to find its '.luacheckrc' file
    cwd = function(params)
      -- falls back to root if return value is nil
      return params.root:match ".luacheckrc"
    end,

    -- cwd = nls_cache.by_bufnr(function(params)
    --   return root_pattern ".luacheckrc" (params.bufname)
    -- end),

    -- runtime_condition = nls_cache.by_bufnr(function(params)
    --   return path.exists(path.join(params.root, ".luacheckrc"))
    -- end),
  },
}
```

> I think this PR is still valuable in that it makes it easier to provide sensible defaults, which is a major goal of this project, but the core functionality is already there.

I've been pushing this for a while now until I'm able to figure out how to combine those two aforementioned behaviors, but sadly I think I'm stuck, and the only way I can think of to fix this is to actually implement workspaces, and potentially let each source act as its own LS.

jose-elias-alvarez:
After merging #1063. the `command_resolver` structure has changed to *only* return a command and not mess with `cwd` at all. I think the ideal approach is to add `cwd` markers to specific built-ins and not for languages in general. 

I'm still having a hard time wrapping my head around the exact issue you describe with `cwd` and `runtime_condition`, but feel free to open another issue / PR.

PR #750, opened 2022-03-12
Add stylint diagnostics

ObserverOfTime:
An alternative to stylelint for the stylus preprocessor.


jose-elias-alvarez:
Thanks for the contributions! I just merged a PR to automatically generate documentation from built-in metadata, which should make things easier. Would you mind rebasing your PRs against `main` and moving the documentation to a `meta` key for the built-in? [This example](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/3f065b808663f598ad1f9db659ed11f90f05860a/lua/null-ls/builtins/diagnostics/eslint.lua#L28) should hopefully illustrate what I mean, but let me know if it's unclear.

jose-elias-alvarez:
Thanks for the quick turnaround, and thanks again for all the contributions!

Issue #754, opened 2022-03-13
New `tidy` diagnostics source does not account for XML files correctly


andrewferrier:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.7.0-dev+1186-g83fc91433

### Operating System

Linux

### Minimal config

N/A

### Steps to reproduce

* Use null-ls commit 4cbdc9ff36c
* Enable the new `tidy` diagnostics built-in
* Load any valid `.xml` file.
* See lots of irrelevant diagnostics that only apply 

### Expected behavior

Only get errors from `tidy` that are relevant to `.xml` files

### Actual behavior

As above.

### Debug log

X

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I think the problem here is that the `-xml` argument needs to be added to `tidy` to get it to validate XML, not HTML files. It looks like the source wasn't tested on XML files.

Is there a way to vary the arguments dynamically on the built-in, or is another source needed?

### Requirements

- [X] I have read and followed the instructions above and understand that my issue will be closed if I did not provide the required information.

jose-elias-alvarez:
Sure, it's possible to alter a built-in's arguments based on the file type of the buffer it's running on, something like this:

```lua
args = function(params)
    return params.ft == "xml" and { "xml", "args" } or { "default", "args" }
end
```

It sounds like that might be the fix for this case. PR welcome!

jose-elias-alvarez:
Closed by #755, thanks again for catching this and fixing it @andrewferrier!

PR #755, opened 2022-03-13
fix(builtins): Add `-xml` parameter for XML files

andrewferrier:
Ensures that tidy diagnostics appropriate to XML files are shown, rather than HTML files.

Should close issue https://github.com/jose-elias-alvarez/null-ls.nvim/issues/754.


jose-elias-alvarez:
Excellent, thanks for putting this in!

PR #771, opened 2022-03-18
Add ruleId to eslint adapter

benwainwright:
Without the ruleid in the diagnostic, this plugin is quite hard to use with eslint...

### DIFF
--- a/lua/null-ls/builtins/diagnostics/eslint.lua
+++ b/lua/null-ls/builtins/diagnostics/eslint.lua
@@ -15,7 +15,6 @@ local add_rule_id_to_messages = function(messages)
 end
 
 local handle_eslint_output = function(params)
-    print(vim.inspect(params))
     params.messages = params.output and params.output[1] and params.output[1].messages or {}
     if params.err then
         table.insert(params.messages, { message = params.err })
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
I don't think this should be a default, as it's inconsistent with other sources. You can already change the format of diagnostic messages from a specific linter using [the method described in the documentation](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#eslint).

benwainwright:
@jose-elias-alvarez if its inconsistent with other sources then they should be changed to. This is a massive omission and serious usability issue - its pretty much unusable without it because most people respond to a linter error by googling the rule id if its in any way confusing (which is a lot)

benwainwright:
@jose-elias-alvarez are you serious? This plugin is completely unusable for eslint without this. If you even slightly use a linter you will find yourself googling rules based on the rule id _all the time_. If its inconsistent with other sources then the other sources should be changed...

benwainwright:
@jose-elias-alvarez please reconsider. I don't see anything in the documentation you linked to which would provide the behaviour I'm after

benwainwright:
What I'm proposing is pretty standard. E.g. https://github.com/creativenull/diagnosticls-configs-nvim/blob/eed632a079064bdf415f1f9e128998da3b61bff0/lua/diagnosticls-configs/linters/eslint.lua#L14 (I really don't want to use `diagnosticls` though)

jose-elias-alvarez:
Apologies, I pasted the wrong link. [Here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md#diagnostics-format) is the relevant page.

PR #772, opened 2022-03-18
fix(diagnostics): Compatibility with older versions of PHPMD work

Geraint:
Due to [a bug](https://github.com/phpmd/phpmd/issues/941) in PHPMD, STDIN doesn't work on the current version unless you explicitly update its dependency, PDepend. (see https://github.com/pdepend/pdepend/pull/593).

This update to the builtin makes it use a temp file instead, meaning it will work with PHPMD without requiring the PDepend fix.

I have also removed the `--ignore-errors-on-exit` argument, because it isn't supported on older versions of PHPMD and we can instead handle non-zero exit codes in `check_exit_code` .

### DIFF
--- a/lua/null-ls/builtins/diagnostics/phpmd.lua
+++ b/lua/null-ls/builtins/diagnostics/phpmd.lua
@@ -11,18 +11,11 @@ return h.make_builtin({
     filetypes = { "php" },
     generator_opts = {
         command = "phpmd",
-        args = {
-            "--ignore-violations-on-exit",
-            "--ignore-errors-on-exit",
-            "-", -- process stdin
-            "json",
-            -- 'phpmd.xml',
-        },
+        args = { "$FILENAME", "json" },
         format = "json",
-        to_stdin = true,
-        from_stderr = false,
+        to_temp_file = true,
         check_exit_code = function(code)
-            return code <= 1
+            return code <= 2
         end,
         on_output = function(params)
             local parser = h.diagnostics.from_json({
@@ -37,6 +30,7 @@ return h.make_builtin({
                     h.diagnostics.severities["error"],
                     h.diagnostics.severities["warning"],
                     h.diagnostics.severities["information"],
+                    h.diagnostics.severities["hint"],
                     h.diagnostics.severities["hint"],
                 },
             })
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/phpmd.lua
```diff
@@ -38,6 +31,7 @@ return h.make_builtin({
                     h.diagnostics.severities["warning"],
                     h.diagnostics.severities["information"],
                     h.diagnostics.severities["hint"],
+                    h.diagnostics.severities["hint"],
```
jose-elias-alvarez, line 34: Is this a typo?

jose-elias-alvarez (commented):
I'm okay with this change for now. Generally, we want to avoid temp file usage wherever we can, since it adds overhead (especially for linters that run so frequently) so I hope we can remove this in the future.

lua/null-ls/builtins/diagnostics/phpmd.lua
Geraint: No, although it's possible I've misunderstood how that works.

PHPMD has 5 different severities:

> priority: The priority for the rule. This can be a value in the range 1-5, where 1 is the highest priority and 5 the lowest priority.

source: https://phpmd.org/documentation/writing-a-phpmd-rule.html

I set severities with `hint` twice because I *think* this means that rules with priority `4` *or* `5` will both equate to a severity of `hint`.

If I have misunderstood how that works, then please let me know.

As it happens, whilst looking up the documentation for that, I discovered that `3` is an exit code, too, so I've updated `check_exit_code` accordingly.

See the section "Exit codes" at https://phpmd.org/documentation/index.html.

jose-elias-alvarez: You're right, I scanned this too quickly and didn't understand at first glance. Having said that, maybe we could make this a bit more explicit to show that it's a map?

```lua
severities = {
    [1] = h.diagnostics.severities["information"],
    [2] = h.diagnostics.severities["warning"],
    [3] = h.diagnostics.severities["error"],
    [4] = h.diagnostics.severities["hint"],
    [5] = h.diagnostics.severities["hint"],
}
```

Geraint: OK, I have made that change and it still seems to work for me locally.

jose-elias-alvarez: Great, I think it helps readability a bit. Thanks for the PR!

Issue #768, opened 2022-03-16
Shellcheck code_action break multi-line commands


jvanbruegge:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.6.1

### Operating System

NixOS (Linux)

### Minimal config

```
null_ls.setup({
        sources = { null_ls.builtins.code_actions.shellcheck },
        debug = true,
    })
```

### Steps to reproduce

Create file with content:
```
#!/usr/bin/env bash

FOO=$()
echo \
  $FOO
```
1. `nvim --clean -u minimal_init.lua script.sh`
2. Navigate to `$FOO` and activate the code action
3. Select `Disable ShellCheck rule 2086 for this line`

### Expected behavior

```
# shellcheck disable=2086
echo \
  $FOO
```

### Actual behavior

```
echo \
  # shellcheck disable=2086
  $FOO
```

### Debug log

""

### Help

No

### Implementation help

The row number [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/code_actions/shellcheck.lua#L81) probably needs to check for the first line that does not end with a backslash.

### Requirements

- [X] I have read and followed the instructions above and understand that my issue will be closed if I did not provide the required information.

jose-elias-alvarez:
Your suggestion seems reasonable. PR welcome!

jose-elias-alvarez:
Fixed by #775.

PR #775, opened 2022-03-20
fix(code_actions): search for first non-command line for disable pragma

jvanbruegge:
I have confirmed that this works as expected

ISSUES_FIXED: #768


jose-elias-alvarez:
Seems reasonable, thanks for identifying the issue and for the fix!

b0o:
Nice catch! One thing, this doesn't seem to handle escaped backslashes:

```sh
echo "foo" \\
echo "bar" $foobar
```

Expected:

```sh
echo "foo" \\
# shellcheck disable=2086
echo "bar" $foobar
```

Actual:

```sh
# shellcheck disable=2086
echo "foo" \\
echo "bar" $foobar
```

PR #777, opened 2022-03-23
feat: add buf linter and formater

Cluas:


lua/null-ls/builtins/diagnostics/buf.lua
```diff
@@ -0,0 +1,31 @@
...
+
+local DIAGNOSTICS_ON_SAVE = methods.internal.DIAGNOSTICS_ON_SAVE
+
+return h.make_builtin({
+    name = "buf_lint",
+    meta = {
+        url = "https://github.com/bufbuild/buf",
+        description = "A new way of working with Protocol Buffers.",
+    },
+    method = DIAGNOSTICS_ON_SAVE,
```
jose-elias-alvarez, line 12: I always want to check this, but: Is there a specific reason for this linter to run on save only? It shouldn't be necessary if it accepts `stdin`.

jose-elias-alvarez (commented):
Looks good! Just want to clarify if it's possible to run this on save, but if not let me know.

lua/null-ls/builtins/diagnostics/buf.lua
Cluas: Your are right, indeed there is no effect, directly using DIAGNOSTICS

jose-elias-alvarez:
Excellent, thank you for the contribution! For future reference, the documentation is generated automatically on merge, so no need to do it manually.

PR #784, opened 2022-03-27
feat: Add sqlfluff as sql format and diag source

KingMichaelPark:
Add sqlfluff as a diagnosis and formatting option for sql files. (Thanks again for the help on the question answered the other day).

### DIFF
--- a/lua/null-ls/builtins/diagnostics/sqlfluff.lua
+++ b/lua/null-ls/builtins/diagnostics/sqlfluff.lua
@@ -1,0 +1,45 @@
+local null_ls = require("null-ls")
+local helpers = require("null-ls.helpers")
+
+return helpers.make_builtin({
+    name = "sqlfluff",
+    meta = {
+        url = "https://github.com/sqlfluff/sqlfluff",
+        description = "A SQL linter and auto-formatter for Humans",
+    },
+    method = null_ls.methods.DIAGNOSTICS,
+    filetypes = { "sql" },
+    generator_opts = {
+        command = "sqlfluff",
+        args = {
+            "lint",
+            "-f",
+            "github-annotation",
+            "-n",
+            "--disable_progress_bar",
+            "-",
+        },
+        from_stderr = true,
+        to_stdin = true,
+        format = "json",
+        check_exit_code = function(c)
+            return c <= 1
+        end,
+        on_output = helpers.diagnostics.from_json({
+            attributes = {
+                row = "line",
+                col = "start_column",
+                end_col = "end_column",
+                severity = "annotation_level",
+                message = "message",
+            },
+            severities = {
+                helpers.diagnostics.severities["information"],
+                helpers.diagnostics.severities["warning"],
+                helpers.diagnostics.severities["error"],
+                helpers.diagnostics.severities["hint"],
+            },
+        }),
+    },
+    factory = helpers.generator_factory,
+})
--- a/lua/null-ls/builtins/formatting/sqlfluff.lua
+++ b/lua/null-ls/builtins/formatting/sqlfluff.lua
@@ -1,0 +1,27 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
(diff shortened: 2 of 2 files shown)


lua/null-ls/builtins/diagnostics/sqlfluff.lua
```diff
@@ -0,0 +1,41 @@
...
+        description = "A SQL linter and auto-formatter for Humans",
+    },
+    method = null_ls.methods.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "sql" },
+    to_ = true,
+    generator_opts = {
+        async = true,
+        command = "sqlfluff",
+        args = { "lint", "-f", "github-annotation", "-n", "--disable_progress_bar", "$FILENAME" },
+        -- choose an output format (raw, json, or line)
```
jose-elias-alvarez, line 17:
```suggestion
```

lua/null-ls/builtins/diagnostics/sqlfluff.lua
```diff
@@ -0,0 +1,41 @@
...
+
+return helpers.make_builtin({
+    name = "sqlfluff",
+    meta = {
+        url = "https://github.com/sqlfluff/sqlfluff",
+        description = "A SQL linter and auto-formatter for Humans",
+    },
+    method = null_ls.methods.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "sql" },
+    to_ = true,
```
jose-elias-alvarez, line 12:
```suggestion
```

lua/null-ls/builtins/diagnostics/sqlfluff.lua
```diff
@@ -0,0 +1,41 @@
+local null_ls = require("null-ls")
+local helpers = require("null-ls.helpers")
+
+return helpers.make_builtin({
+    name = "sqlfluff",
+    meta = {
+        url = "https://github.com/sqlfluff/sqlfluff",
+        description = "A SQL linter and auto-formatter for Humans",
+    },
+    method = null_ls.methods.DIAGNOSTICS_ON_SAVE,
```
jose-elias-alvarez, line 10: Running on save should only be necessary for linters that aren't able to accept `stdin`. I took a quick look at [the documentation](https://docs.sqlfluff.com/en/stable/cli.html) and it looks like it's supported, so could we give that a shot?

Edit: I saw that you used `stdin` for the formatter, but did it not work for diagnostics?

lua/null-ls/builtins/diagnostics/sqlfluff.lua
```diff
@@ -0,0 +1,41 @@
...
+    filetypes = { "sql" },
+    to_ = true,
+    generator_opts = {
+        async = true,
```
jose-elias-alvarez, line 14:
```suggestion
```

Not necessary to specify this when using `generator_factory`.

lua/null-ls/builtins/diagnostics/sqlfluff.lua
```diff
@@ -0,0 +1,41 @@
...
+        command = "sqlfluff",
+        args = { "lint", "-f", "github-annotation", "-n", "--disable_progress_bar", "$FILENAME" },
+        -- choose an output format (raw, json, or line)
+        from_stderr = true,
+        use_cache = true,
```
jose-elias-alvarez, line 19:
```suggestion
```

I think that this may cause issues. We should only use `use_cache` when two different sources can both operate on exactly the same output, but since the linter and formatter use different arguments, you may run into problems.

jose-elias-alvarez (commented):
Glad you figured this out! Left comments in a few places we might be able to clean this up.

lua/null-ls/builtins/diagnostics/sqlfluff.lua
KingMichaelPark: It doesn't work for diagnosis but I can enable diagnostics itself and it works when used in conjunction with to_stdin argument and the "-" argument instead of `$FILENAME`

jose-elias-alvarez:
Thanks for the revisions, LGTM!

PR #746, opened 2022-03-10
feat(builtins) add haml-lint diagnostics

smarquez1:
See https://github.com/sds/haml-lint

PR #787, opened 2022-03-29
feat: add erblint diagnostics and formatter

smarquez1:
Adds erblint diagnostics and formatter.

### DIFF
--- a/lua/null-ls/builtins/formatting/erb_lint.lua
+++ b/lua/null-ls/builtins/formatting/erb_lint.lua
@@ -11,14 +11,17 @@ return h.make_builtin({
     },
     method = FORMATTING,
     filetypes = { "eruby" },
+    factory = h.generator_factory,
     generator_opts = {
         command = "erblint",
-        args = {
-            "--autocorrect",
-            "--stdin",
-            "$FILENAME",
-        },
+        args = { "--autocorrect", "--stdin", "$FILENAME" },
+        ignore_stderr = true,
         to_stdin = true,
+        output = "raw",
+        on_output = function(params, done)
+            local output = params.output
+            local metadata_end = output:match(".*==()") + 1
+            return done({ { text = output:sub(metadata_end) } })
+        end,
     },
-    factory = h.formatter_factory,
 })
--- a/lua/null-ls/builtins/formatting/erb_lint.lua
+++ b/lua/null-ls/builtins/formatting/erb_lint.lua
@@ -1,0 +1,24 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "erb-lint",
+    meta = {
+        url = "https://github.com/Shopify/erb-lint",
+        description = "Lint your ERB or HTML files",
+    },
+    method = FORMATTING,
+    filetypes = { "eruby" },
+    generator_opts = {
+        command = "erblint",
+        args = {
+            "--autocorrect",
+            "--stdin",
+            "$FILENAME",
+        },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -955,6 +955,44 @@ describe("diagnostics", function()
         end)
     end)
 
+    describe("hamllint", function()
+        local linter = diagnostics.haml_lint
+        local parser = linter._opts.on_output
+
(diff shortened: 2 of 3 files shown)


lua/null-ls/builtins/formatting/erb_lint.lua
```diff
@@ -0,0 +1,24 @@
...
+        command = "erblint",
+        args = {
+            "--autocorrect",
+            "--stdin",
+            "$FILENAME",
+        },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})
```
smarquez1, line 24: This is producing errors, After I format a file it adds some unwanted output to it, for ex:

it produces this output:
```
Linting 1 files with 11 autocorrectable linters...

No errors were found in ERB files
================ /Users/.../test.html.erb ==================
<div>
```

from this file:
```<div       >```

How can I removed that?

File: test/spec/builtins/diagnostics_spec.lua [hunk 2]
```diff
@@ -954,4 +954,45 @@ describe("diagnostics", function()
...
+
+            local diagnostic = parser({ output = output })
+            assert.same({
+                row = 1,
+                end_row = 1,
+                col = 4,
+                end_col = 8,
+                code = "SpaceInHtmlTag",
+                message = 'Extra space detected where there should be no space.',
+            }, diagnostic)
```
smarquez1, line 995: This spec can be fixed  by comparing that object against `diagnostic[1]` instead of `diagnostic` but I wasn't sure if doing that is right. Am I missing anything else? Diagnostics display correctly in my local setup

lua/null-ls/builtins/formatting/erb_lint.lua
smarquez1: I found out that erblint's vscode extension clears the output after formatting, can I do such a thing using null-ls? Is there any example available? https://github.com/manuelpuyol/vscode-erb-linter/blob/main/src/erbLintAutocorrectProvider.ts#L38-L60

jose-elias-alvarez: Yep, you can do something similar with null-ls by handling the output manually. I don't think we have any examples but hopefully this should get you started:

```diff
diff --git a/lua/null-ls/builtins/formatting/erb_lint.lua b/lua/null-ls/builtins/formatting/erb_lint.lua
index f4a9118..e20e60b 100644
--- a/lua/null-ls/builtins/formatting/erb_lint.lua
+++ b/lua/null-ls/builtins/formatting/erb_lint.lua
@@ -12,6 +12,7 @@ return h.make_builtin({
     method = FORMATTING,
     filetypes = { "eruby" },
     generator_opts = {
+        ignore_stderr = true,
         command = "erblint",
         args = {
             "--autocorrect",
@@ -19,6 +20,13 @@ return h.make_builtin({
             "$FILENAME",
         },
         to_stdin = true,
+        on_output = function(params, done)
+            local output = params.output
+            -- this is just an example that may or may not work
+            -- the idea is to find the last == sequence, but idk if that's a good assumption
+            local metadata_end = output:match(".*==()") + 1
+            return done({ { text = output:sub(metadata_end) } })
+        end,
     },
-    factory = h.formatter_factory,
+    factory = h.generator_factory,
 })
```

test/spec/builtins/diagnostics_spec.lua
jose-elias-alvarez: Depending on the format, a generator may return either a list of diagnostics or a single diagnostic, so referencing `diagnostic[1]` is fine. Another approach we use in [other tests](https://github.com/smarquez1/null-ls.nvim/blob/0edc274049258243a7a01f46ca86f255f639e49b/test/spec/builtins/diagnostics_spec.lua#L412-L422) is to wrap the expected output with a list.

lua/null-ls/builtins/formatting/erb_lint.lua
smarquez1: Cool. You are doing something similar on the eslint formatted but I couldn't make it work. I think what made it work now is using ignore_stderr.

Thanks for all your help

test/spec/builtins/diagnostics_spec.lua
smarquez1: I ended up wrapping it on a list, I think it's cleaner that way

smarquez1:
I added tests for haml_lint diagnostics, for which I added support in this [PR](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/746).

I can move that test to a new PR if you think that's necessary

jose-elias-alvarez:
No worries, including the extra coverage is totally fine. Thanks again for the contributions!

PR #797, opened 2022-04-04
Do not set diagnostics for nonexisting buffer

przepompownia:
It can occur after (LSP) renaming the class without opening it (e.g. on its some reference).

### DIFF
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -131,6 +131,10 @@ M.handler = function(original_params)
     local method, uri = original_params.method, original_params.textDocument.uri
     local bufnr = vim.uri_to_bufnr(uri)
 
+    if not vim.api.nvim_buf_is_valid(bufnr) then
+        return
+    end
+
     if method == methods.lsp.DID_CLOSE then
         changedticks_by_uri[uri] = nil
         s.clear_cache(uri)
--- a/test/spec/diagnostics_spec.lua
+++ b/test/spec/diagnostics_spec.lua
@@ -18,6 +18,7 @@ describe("diagnostics", function()
         stub(u, "make_params")
         stub(generators, "run_registered")
         stub(vim, "uri_to_bufnr")
+        stub(vim.api, "nvim_buf_is_valid")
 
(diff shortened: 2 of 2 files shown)


przepompownia:
`vim.api.nvim_buf_is_valid` needs to be mocked, but I do not idea how to do it.

jose-elias-alvarez:
These notifications are sent by Neovim, so it's a little strange that this can happen in the first place, but it's a minor change so I'm not opposed to it if it solves a problem for you. Mocking the API method should be pretty straightforward:

```diff
diff --git a/test/spec/diagnostics_spec.lua b/test/spec/diagnostics_spec.lua
index e6b12e3..a7aaca8 100644
--- a/test/spec/diagnostics_spec.lua
+++ b/test/spec/diagnostics_spec.lua
@@ -18,6 +18,7 @@ describe("diagnostics", function()
         stub(u, "make_params")
         stub(generators, "run_registered")
         stub(vim, "uri_to_bufnr")
+        stub(vim.api, "nvim_buf_is_valid")
 
         local mock_uri = "file:///mock-file"
         local mock_handler = stub.new()
@@ -39,6 +40,7 @@ describe("diagnostics", function()
 
             u.make_params.returns(mock_params)
             vim.uri_to_bufnr.returns(mock_bufnr)
+            vim.api.nvim_buf_is_valid.returns(true)
         end)
 
         after_each(function()
```

przepompownia:
I tried this way and overlooked the boundary between the old and the new long test stdout therefore I was convinced it did not work  :wink:

przepompownia:
https://github.com/neovim/neovim/blob/master/runtime/lua/vim/diagnostic.lua#L51 Neovim could also have such validation, independetly.

jose-elias-alvarez:
Not sure the linked line is related, since if there's an issue here, it would happen when null-ls receives a diagnostic notification for an invalid buffer, which is part of the LSP architecture and not the diagnostic API. Either way, like I said this change is harmless, so I'm happy to merge it. Thanks!

przepompownia:
Thank you too. Both neovim's diagnostic caching mentioned above and LSP rename handler also could be fixed.

PR #798, opened 2022-04-04
Use stdin-path option for phpcs

weeman1337:
[PHPCS supports a `stdin-path` option](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Usage) to specify the path of the file passed to stdin.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/phpcs.lua
+++ b/lua/null-ls/builtins/diagnostics/phpcs.lua
@@ -27,6 +27,7 @@ return h.make_builtin({
             "ignore_errors_on_exit",
             "1",
             -- process stdin
+            "--stdin-path=$FILENAME",
             "-",
         },
         format = "json_raw",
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/diagnostics/phpcs.lua
```diff
@@ -46,11 +47,15 @@ return h.make_builtin({
                     WARNING = h.diagnostics.severities["warning"],
                 },
             })
-            params.messages = params.output
-                    and params.output.files
-                    and params.output.files["STDIN"]
-                    and params.output.files["STDIN"].messages
-                or {}
+
+            params.messages = {}
```
jose-elias-alvarez, line 51: Is this other change required because of the change in arguments? Otherwise I'm not sure I see a big advantage here.

weeman1337: Yes. Before the change PHPCS returned the `STDIN` file. After setting `--stdin-path=$FILENAME` this becomes the actual filename. I couldn't find any easier way getting the first element from the table. If you know a better way I am happy to learn.

weeman1337: Now using `bufname`

jose-elias-alvarez:
Understood, this looks good then. Thanks!

PR #799, opened 2022-04-04
Allow overriding to_temp_file

weeman1337:
Allows overriding the `to_temp_file` option.

The reason behind is that for example PHPStan doesn't support std in.  
If using temp files you may loose path-based project specific configuration.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/phpstan.lua
+++ b/lua/null-ls/builtins/diagnostics/phpstan.lua
@@ -23,13 +23,16 @@ return h.make_builtin({
             return code <= 1
         end,
         on_output = function(params)
-            local parser = h.diagnostics.from_json({})
-            params.messages = params.output
-                    and params.output.files
-                    and params.output.files[params.temp_path]
-                    and params.output.files[params.temp_path].messages
-                or {}
+            params.messages = {}
 
+            if params.output and params.output.files and type(params.output.files) == "table" then
+                for k in pairs(params.output.files) do
+                    params.messages = params.output.files[k].messages or {}
+                    break
+                end
+            end
+
+            local parser = h.diagnostics.from_json({})
             return parser({ output = params.messages })
         end,
     },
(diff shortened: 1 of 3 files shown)


lua/null-ls/builtins/diagnostics/phpstan.lua
```diff
@@ -23,13 +23,16 @@ return h.make_builtin({
             return code <= 1
         end,
         on_output = function(params)
-            local parser = h.diagnostics.from_json({})
-            params.messages = params.output
-                    and params.output.files
-                    and params.output.files[params.temp_path]
-                    and params.output.files[params.temp_path].messages
-                or {}
+            params.messages = {}
```
jose-elias-alvarez, line 26: I think this is similar to the change in your other PR, but if it's just a refactor then I prefer the other style (or at least don't consider this better enough to justify the change).

jose-elias-alvarez (commented):
So generally speaking, we only use `to_temp_file` when there's no other option, e.g. in the case you mentioned where a linter doesn't support `stdin`. The only case where I can imagine this would be viable is if a user disables `to_temp_file` *and* changes the method so that the source only runs on save. Is that what you have in mind here?

If so, I think at a minimum we should improve the documentation here to guide users on potential issues they might encounter and how to fix them, because overriding `to_temp_file` alone will cause weird behavior.

weeman1337:
> The only case where I can imagine this would be viable is if a user disables to_temp_file and changes the method so that the source only runs on save. Is that what you have in mind here?

Exactly. The issue with temp files is that all path specific configuration won't be applied.

> If so, I think at a minimum we should improve the documentation here to guide users on potential issues they might encounter and how to fix them, because overriding to_temp_file alone will cause weird behavior.

I checked the other diagnostic builtins. They look fine to work if setting `to_temp_file` to `false`. Will update the documentation.

weeman1337:
Another option that came into my mind is to provide an option that is more explicit such as `run_in_place` (or `inplace`). Diagnostics that support it can pick it up and switch to in place validation on save.

weeman1337:
Thinking about it and after digging into the code I think the best solution is that phpstan diagnostic should support the `DIAGNOSTICS` and `DIAGNOSTICS_ON_SAVE` methods. What do you think @jose-elias-alvarez ?

PR #800, opened 2022-04-05
Allow overriding to_temp_file

weeman1337:
Unfortunately I was unable to reopen #799

Like discussed in the previous PR

- Added some docs
- Simplified phpcs code


jose-elias-alvarez:
Nice, I like the description and think the change in behavior makes sense, too. Thanks for the contributions!

PR #803, opened 2022-04-07
builtin: fix cppcheck builtin to use temp file

Ultra-Code:


jose-elias-alvarez:
Awesome, thanks for looking into this!

PR #804, opened 2022-04-07
Fix actionlint running against all yaml files

otahontas:
I've been using this condition locally to avoid running actionlint against all yaml files. However, since [Github docs ](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions) quite clearly specify that `.github/workflows` -folder is the default folder for all workflows, I'd say it's safe to default to same pattern here as well. 

Any thoughts on this?

### DIFF
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -3,6 +3,12 @@ local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
+local buf_path_in_workflow_folder = function()
+    local api = vim.api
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
I'm not opposed to this in principle, since the documentation is pretty clear. The issue is that `condition` is only checked once upon entering the first buffer with a matching filetype. With this setup, it would check the first `yaml` buffer entered, then remain disabled or enabled for all subsequent `yaml` buffers.

A way around this would be to use `runtime_condition`, which would check each buffer independently. However, without some form of caching it would run the check each time the source runs, which could cause slowdown (diagnostic sources run very, very often). I've been thinking about working on a generic caching solution for the results of conditions that aren't expected to change over the buffer's lifetime, which would let us enable things like this out-of-the-box, but until then I think this is better as a local modification.

PR #807, opened 2022-04-09
feat: add `jq` formatter for `json`

Logarithmus:
Written in C, so it's blazing fast


jose-elias-alvarez:
This is a good addition, thanks!

PR #808, opened 2022-04-11
Fix Do not set diagnostics for nonexisting buffer

przepompownia:


jose-elias-alvarez:
Like before, this is a safe change, so I'm happy to merge this as-is if this is causing you trouble. Otherwise, it'll take me a few days until I'm able to try out your repro, so let me know.

przepompownia:
In my real environment, similarly to the example project, I keep nvim extensions as git submodules so no problem if null-ls will have to wait few days on this branch. 

- The essential thing happens in https://github.com/przepompownia/nvim-test-null-ls/blob/master/test/01-nonexisting-buffer/run.lua:  - open only `test.php` (to kick on null-ls configured in https://github.com/przepompownia/nvim-test-null-ls/blob/master/plugin/null-ls.lua) 
- assuming that `old.php` have no buffer yet rename it the way LSP rename handler does it

```lua

local workspace = vim.loop.cwd() .. '/workspace/'
local util = require('vim.lsp.util')

local newFile = workspace .. 'new.php'
local oldFile = workspace .. 'old.php'

local bufNr = vim.fn.bufadd(workspace .. 'test.php')
vim.fn.bufload(bufNr)

os.rename(newFile, oldFile) -- only for making repeating test easy
local timer = vim.loop.new_timer()
timer:start(1000, 0, vim.schedule_wrap(function()
  util.rename(oldFile, newFile) -- used by LSP rename handler
end))
```

jose-elias-alvarez:
This is definitely the most impressive reproduction I've ever seen - very cool! I can see what's happening clearly, and while I think there's potentially room for improvement in upstream behavior, I think this is a sensible defensive change on our end, so I'm happy to merge it. Thanks for investigating!

przepompownia:
I'm glad you found this model useful and you were able to see what happens at different levels, including nvim diagnostic api (which probably should have similar quick return too).

PR #774, opened 2022-03-19
feat: adding solhint diagnostic support for solidity files

ahashim:
Closes #171.

Adding support for linting [Solidity](https://soliditylang.org/) files using [solhint](https://protofire.github.io/solhint/).

Changes include:

- solhint diagnostic builtin
- solhint documentation in BUILTINS.md
- solhint diagnostic tests

Tested with `solhint v3.3.7` on `Pop!_OS 21.10 x86_64` and `nvim v0.6.1`. Screenshot below showing an example of null-ls output with solhint via [trouble](https://github.com/folke/trouble.nvim).

Screenshot from 2022-03-19 12-12-54

PR #811, opened 2022-04-13
fix: switching from_stderr to `false` for solhint diagnostics

ahashim:
This fixes not seeing warnings/errors when running null-ls diagnostics on Solidity files. The change will now make diagnostic messages to actually show up in the proper output buffer.

Here is the `NullLsLog` output for said diagnostics in debug mode:

solhint-diagnostics-fix

### DIFF
--- a/lua/null-ls/builtins/diagnostics/solhint.lua
+++ b/lua/null-ls/builtins/diagnostics/solhint.lua
@@ -18,7 +18,7 @@ return h.make_builtin({
         },
         command = "solhint",
         format = "line",
-        from_stderr = true,
+        from_stderr = false,
         on_output = h.diagnostics.from_pattern("([^:]*):([%d]+):([%d]+): (.*) %[([%a]+)/([%a%p]+)%]", {
             "filename",
             "row",


ahashim:
Thanks for looking at this, and there is no rush - i've got it fixed locally via my branch. 

It's unclear to me why it stopped working, but I suspect that I did not committed the proper configuration in [my original PR](https://github.com/jose-elias-alvarez/null-ls.nvim/pull/774), and just didn't notice until now when i ran the linter manually in the consle. My nvim & solhint versions are still the same as the ones in that linked PR.

Feel free to verify & test locally before I mark it as ready for review. And thanks again!

jose-elias-alvarez:
So I tested this out with a fresh install of `solhint` from `npm` (version 3.3.7) with a default config file generated with `solhint --init` and it works for me *without* this change, but not with it. I wonder what's going on? I'm not familiar with the ecosystem, but it looks like the latest version from `npm` does output to `stderr`.

ahashim:
Interesting, then there must be something wrong with my local configuration. Thanks for verifying this works as intended out the box with vanilla solhint. I'll close this PR for now & investigate my local configuration further.

Thanks again!

DavideSilva:
@ahashim I am also getting the same issue as you have. `solhint` does indeed report errors to `null-ls`. That I can verify by checking the `null-ls.log`, but the diagnostic only appears if I set the `stderr` option in the `solhint.lua` to `false`.

My `nvim` version is `0.7.0` and I installed the latest `solhint`, version `3.3.7`, like @jose-elias-alvarez.

Should we reopen this issue until we can figure out exactly what's going on?

ahashim:
@DavideSilva thanks for verifying. Yes, I'll reopen the PR, but will not mark it as ready until we figure out the root cause of the regression. I currently don't have much bandwidth to investigate, but can create a demo solidity project to experiment with null-ls settings sometime this week.

Appreciate you helping to triage the issue with @jose-elias-alvarez.

ahashim:
I did some experimenting in a clean environment this morning to reproduce the bug, and here are my findings:

### Operating System
`Pop!_OS 21.10 x86_64`

### Software
`nvim v0.7.0`
`solhint v3.3.7`

### Repository
https://github.com/ahashim/null-ls-solhint-sandbox

### Steps to reproduce

1. Ensure you are running the specific versions of the software listed above, and clone the `null-ls-solhint-sandbox` repository to your local machine.
2. Ensure null-ls is in [`debug mode`](https://github.com/jose-elias-alvarez/null-ls.nvim/#how-do-i-enable-debug-mode-and-get-debug-output)
3. In the root directory, run the following command: `solhint SolhintTest.sol --formatter=unix`, and you will see the following output: 
```
SolhintTest.sol:10:5: Function name must be in mixedCase [Warning/func-name-mixedcase]
SolhintTest.sol:10:62: Code contains empty blocks [Warning/no-empty-blocks]
```
4. Open `SolhintTest.sol` in neovim, and check null-ls output with `:NullLsLog`

Using @jose-elias-alvarez's `main` branch for the solhint diagnostics builtin, the output is as follows:

Screenshot from 2022-04-19 12-16-03

Using @ahashim's `solhint-bugfix` branch for the solhint diagnostics builtin, the output is as follows:

Screenshot from 2022-04-19 12-18-49

As you can see, the error output for both branches is `nil`, and @ahashim's `solhint-bugfix` branch correctly allows the diagnostics to be propagated (in this case to [`trouble.nvim`](https://github.com/folke/trouble.nvim)).

It seems that tests are also passing for the `solhint-bugfix` branch as well [as seen in this CI run](https://github.com/jose-elias-alvarez/null-ls.nvim/runs/6080922742?check_suite_focus=true#step:5:376). To me, it seems that this PR fixes the bug & produces the intended behavior. However, I'd love to get some more verification on this before proceeding. 

@DavideSilva or @jose-elias-alvarez If you can verify for  me the above is correct, then perhaps we can figure out where the disconnect is. AFAIK `solhint` has not changed the way it produces output, so this situation is a bit puzzling to me.

Thanks again for taking the time to help with this 🙏🏾

DavideSilva:
Thank you for putting this testing repo together. I can confirm that I have exactly the same behavior as you @ahashim.

Using the same Neovim and Solhint versions

Screenshot 2022-04-20 at 09 26 54

Opening the `SolhintTest.sol` file with Neovim on the `main` branch of `null-ls`

Screenshot 2022-04-20 at 09 31 57

Changing the `stderr` option to `false` in the `solhint.lua` diagnostic file, essentially using the `ahashim:solhint-bugfix` branch, the warnings now appear in Neovim

Screenshot 2022-04-20 at 09 34 51

jose-elias-alvarez:
Alright, not 100% sure what's going on but I can confirm that I see the behavior you describe with your reproduction. Let me know if this is okay to merge (I went ahead and pushed a fix that was causing CI to fail for an unrelated reason).

ahashim:
Thanks for testing this & confirming the expected behavior @jose-elias-alvarez & @DavideSilva. I suspect the error was still mine when I initially committed the PR (and just didn't realize it for a bit). 

Regardless, I'm happy to see this working again. Thank you for fixing the tests, and feel free to merge at your earliest convenience ✌🏾

jose-elias-alvarez:
Thanks all!

KalleChen:
> So I tested this out with a fresh install of `solhint` from `npm` (version 3.3.7) with a default config file generated with `solhint --init` and it works for me _without_ this change, but not with it. I wonder what's going on? I'm not familiar with the ecosystem, but it looks like the latest version from `npm` does output to `stderr`.

Hi @jose-elias-alvarez , actually my situation is the same as yours. Right now, I have to set `stderr` to `true` to see the diagnostic messages... But I'm new to solhint and lua, so I'm not sure which is correct.
neovim: v0.7.0
solhint: 3.3.7

KalleChen:
Hi guys, @ahashim @jose-elias-alvarez although this PR was closed, but seems that this setting is not 100% sure, so I'm here to share what I have found. 

Currently, I encounter a similar error with my project but I have to set `stderr` back to true to get my diagnostics. But after testing `null-ls-solhint-sandbox` repo, I have the same behavior above(setting `stderr` to false to get diagnostics). 

After comparing the difference between my project and `null-ls-solhint-sandbox`, I found that if you set `compiler-version` rule in `.solhint.json`, solhint will not output to `stderr`, but if `compiler-version` rule was not set, solhint will output to `stderr`.  

But I'm not familiar with solhint, so I don't know why this happens.

KalleChen:
@ahashim Thanks for your response. Yeah, I think it should be rectified in the solhint repository. In my test case, changing `console.log` in `printReports [solhint.js]` to `console.error` would be the simplest way to make sure it sent message to `std_err`. But I'm not sure if that will affect other projects which use `solhint`, therefore, I'm still figuring out why `solhint` sends to `std_err` if there're errors.

Besides, while I investigated the `solhint` repo, it seems that the main maintainer `fvictorio` has no permission to write the repo, but he can still publish it on npm. You can found his response [here](https://github.com/protofire/solhint/issues/328#issuecomment-1034878542). And the updated repo is [here](https://github.com/fvictorio/solhint/tree/upgrade-parser).

Also, I've opened an PR [here](https://github.com/fvictorio/solhint/pull/1) to add formatter support for stdin process in `solhint`, once that is merged and is published. We can now use stdin for `null-ls` here for real-time diagnostics. But there're still other issue for `solhint-plugin-prettier` while using stdin, I've open an issue [here](https://github.com/fvictorio/solhint-plugin-prettier/issues/10)

Thanks again.

KalleChen:
@ahashim Ok, I've found why this will happen. It seems that if there're `Errors` in diagnostics reports the process will exit with 1, therefore, the `console.log` output will be sent to std_err. I've opened the [PR](https://github.com/fvictorio/solhint/pull/2) to fix this. 

What're your thoughts about this?

jose-elias-alvarez:
Thanks all for investigating! The behavior we were seeing was pretty confusing, so I think the upstream solution makes sense.

PR #812, opened 2022-04-14
feat: Add bibclean formatter for BibTeX

even4void:
bibclean is a fast fixer (and also linter) for BibTeX files, written in C. It is usually available on CTAN, but latest versions are available on the author's website.

### DIFF
--- a/lua/null-ls/builtins/formatting/bibclean.lua
+++ b/lua/null-ls/builtins/formatting/bibclean.lua
@@ -1,0 +1,23 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "bibclean",
+    meta = {
+        url = "https://www.ctan.org/pkg/bibclean",
+        description = "A portable program (written in C) that will pretty-print, syntax check, and generally sort out a BibTeX database file.",
+        notes = {
+            "See [bibclean: prettyprint and syntax check BibTeX and Scribe bibliography data base files](https://ftp.math.utah.edu/pub/bibclean/) for latest version.",
+        },
+    },
+    method = FORMATTING,
+    filetypes = { "bib" },
+    generator_opts = {
+        command = "bibclean",
+        args = { "-align-equals", "-delete-empty-values" },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lua/null-ls/builtins/formatting/bibclean.lua
```diff
@@ -0,0 +1,20 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "bibclean",
+    meta = {
+        url = "https://www.ctan.org/pkg/bibclean",
+        description = "A portable program (written in C) that will pretty-print, syntax check, and generally sort out a BibTeX database file. See https://ftp.math.utah.edu/pub/bibclean/ for latest version.",
```
jose-elias-alvarez, line 10: Nit: Could we move the mention of the latest version into a `notes` entry and turn the URL into a Markdown link? See [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/64d9e2b58a17cd85138c265ab816198000e126d2/lua/null-ls/builtins/formatting/eslint.lua#L12-L14) for an example.

jose-elias-alvarez (commented):
Looks good! One small change we can make to how the documentation is structured. Also, will this work out-of-the-box with the release version of the formatter? Otherwise I think we should mention that, too.

lua/null-ls/builtins/formatting/bibclean.lua
even4void: Ah, I was looking for such an additional field. Thanks!

even4void:
> Also, will this work out-of-the-box with the release version of the formatter? Otherwise I think we should mention that, too.

Not sure if you mean null-ls.nvim or bibclean. 
I added a source in my null-ls config (`null_ls.builtins.formatting.bibclean`), and I tested that it indeed works with one of the latest release of bibclean:

```shell
~ % bibclean --version                                                            
bibclean Version 3.04 [21-Nov-2019]

Compiled by <chl@aliquote> on Dec 22 2020 09:27:27
with native pattern matching
E-mail bug reports to Nelson H. F. Beebe <beebe@math.utah.edu>

Copyright (C) 1990--2018 by Nelson H. F. Beebe and the Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Although it just works on my machine, I assume that the current settings wiil equally work with any release, but I can perform further testing if you like, e.g. with older version on CTAN and latest release (v3.06).

jose-elias-alvarez:
> Not sure if you mean null-ls.nvim or bibclean.

Sorry, should have clarified that I meant bibclean. I'm not at all familiar with the ecosystem, but it would be great if we could check this against the release version, since otherwise users may be confused if it doesn't work.

even4void:
>  but it would be great if we could check this against the release version, since otherwise users may be confused if it doesn't work.

Done. It works with the latest release (v3.06) as well.

```shell
~/tmp % uname -a                                                                                                                                                                
Linux aliquote 5.13.0-39-generic #44~20.04.1-Ubuntu SMP Thu Mar 24 16:43:35 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
~/tmp % wget https://ftp.math.utah.edu/pub/bibclean/bibclean-3.06.tar.xz
~/tmp % cd bibclean-3.06
~/tmp % ./configure && make
~/tmp % mv bibclean ~/.local/bin
~/tmp % bibclean --version                                                                                                                                                      
bibclean Version 3.06 [06-Mar-2021]

Compiled by <chl@aliquote> on Apr 14 2022 17:48:10
with native pattern matching
E-mail bug reports to Nelson H. F. Beebe <beebe@math.utah.edu>

Copyright (C) 1990--2018 by Nelson H. F. Beebe and the Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Minimal BibTeX test file (`test.bib`):

```bibtex
@Book{abelson-1996-struc-inter,
  author       = "Harold Abelson and Gerald Jay Sussman",
  title        = "Structure and Interpretation of Computer Programs",
  year         = "1996",
  publisher    = "MIT Press",
  edition      = "2",
}
```

To test that the formatter is working:

1. Add the `bibclean` source in your Neovim config file;
2. Open `test.bib`, then delete the publication year but keep the empty quotes, and delete some space before any equal signs, e.g. `gg03jf"di"k6hxx`. 

When saving, the entry will be automagically formatted (equal signs aligned and empty record deleted).

even4void:
Now that I thought about the args list, do you think it is a good idea to provide so many default arguments, instead of letting users decide what's necessary for their own use, using `with({ extra_args = { ... } })`? I think the ones I retained are the most interesting options, but if you disagree we could just keep "-align-equals" and "-delete-empty-values" (default values are set to FALSE) since they are used in other fixers (at least in Emacs).

jose-elias-alvarez:
Thanks for the thorough investigation! If we know it works with the release version there should be no issues. For the arguments issue, we should definitely keep it minimal and let users configure anything beyond that (I think following the Emacs example is sensible).

even4void:
> For the arguments issue, we should definitely keep it minimal and let users configure anything beyond that (I think following the Emacs example is sensible).

I force-pushed yet another commit, which discards non-essential flags. I also removed the "-quiet" (equivalent to "-no-warnings") argument since the author himself recommends to keep it as it may signal severe issues with malformed BibTeX file.

jose-elias-alvarez:
Awesome, I think this is all set, then. Thank you for the contribution!

PR #818, opened 2022-04-18
feat(builtin): Add `rome format` for javascript and typescript

p-inks:
This adds support [Rome](https://rome.tools) formatting source.


jose-elias-alvarez:
Looks good, thanks! I pushed a fix that was causing CI to fail but it's unrelated to your PR.

PR #821, opened 2022-04-20
feat(builtins): rewrite teal diagnostics

Saecki:
- support different severities
- don't show diagnostics for different files
- improve parsing of spans

### DIFF
--- a/lua/null-ls/builtins/diagnostics/teal.lua
+++ b/lua/null-ls/builtins/diagnostics/teal.lua
@@ -1,7 +1,81 @@
-local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local h = require("null-ls.helpers")
+local end_col_from_quote = h.diagnostics.adapters.end_col.from_quote.end_col
 
-local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+local severities = {
+    error = h.diagnostics.severities.error,
+    errors = h.diagnostics.severities.error,
+    warning = h.diagnostics.severities.warning,
+    warnings = h.diagnostics.severities.warning,
+}
+
+local function parse_diagnostics(params, done)
+    if not params.output then
+        done(nil)
+        return
+    end
+
+    local diagnostics = {}
+    local current_severity = h.diagnostics.severities.error
+
+    local output = params.output:gsub("\r\n?", "\n") or params.output
+    if not output then
+        done(nil)
+        return
+    end
+
+    local function parse_line(line)
+        local _, severity = line:match([[^(%d*) ([%w]+):$]])
+        if severity then
+            current_severity = severities[severity] or h.diagnostics.severities.error
+            return
+        end
+
+        local filename, row, col, message = line:match([[([^:]+):(%d+):(%d+): (.*)$]])
+        if not (filename and row and col and message) then
+            return
+        end
+
+        -- only return diagnostics for the current file
+        if filename ~= params.temp_path then
+            return
+        end
+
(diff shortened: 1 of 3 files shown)


Saecki:
I've also noticed that all diagnostics that use the `end_col.from_quote` helper also use an offset of 1. Should I open a PR that offsets the `end_col` inside the helper itself rather than every builtin?

lua/null-ls/builtins/diagnostics/teal.lua
```diff
@@ -1,15 +1,22 @@
...
+local end_col_from_quote = h.diagnostics.adapters.end_col.from_quote.end_col
 
-local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+local severities = {
+    error = h.diagnostics.severities.error,
+    errors = h.diagnostics.severities.error,
+    warning = h.diagnostics.severities.warning,
+    warnings = h.diagnostics.severities.warning,
+}
+local current_serverity = h.diagnostics.severities.error
```
jose-elias-alvarez, line 11: Hmm, I think I see what you're doing here, but I don't really like having this variable outside of the `on_output` closure. An option I prefer is to delete `output_format` and leave it unspecified, which will pass all linter output as `params.output`. You can then iterate over the lines manually and keep this variable within the closure. 

Let me know if that makes sense - it might require a few more changes but your overall logic (which is a big improvement) should still work for the most part.

jose-elias-alvarez:
> I've also noticed that all diagnostics that use the `end_col.from_quote` helper also use an offset of 1. Should I open a PR that offsets the `end_col` inside the helper itself rather than every builtin?

Yep, I think that makes sense. Also, the CI failure is most likely unrelated to your changes - if you rebase onto `main` it should be all good.

lua/null-ls/builtins/diagnostics/teal.lua
Saecki: Yeah, I was trying to avoid duplicating the code to split up lines, but I have to admit I also didn't like having that "global" variable there.

lua/null-ls/builtins/diagnostics/teal.lua
```diff
@@ -1,40 +1,99 @@
...
+}
+
+local function parse_diagnostics(params, done)
+    if not params.output then
+        done(nil)
+        return
+    end
+
+    local diagnostics = {}
+    local current_serverity = h.diagnostics.severities.error
```
jose-elias-alvarez, line 19:
```suggestion
    local current_severity = h.diagnostics.severities.error
```

test/spec/e2e_spec.lua
```diff
@@ -344,7 +344,7 @@ describe("e2e", function()
 
             local tl_check_diagnostic = buf_diagnostics[1]
             assert.equals(tl_check_diagnostic.message, "in return value: got string, expected number")
-            assert.equals(tl_check_diagnostic.source, "tl check")
+            assert.equals(tl_check_diagnostic.source, "teal")
```
jose-elias-alvarez, line 347: I think this is a sensible change, but I want to make a note to myself to flag this a technical breaking change, since users may be relying on the name of the source.

jose-elias-alvarez (commented):
Looks great! Just one typo fix and I think this is good to go.

lua/null-ls/builtins/diagnostics/teal.lua
Saecki: Oops

jose-elias-alvarez:
Great, thanks!

PR #823, opened 2022-04-22
refactor(helpers): offset end_col.from_quote by 1

Saecki:
As mentioned in #821, all builtins that used end_col.from_quote added a manual offset of 1. This removes the need for that by offsetting the end_col in the helper itself.
This also removes the adapter from luacheck since no `_quote` field was parsed anyway.


jose-elias-alvarez:
Makes sense, thanks for noticing this and for putting in the PR!

PR #829, opened 2022-04-26
feat(builtins): add crates.nvim code actions

Saecki:
This adds code actions for [`crates.nvim`](https://github.com/saecki/crates.nvim) I will merge the PR at https://github.com/Saecki/crates.nvim/pull/24 as soon as you're okay with this one.

### DIFF
--- a/lua/null-ls/builtins/code_actions/crates.lua
+++ b/lua/null-ls/builtins/code_actions/crates.lua
@@ -1,0 +1,40 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local CODE_ACTION = methods.internal.CODE_ACTION
+
+return h.make_builtin({
+    name = "crates",
+    meta = {
+        url = "https://github.com/saecki/crates.nvim",
+        description = "Code actions for editing `Cargo.toml` files.",
+    },
+    method = CODE_ACTION,
+    filetypes = { "toml" },
+    condition = function()
+        return vim.fn.expand("%:t") == "Cargo.toml"
+    end,
+    generator = {
+        fn = function(params)
+            local ok, crates_actions = pcall(require, "crates.actions")
+            if not ok or not crates_actions then
+                return
+            end
+
+            local name_to_title = function(name)
+                return name:sub(1, 1):upper() .. name:gsub("_", " "):sub(2)
+            end
+
+            local actions = {}
+            for name, action in pairs(crates_actions.get_actions()) do
+                table.insert(actions, {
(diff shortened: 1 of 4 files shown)


lua/null-ls/builtins/code_actions/crates.lua
```diff
@@ -0,0 +1,40 @@
...
+
+return h.make_builtin({
+    name = "crates",
+    meta = {
+        url = "https://github.com/saecki/crates.nvim",
+        description = "Code actions for editing `Cargo.toml` files.",
+    },
+    method = CODE_ACTION,
+    filetypes = { "toml" },
+    condition = function()
```
jose-elias-alvarez, line 14: An issue with `condition` is that this will be checked the first time the user opens a `toml` buffer. That means that if the first buffer's name is *not* `Cargo.toml`, the source will never be active, and if it *is*, the source will be active for all other `toml` buffers. This is an issue with the null-ls architecture, which is based on filetypes and not specific buffers (I eventually plan on rewriting so that sources "attach" to a buffer).

For now, we have another option called `runtime_condition` which will take care of this case, since it will check if the source is active every time the user requests code actions. Normally, users won't request code actions often unless they're using something like [nvim-lightbulb](https://github.com/kosayoda/nvim-lightbulb) so we should be okay.

lua/null-ls/builtins/code_actions/crates.lua
```diff
@@ -0,0 +1,40 @@
...
+        return vim.fn.expand("%:t") == "Cargo.toml"
```
jose-elias-alvarez, line 15: Might be slightly faster and more accurate to avoid the Vim function and instead check `vim.api.nvim_buf_get_name(params.bufnr)` with a pattern.

jose-elias-alvarez (commented):
Looks good overall! You could theoretically use[the source API](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/SOURCES.md#registerto_register) to register the source directly from your plugin and avoid using the built-in as a bridge. I'm happy to merge this either way, but I wanted to point out the option since it might make setup for your users easier if they can just enable the integration with a single config flag (which you might even consider turning on by default if they have null-ls installed).

Saecki:
> You could theoretically use the source API to register the source directly from your plugin and avoid using the built-in as a bridge.

That does indeed sound like a good idea.

> I wanted to point out the option since it might make setup for your users easier if they can just enable the integration with a 
single config flag.

I totally agree, closing this PR.

PR #833, opened 2022-04-29
fix: eslint cwd

hori-ryota:


lua/null-ls/builtins/code_actions/eslint.lua
```diff
@@ -166,6 +167,9 @@ return h.make_builtin({
             return code_action_handler(params)
         end,
         dynamic_command = cmd_resolver.from_node_modules,
+        cwd = function(params)
+            return u.root_pattern(".eslintrc", ".eslintrc.*")(params.bufname)
```
jose-elias-alvarez, line 171: I don't think this will work as expected - we'd have to declare the patterns similar to how [lspconfig does it](https://github.com/neovim/nvim-lspconfig/blob/21102d5e3b6ffc6929d60418581ac1a29ee9eddd/lua/lspconfig/server_configurations/eslint.lua#L65).

jose-elias-alvarez (commented):
I think this is a reasonable change, but it's one I'd rather not merge as-is. The `cwd` callback runs every time the source runs, meaning that it'll scan the filesystem very frequently (especially if you're using the diagnostics source) and noticeably affect performance. That's fine if a user has specifically enabled it and is okay with the tradeoff, but it's not great as a default, especially since we don't have a good way of shutting it off. 

I talked about this a bit in #804, but I'd like to add a generic caching mechanism for the output of these callbacks so that we can use them as defaults without worrying as much about performance. I'm planning on working on that when I have free time, but otherwise I'm happy to review a PR if you or anybody else wants to work on it.

lua/null-ls/builtins/code_actions/eslint.lua
hori-ryota: fixed

hori-ryota:
Ok, I'll apply it only my settings like following. thanks.

https://github.com/hori-ryota/dotfiles/commit/0108de152aa28abe8edfe014ff39d728d683bd06

jose-elias-alvarez:
I added a helper that should handle the case I described in 490a09c0ccfc8da22f6840d6a03398ef5d1b3035. I briefly tried it out with the changes in your PR and it seems to work well, but I'd appreciate it if you could test and update the PR if it works for you, too (and let me know if it doesn't so I can fix it). This is what the logic should look like:

```lua
return h.make_builtin({
    name = "eslint",
    meta = {
        url = "https://github.com/eslint/eslint",
        description = "A linter for the JavaScript ecosystem.",
    },
    method = DIAGNOSTICS,
    filetypes = { "javascript", "javascriptreact", "typescript", "typescriptreact", "vue" },
    generator_opts = {
        command = "eslint",
        args = { "-f", "json", "--stdin", "--stdin-filename", "$FILENAME" },
        to_stdin = true,
        format = "json_raw",
            return code <= 1
        end,
        use_cache = true,
        on_output = handle_eslint_output,
        dynamic_command = cmd_resolver.from_node_modules,
        -- this is the change - instead of running the callback each time, it'll only run it once per bufnr
        cwd = h.cache.by_bufnr(function(params)
            return u.root_pattern(
                ".eslintrc",
                ".eslintrc.js",
                ".eslintrc.cjs",
                ".eslintrc.yaml",
                ".eslintrc.yml",
                ".eslintrc.json",
                "package.json"
            )(params.bufname)
        end),
    },
})
```

jose-elias-alvarez:
Great, thanks! Let's give this a shot and see how it goes.

saurabhchardereal:
Shouldn't this also applies to `eslint_d` and `prettier{,_d}`?

jose-elias-alvarez:
> Shouldn't this also applies to `eslint_d` and `prettier{,_d}`?

PR welcome - putting in the equivalent change for Prettier should be safe, but in my experience we can't rely on `eslint_d` and `prettier_d` to behave identically, so I'd appreciate it if someone was willing to spend some time testing them out.

jose-elias-alvarez:
> Yes, but since `eslint_d` inherits from `eslint`, it seems that it has already been applied at this config.

Whoops, forgot about this but you're right. Anyone who wants to put in a PR to do something similar for Prettier / other sources should feel free to do so.

PR #837, opened 2022-04-30
fix: adapt to 0.8 breaking changes

### LINKED ISSUE #835: bug: for_each_buffer_client has a nil value for server_capabilities of an attached client
### FAQ

### Issues

### Neovim Version

v0.8.0-dev+84-g5b04e46d23

### Operating System

Arch Linux

### Minimal config

Any configuration would do.

### Steps to reproduce

With Neovim `v0.8.0-dev+83-geecc6535eb` installed, write a buffer that has null-ls attached.

### Expected behavior

No errors.

### Actual behavior

Neovim shows the following error message:
```
Error detected while processing BufWritePost Autocommands for "<buffer=1>":                                                                                                                                                      
E5108: Error executing lua vim/shared.lua:0: attempt to index a boolean value                                                                                                                                                    
stack traceback:                                                                                                                                                                                                                 
        vim/shared.lua: in function 'tbl_get'                                                                                                                                                                                    
        /usr/share/nvim/runtime/lua/vim/lsp.lua:1214: in function 'fn'                                                                                                                                                           
        /usr/share/nvim/runtime/lua/vim/lsp.lua:164: in function 'for_each_buffer_client'                                                                                                                                        
        /usr/share/nvim/runtime/lua/vim/lsp.lua:1211: in function '_text_document_did_save_handler'                                                                                                                              
        [string ":lua"]:1: in main chunk                                                   
```

Seems that the `client.server_capabilities` should be provided.

Related Neovim bug report: neovim/neovim#18319

### Debug log

No output is produced.

### Help

No

### Implementation help

_No response_

### Requirements

jose-elias-alvarez:
Should fix #835. The specific issue is that the type of `textDocumentSync.save` doesn't match what Neovim expects. I think this is an upstream regression, since the specification indicates that a boolean value is acceptable:

Screen Shot 2022-04-30 at 10 55 13 AM

But it's an easy fix on our end. I've also included a fix for the deprecation warning from accessing `client.resolved_capabilities` on 0.8.

Edit: tests will fail until there's a new nightly build, since the CI action doesn't build from source, but they're passing for me locally. I'd appreciate confirmation from anyone who commented on the original issue!

### DIFF
--- a/lua/null-ls/client.lua
+++ b/lua/null-ls/client.lua
@@ -30,10 +30,24 @@ local should_attach = function(bufnr)
 end
 
 local on_init = function(new_client, initialize_result)
+    local capability_is_disabled
+    if u.has_version("0.8") then
+        capability_is_disabled = function(method)
+            -- TODO: extract map to prevent future issues
+            local required_capability = lsp._request_name_to_capability[method]
+            return not required_capability
+                or vim.tbl_get(new_client.server_capabilities, unpack(required_capability)) == false
+        end
+    else
+        capability_is_disabled = function(method)
+            return new_client.resolved_capabilities[methods.request_name_to_capability[method]] == false
+        end
+    end
+
     -- null-ls broadcasts all capabilities on launch, so this lets us have finer control
     new_client.supports_method = function(method)
-        -- capability was not declared or is specifically disabled
-        if new_client.resolved_capabilities[methods.request_name_to_capability[method]] == false then
+        -- allow users to specifically disable capabilities
+        if capability_is_disabled(method) then
             return false
         end
 
--- a/lua/null-ls/rpc.lua
+++ b/lua/null-ls/rpc.lua
@@ -42,7 +42,7 @@ local capabilities = {
     textDocumentSync = {
         change = 1, -- prompt LSP client to send full document text on didOpen and didChange
         openClose = true,
-        save = true,
+        save = { includeText = true },
     },
     hoverProvider = true,
 }
--- a/test/spec/client_spec.lua
+++ b/test/spec/client_spec.lua
@@ -4,6 +4,7 @@ local mock = require("luassert.mock")
 local c = require("null-ls.config")
 local methods = require("null-ls.methods")
 local tu = require("null-ls.test-utils")
+local u = require("null-ls.utils")
 
 local lsp = mock(vim.lsp, true)
 local sources = mock(require("null-ls.sources"), true)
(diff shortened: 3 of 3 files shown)


arsham:
This fixes the issue, thank you!

przepompownia:
@jose-elias-alvarez accessing `client.resolved_capabilities` has been deprecated - is it hard to quick replace it by `server_capabilities` or any other equivalent?

arsham:
> @jose-elias-alvarez accessing `client.resolved_capabilities` has been deprecated - is it hard to quick replace it by `server_capabilities` or any other equivalent?

You should replace:
```
client.resolved_capabilities.document_formatting
```
With:
```
client.server_capabilities.documentFormattingProvider
```
And so on.

jose-elias-alvarez:
If you are specifically doing it for formatting, then you should use the new vim.lsp.buf.format API to only send requests to the servers you want, but yeah, you should still be able to modify server_capabilities and achieve the same result. 

This needs one small change that I’ll make in a couple of hours when I’m back at a computer but otherwise should be good to go, thanks for confirming!

arsham:
Sorry if this is off-topic, but I think maybe someone comes to this PR looking for the answer since it's been brought up.

Here is an example that works with the new API. Before calling the `on_attach` function:
```lua
client.server_capabilities.documentFormattingProvider = false
client.server_capabilities.documentRangeFormattingProvider = false
on_attach(client, bufnr)
```

And in the format function:
```lua
vim.lsp.buf.format({
  filter = function(clients)
    return vim.tbl_filter(
      function(cl) return cl.name == client.name end,
      clients
    )
  end
}, 2000) 
```

jose-elias-alvarez:
> Sorry if this is off-topic, but I think maybe someone comes to this PR looking for the answer since it's been brought up.

Not at all, I think a lot of people are going to come in here looking for this info, so I'm actually going to post it in the breaking changes issue. Thanks for the help!

arsham:
Works great, however I've faced a corner case with `callHierarchyProvider` capability with the following error:
```
null-ls.nvim/lua/null-ls/client.lua:38: bad
 argument #1 to 'unpack' (table expected, got nil)        
```

jose-elias-alvarez:
@arsham Is this the case even with the fix I merged? I added a check for that case, so it shouldn't be happening. If not, could you open an issue with reproduction instructions?

arsham:
@jose-elias-alvarez no it doesn't happen anymore. I'm guessing another plugin's update had something to do with it, which is not happening anymore. So all is good.

kylo252:
@jose-elias-alvarez, you probably missed adjusting another part of the readme in https://github.com/jose-elias-alvarez/null-ls.nvim/commit/b74bebd65cfa637f76f221a29dcf9cb31f007d0a in regards to `resolved_capabilities`

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b74bebd65cfa637f76f221a29dcf9cb31f007d0a/README.md#how-do-i-format-files-on-save

jose-elias-alvarez:
@kylo252 Thanks! I keep forgetting that the autocmd API got merged. I moved that whole section to the wiki for easier updating.

PR #852, opened 2022-05-06
Fix (builtins): stylelint format from stderr

volucris1:
Null-ls was ignoring stylelint stderr

### DIFF
--- a/lua/null-ls/builtins/formatting/stylelint.lua
+++ b/lua/null-ls/builtins/formatting/stylelint.lua
@@ -17,7 +17,6 @@ return h.make_builtin({
         args = { "--fix", "--stdin", "--stdin-filename", "$FILENAME" },
         to_stdin = true,
         from_stderr = true,
-        ignore_stderr = false,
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/formatting/stylelint.lua
```diff
@@ -16,6 +16,8 @@ return h.make_builtin({
         command = "stylelint",
         args = { "--fix", "--stdin", "--stdin-filename", "$FILENAME" },
         to_stdin = true,
+        from_stderr = true,
+        ignore_stderr = false,
```
jose-elias-alvarez, line 20:
```suggestion
        ignore_stderr = false,
```

`from_stderr` makes this line superfluous.

jose-elias-alvarez (commented):
Not sure how we missed this but I verified that it works!

jose-elias-alvarez:
Sorry, I think my comment maybe wasn't clear - we definitely need this fix, I just wanted to remove a specific line that wasn't needed, so I went ahead and fixed it up. Thanks for catching this and for putting up the PR!

PR #855, opened 2022-05-08
Add clj-kondo diagnostics

dhleong:


lua/null-ls/builtins/diagnostics/clj_kondo.lua
```diff
@@ -0,0 +1,42 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local on_output = h.diagnostics.from_patterns({
```
jose-elias-alvarez, line 6: Once small change: since we're only matching against a single pattern, could we use `from_pattern` instead?

dhleong: You bet! I saw another thing using patterns with the named params and it seemed nice, but nbd to me.

jose-elias-alvarez:
I went ahead and fixed up formatting to make CI happy. Thanks for the contribution!

PR #873, opened 2022-05-17
Fix mypy builtin not displaying some diagnostics

cblessing24:
The mypy builtin was not displaying some diagnostics due to failing to
parse them properly. The previously used regex pattern assumed that
each line in mypy's output would have a column number and a error code.
This is the case for most lines but not for all of them. For example the
following line contains no column number or error code:

```tests/slack_app/conftest.py:10: error: Unused "type: ignore" comment```

In this case null-ls would fail to display this error in the editor
because it could not be parsed. This commit fixes that issue by adding a
additional regex pattern that can parse lines in the format of the
example.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -3,6 +3,14 @@ local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
+local overrides = {
+    severities = {
+        error = h.diagnostics.severities["error"],
+        warning = h.diagnostics.severities["warning"],
+        note = h.diagnostics.severities["information"],
+    },
+}
+
 return h.make_builtin({
     name = "mypy",
     meta = {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jose-elias-alvarez** [comment]
I'm curious about the difference in the output format, does `mypy` natively produce diagnostics in both formats, or is there something else going on (e.g. a plugin)? Either way I think this fix looks reasonable.

**Comment by cblessing24** [comment]
I think mypy produces both formats natively. At least I am not aware of using any plugins myself. I posted a minimal example below if anyone wants to try for themselves.

Contents of `bar.py`:
```
import baz


def foo() -> float: 
    return 1.0 #type: ignore
```
Command:
```
mypy --show-error-codes --show-column-numbers --warn-unused-ignore bar.py
```
Output:
```
bar.py:1:1: error: Cannot find implementation or library stub for module named "baz"  [import]
bar.py:1:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
bar.py:5: error: Unused "type: ignore" comment
```
Note how the second line has no error code which makes sense because it is not an error. The third line has neither a column number nor an error code which is less understandable imo.

jose-elias-alvarez:
I see, that's pretty unusual but we should definitely support it if it's something `mypy` does natively. Thanks for the explanation and for the fix!

PR #874, opened 2022-05-17
style: add luacheck vim global variable

pedro757:


jose-elias-alvarez:
I realized that this isn't mentioned anywhere in the documentation, but we actually use [selene](https://github.com/Kampfkarren/selene) for linting (we have a CI action to run it on PRs). It's faster than luacheck, and we already have a selene config for `vim` and other globals, so I'll add a mention to the documentation.

pedro757:
Oh Thank you, I hadn't heard about selene before, I just gave it a try It's really fast, I will use this from now. Thanks.

PR #883, opened 2022-05-23
elixir credo integration: read stderr or stdout depending on output

emmanueltouzery:
if the elixir project has warnings, the elixir compiler will first output
warnings on stderr, which are NOT in json format and break the credo
integration => ignore stderr for the credo integration.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/credo.lua
+++ b/lua/null-ls/builtins/diagnostics/credo.lua
@@ -37,36 +37,42 @@ return h.make_builtin({
         args = { "credo", "suggest", "--format", "json", "--read-from-stdin", "$FILENAME" },
         format = "raw",
         to_stdin = true,
-        from_stderr = false,
         on_output = function(params, done)
             local issues = {}
 
-            -- report any unexpected errors, such as partial file attempts
-            if params.err then
-                table.insert(issues, generic_issue(params.err))
+            -- credo is behaving in a bit of a tricky way:
+            -- 1. if there are no elixir warnings, it will give its output
+            --    on stderr, and stdout will be nil
+            -- 2. if there are elixir warnings, it will report the elixir
+            --    warnings on stderr, and its own output on stdout
+            --
+            -- also note.. "warnings".. this has been reproduced by doing a
+            -- "mix new" project, then changing in mix.exs "def project" to add:
+            -- compilers: Mix.compilers(),
+            -- Then creating a file config/config.exs containing "use Mix.Config"
+            local output = params.err -- assume there are no elixir warnings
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Your description sounds reasonable, but the change in the code will make it so that null-ls will expect output from `stdout` and not `stderr`. Are you sure this is the right fix?

emmanueltouzery:
Yes. What is happening is that credo outputs JSON on stdout, while the elixir compiler outputs non-JSON warnings on stderr.
The reason others who contributed credo support to null-ls didn't need this is that the elixir codebase they worked on didn't have warnings.
I tested this in my case and confirmed that this change fixed my issues. I can paste null-ls debug output if you want?

emmanueltouzery:
Oh wow, I'll definitely try the case of a clean mix new project and update the PR.

emmanueltouzery:
OK so I checked it out... This is really strange and possibly a credo bug. You may not want to merge this, I would understand 100%. That wouldn't prevent me to use my fork of null-ls, and one day i'll supposedly fix this warning :)

so the comment is in the code, but to repeat it...

1. if there are no elixir warnings, it will give its output on stderr, and stdout will be nil
2. if there are elixir warnings, it will report the elixir warnings on stderr, and its own output on stdout

BUT. this may not be on just "any" warning. I can reproduce it with a warning in the configuration, which I think is quite low-level. This may not occur with other warnings.

To reproduce this...

```
$ mix new xx
$ cd xx
$ mkdir config
$ cat > config/config.exs
use Mix.Config
^D
(by ^D i mean press control-d)
```

modify mix.exs:
```diff
--- mix.exs	2022-05-24 20:39:53.887114519 +0200
+++ mix1.exs	2022-05-24 20:42:25.016265194 +0200
@@ -6,6 +6,7 @@
       app: :xx,
       version: "0.1.0",
       elixir: "~> 1.13",
+      compilers: Mix.compilers(),
       start_permanent: Mix.env() == :prod,
       deps: deps()
     ]
```

Then normally add credo to the project, and you'll reproduce the issue.

Honestly as I said, I now realise this seems like a freak issue so you may not want to merge this. I'll leave it up to you.

I also removed this:
```diff
-            -- report any unexpected errors, such as partial file attempts
-            if params.err then
-                table.insert(issues, generic_issue(params.err))
```

I don't think this is useful in this case.

jose-elias-alvarez:
Wow, I'm always amazed to see the variety of linter behavior in the wild. I'm not really familiar with the ecosystem, but your fix seems reasonable given the linter's behavior here, so let's give it a shot and see how it goes. Thanks!

PR #887, opened 2022-05-28
WIP Filtering Diagnostic Messages

nateroling:
This is a rough first pass at being able to filter diagnostic messages. Am I somewhere in the ballpark? 😅

I've been using it locally for the last couple weeks, seems to be working well. I was looking at adding unit tests but I was unable to get the e2e tests to pass locally. I figured I would get some feedback before I went too deep.

My vim config looks like this, with a new `diagnostics_filter` function parameter:

```
null_ls.setup({
	on_attach = function(client)
		if client.resolved_capabilities.document_formatting then
			-- format on save
			vim.cmd("autocmd BufWritePost <buffer> lua vim.lsp.buf.formatting_sync(nil, 2000)")
		end
	end,
	diagnostics_format = "[#{c}] #{m} (#{s})",
	sources = {
		null_ls.builtins.diagnostics.eslint_d.with({
			diagnostics_filter = function(diagnostic)
				if diagnostic.code == "prettier/prettier" then
					return false
				else
					return true
				end
			end,
		}),
		null_ls.builtins.code_actions.eslint_d,
		null_ls.builtins.formatting.eslint_d,
		null_ls.builtins.formatting.stylua,
		null_ls.builtins.formatting.prettierd,
	},
})
```

### DIFF
--- a/doc/BUILTIN_CONFIG.md
+++ b/doc/BUILTIN_CONFIG.md
@@ -178,15 +178,11 @@ be a function that returns `true` to keep the result, and `false` to remove it.
 
 ```lua
 local sources = {
-        null_ls.builtins.diagnostics.eslint_d.with({
-            -- ignore prettier warnings from eslint-plugin-prettier
-            filter = function(diagnostic)
-                if diagnostic.code == "prettier/prettier" then
-                    return false
-                else
-                    return true
-                end
-            end
+    null_ls.builtins.diagnostics.eslint_d.with({
+        -- ignore prettier warnings from eslint-plugin-prettier
+        filter = function(diagnostic)
+            return diagnostic.code != "prettier/prettier"
+        end
     }),
 }
 ```
--- a/doc/BUILTIN_CONFIG.md
+++ b/doc/BUILTIN_CONFIG.md
@@ -171,6 +171,26 @@ local sources = {
 }
 ```
 
+### Filtering
+
+You can filter generator results using the `filter` option. The option should
+be a function that returns `true` to keep the result, and `false` to remove it.
+
+```lua
+local sources = {
+        null_ls.builtins.diagnostics.eslint_d.with({
+            -- ignore prettier warnings from eslint-plugin-prettier
+            filter = function(diagnostic)
+                if diagnostic.code == "prettier/prettier" then
+                    return false
+                else
+                    return true
+                end
+            end
+    }),
+}
+```
+
 ### Diagnostics format
 
 For diagnostics sources, you can change the format of diagnostic messages by
--- a/doc/BUILTIN_CONFIG.md
+++ b/doc/BUILTIN_CONFIG.md
@@ -174,7 +174,7 @@ local sources = {
 ### Filtering
 
 You can filter generator results using the `filter` option. The option should
-be a function that returns `true` to keep the result, and `false` to remove it.
+be a function that returns `true` to keep the result, and `false` or `nil` to ignore it.
 
 ```lua
 local sources = {
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -76,15 +76,6 @@ local postprocess = function(diagnostic, _, generator)
     diagnostic.message = formatted
 end
 
-local filter = function(diagnostic, generator)
-    local user_filter = generator.opts.diagnostics_filter
-    if user_filter then
-        return user_filter(diagnostic)
-    else
-        return true
-    end
-end
-
 local handle_single_file_diagnostics = function(namespace, diagnostics, bufnr)
     vim.diagnostic.set(namespace, bufnr, diagnostics)
 end
@@ -170,7 +161,6 @@ M.handler = function(original_params)
         method = methods.map[method],
         params = params,
         postprocess = postprocess,
-        filter = filter,
         after_each = function(diagnostics, _, generator)
             if not vim.api.nvim_buf_is_valid(bufnr) then
                 return
--- a/lua/null-ls/diagnostics.lua
+++ b/lua/null-ls/diagnostics.lua
@@ -76,6 +76,15 @@ local postprocess = function(diagnostic, _, generator)
     diagnostic.message = formatted
 end
 
+local filter = function(diagnostic, generator)
+    local user_filter = generator.opts.diagnostics_filter
+    if user_filter then
+        return user_filter(diagnostic)
+    else
+        return true
+    end
+end
+
 local handle_single_file_diagnostics = function(namespace, diagnostics, bufnr)
     vim.diagnostic.set(namespace, bufnr, diagnostics)
 end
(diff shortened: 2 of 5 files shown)


jose-elias-alvarez:
Thanks for putting this in! I'll take a closer look and test later this week, but at a glance it seems to be on the right track. 

One idea: if we're implementing this at the generator level, maybe we could consider having this as an option for all sources, not only diagnostics? This is a pattern that's been added to some Neovim functions lately and I think it could be useful. Just a thought, though.

lua/null-ls/helpers/make_builtin.lua
```diff
@@ -29,6 +29,7 @@ local function make_builtin(opts)
         env = opts.env,
         cwd = opts.cwd,
         diagnostics_format = opts.diagnostics_format,
+        diagnostics_filter = opts.diagnostics_filter,
```
jose-elias-alvarez, line 32:
```suggestion
        filter = opts.filter,
```

lua/null-ls/generators.lua
```diff
@@ -57,6 +57,14 @@ M.run = function(generators, params, opts, callback)
             local ok, results = protected_call(to_run, copied_params)
             a.util.scheduler()
 
+            -- filter diagnostics results with the diagnostic_filter option
+            local filter = opts.filter
```
jose-elias-alvarez, line 61:
```suggestion
            local filter = generator.opts.filter
```

We can greatly simplify this by simply pulling the filter directly from the generator's options, meaning that we can immediately generalize this beyond just diagnostics.

lua/null-ls/generators.lua
```diff
@@ -57,6 +57,14 @@ M.run = function(generators, params, opts, callback)
...
+            if filter and results then
+                results = vim.tbl_filter(function(result)
+                    return filter(result, generator)
+                end, results)
```
jose-elias-alvarez, line 65:
```suggestion
                results = vim.tbl_filter(filter, results)
```

With the other suggested changes, we can make this into a more straightforward predicate (unless you think there's a reason to access `generator` when filtering, but I'd rather avoid a chance to accidentally mutate it).

jose-elias-alvarez (commented):
So in fact, this is actually very close to a generic filtering solution that we can use for all kinds of results, not just diagnostics. The one change we need is to get the `filter` callback directly from the generator's options, and other than that everything else falls into place.

If my suggestions make sense to you, let's go ahead with the changes and add documentation. If you're unable to get the test suite working (it can definitely be finicky) just ping me when you think this is otherwise ready and I'll add in some coverage.

nateroling:
Yeah, that sounds great. I committed those changes, I will see about adding some tests soon. Thanks!!

nateroling:
Not sure if two weeks later is "soon", but I did finally get a couple of simple tests added, and fixed a couple other failures I caused.

I also added documentation to BUILTIN_CONFIG.md

jose-elias-alvarez:
Awesome, thank you! At a glance everything looks okay. I'll take a closer look and test it out this weekend.

doc/BUILTIN_CONFIG.md
```diff
@@ -171,6 +171,26 @@ local sources = {
...
+
+You can filter generator results using the `filter` option. The option should
+be a function that returns `true` to keep the result, and `false` to remove it.
+
+```lua
+local sources = {
+        null_ls.builtins.diagnostics.eslint_d.with({
+            -- ignore prettier warnings from eslint-plugin-prettier
+            filter = function(diagnostic)
+                if diagnostic.code == "prettier/prettier" then
```
jose-elias-alvarez, line 184: I think we can simplify this example a bit:

```lua
return diagnostic.code == "prettier/prettier"
```

lua/null-ls/diagnostics.lua
```diff
@@ -76,6 +76,15 @@ local postprocess = function(diagnostic, _, generator)
     diagnostic.message = formatted
 end
 
+local filter = function(diagnostic, generator)
```
jose-elias-alvarez, line 79: I think this code is left over from the original implementation?

lua/null-ls/diagnostics.lua
```diff
@@ -161,6 +170,7 @@ M.handler = function(original_params)
         method = methods.map[method],
         params = params,
         postprocess = postprocess,
+        filter = filter,
```
jose-elias-alvarez, line 173: Same here, I don't think this is necessary anymore.

lua/null-ls/generators.lua
```diff
@@ -133,11 +139,11 @@ M.run_sequentially = function(generators, make_params, opts, callback)
 end
 
 M.run_registered = function(opts)
-    local filetype, method, params, postprocess, callback, after_each =
-        opts.filetype, opts.method, opts.params, opts.postprocess, opts.callback, opts.after_each
+    local filetype, method, params, postprocess, filter, callback, after_each =
```
jose-elias-alvarez, line 142: The changes in this function should also not be needed anymore, since we're pulling `filter` directly from the generator options.

doc/BUILTIN_CONFIG.md
```diff
@@ -171,6 +171,26 @@ local sources = {
 }
 ```
 
+### Filtering
+
+You can filter generator results using the `filter` option. The option should
+be a function that returns `true` to keep the result, and `false` to remove it.
```
jose-elias-alvarez, line 177:
```suggestion
be a function that returns `true` to keep the result, and `false` or `nil` to ignore it.
```

Minor wording, not a big deal either way.

jose-elias-alvarez (commented):
Works great! I think there's some leftover code from the previous implementation, but otherwise I think this is good to go.

doc/BUILTIN_CONFIG.md
nateroling: Good call. Switched to `!=` to return true for non-prettier items and updated.

lua/null-ls/diagnostics.lua
nateroling: You're right, good catch.

Spo0on:
Sorry for the dumb question, but I'm going crazy for this.
I'm not able to configure filtering for cpplint diagnostics. I would disable a boring diagnostic that hints me I didn't inserted the copyright message for every c file i edit.
I'm using mason-null-ls plugin.I tried adding these lines in my config:
```
require("mason-null-ls").setup_handlers({
	cpplint = function(source_name, methods)
		null_ls.register(null_ls.builtins.diagnostics.cpplint.with({
			filter = function(diagnostic)
				return diagnostic.code != "legal/copyright"
			end,
		}))
	end,
})
```
but it doesn't work.

fitrh:
@Spo0on **Not Equals** operator in lua is `~=`, not `!=`

jose-elias-alvarez:
Are you sure that's the exact code? I would try using a match instead.

fitrh:
Looks like the `cpplint` source does not include the diagnostic code,
You can try something like this

```lua
filter = function(diagnostic)
  return not diagnostic.message:match("No copyright")
end
```

PR #888, opened 2022-05-28
docs: document staticcheck and revive quirks

xenrox:
Those linters cannot correctly parse non-file arguments after the first
file is specified. This causes the extra_args feature of null-ls to not
work.

See discussion https://github.com/jose-elias-alvarez/null-ls.nvim/pull/399


jose-elias-alvarez:
Thanks! I think this will help avoid future misunderstandings.

PR #890, opened 2022-05-29
fix: luasnip not getting all available snippets

oncomouse:
As currently configured, this completion source misses all snippets not directly added to the `snippets` table, including those loaded with [`require("luasnip.loaders.from_lua")`](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#lua) and those loaded with [`require("luasnip.loaders.from_vscode")`](https://github.com/L3MON4D3/LuaSnip/blob/master/DOC.md#vscode). Using `get_snippets` method gets all available snippets, including those loaded by other methods.


jose-elias-alvarez:
Makes sense, thanks for the fix!

PR #891, opened 2022-05-30
docs: document staticcheck and revive quirks v2

xenrox:
Those linters cannot correctly parse non-file arguments after the first
file is specified. This causes the extra_args feature of null-ls to not
work.

Seems like I put the notes in the wrong, old place in version 1 and they got overwritten by the auto-generation of the documents in this [commit](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/82d5120c343b3e874da95946b1f6f1b79f72ae3b). Sorry for that sloppy mistake.


jose-elias-alvarez:
Sorry for merging this without a more thorough review but I think it should be okay now, thanks!

PR #898, opened 2022-06-06
feat: add cuda support for clang-format

cathaysia:
clang-format has the ability to format *.cu file


jose-elias-alvarez:
Thanks!

PR #899, opened 2022-06-08
Support XDG_RUNTIME_DIR for temp_files

joshuarubin:
When using the `buf` formatter, I was getting this error, and the formatter would fail to run:

```
[TRACE Tue 07 Jun 2022 06:41:45 PM MDT] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:217: error output: Failure: failed to enumerate module files: open /tmp/.<some_file_or_dir>: permission denied
```

I've made this modification locally as it seemed logical to use a per-user temp directory to address the issue. However, it could have been resolved similarly with subdirectories of `/tmp` as well.

### DIFF
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -203,13 +203,39 @@ M.timer = function(timeout, interval, should_start, callback)
     }
 end
 
-M.temp_file = function(content, extension)
-    local tmp_dir = u.path.is_windows and vim.fn.getenv("TEMP") or "/tmp"
+local tmp_dir
+local get_temp_dir = function()
+    if tmp_dir then
+        return tmp_dir
+    end
 
+    if u.path.is_windows then
+        tmp_dir = vim.env.TEMP
+        return tmp_dir
+    end
+
+    local xdg_runtime_dir = vim.env.XDG_RUNTIME_DIR
+    if xdg_runtime_dir then
+        tmp_dir = xdg_runtime_dir .. "/nvim"
+        if uv.fs_stat(tmp_dir) == nil then
+            uv.fs_mkdir(tmp_dir, tonumber("700", 8))
+        end
+        return tmp_dir
+    end
+
+    tmp_dir = "/tmp"
+    return tmp_dir
+end
+
+M._revert_tmp_dir = function()
+    tmp_dir = nil
+end
+
+M.temp_file = function(content, extension)
     local fd, tmp_path
     if uv.fs_mkstemp then
         -- prefer fs_mkstemp, since we can modify the directory
-        fd, tmp_path = uv.fs_mkstemp(u.path.join(tmp_dir, "null-ls_XXXXXX"))
+        fd, tmp_path = uv.fs_mkstemp(u.path.join(get_temp_dir(), "null-ls_XXXXXX"))
     else
         -- fall back to os.tmpname, which is Unix-only
         tmp_path = os.tmpname()
--- a/test/spec/loop_spec.lua
+++ b/test/spec/loop_spec.lua
@@ -582,12 +582,13 @@ describe("loop", function()
     describe("temp_file", function()
         local mock_content = "write me to a temp file"
         local mock_fd, mock_path = 57, "/tmp/null-ls-123456"
-
+        local xdg_runtime_dir = "/run/user/1000"
         local fs_mkstemp = uv.fs_mkstemp
         local tmpname
         before_each(function()
             u.path.is_windows = false
 
+            vim.env.XDG_RUNTIME_DIR = xdg_runtime_dir
             tmpname = stub(os, "tmpname")
 
             tmpname.returns(mock_path)
@@ -596,6 +597,8 @@ describe("loop", function()
         end)
         after_each(function()
             tmpname:revert()
+            vim.env.XDG_RUNTIME_DIR = nil
+            loop._revert_tmp_dir()
             uv.fs_write:clear()
             uv.fs_close:clear()
             uv.fs_unlink:clear()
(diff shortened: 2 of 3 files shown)


lua/null-ls/loop.lua
```diff
@@ -204,10 +204,13 @@ M.timer = function(timeout, interval, should_start, callback)
 end
 
 M.temp_file = function(content, extension)
-    local tmp_dir = u.path.is_windows and vim.fn.getenv("TEMP") or "/tmp"
+    local tmp_dir = u.path.is_windows and vim.fn.getenv("TEMP")
```
jose-elias-alvarez, line 207: This logic is starting to get a little complicated for a single line - I think we should extract it into a separate function.

test/spec/loop_spec.lua
```diff
@@ -581,7 +581,9 @@ describe("loop", function()
 
     describe("temp_file", function()
         local mock_content = "write me to a temp file"
-        local mock_fd, mock_path = 57, "/tmp/null-ls-123456"
+        local tmp_dir = vim.fn.getenv("XDG_RUNTIME_DIR") ~= nil and vim.fn.getenv("XDG_RUNTIME_DIR") .. "/nvim"
```
jose-elias-alvarez, line 584: I think we should instead stub `vim.fn.getenv` so we can test all the paths. Plenary is weird, so if you need help with this just let me know and I can take care of it when the rest of the PR is ready.

lua/null-ls/loop.lua
joshuarubin: Agreed. Done.

lua/null-ls/loop.lua
```diff
@@ -204,10 +204,13 @@ M.timer = function(timeout, interval, should_start, callback)
...
+        or vim.fn.getenv("XDG_RUNTIME_DIR") ~= nil and vim.fn.getenv("XDG_RUNTIME_DIR") .. "/nvim"
+        or "/tmp"
 
     local fd, tmp_path
     if uv.fs_mkstemp then
+        vim.fn.mkdir(tmp_dir, "p")
```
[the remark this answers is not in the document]
joshuarubin: Yup, done. Had to add this to allow testing with the cached value. Maybe there's a better way?
```lua
M.revert_tmp_dir = function()
    tmp_dir = nil
end
```

test/spec/loop_spec.lua
joshuarubin: It is weird, but now it's done.

lua/null-ls/loop.lua
```diff
@@ -203,13 +203,37 @@ M.timer = function(timeout, interval, should_start, callback)
...
+    if tmp_dir then
+        return tmp_dir
+    end
+
+    if u.path.is_windows then
+        tmp_dir = vim.fn.getenv("TEMP")
+        return tmp_dir
+    end
+
+    local xdg_runtime_dir = vim.fn.getenv("XDG_RUNTIME_DIR")
```
jose-elias-alvarez, line 217:
```suggestion
    local xdg_runtime_dir = vim.env["XDG_RUNTIME_DIR"]
```

While testing this, I found that `vim.fn.getenv` actually returns `vim.NIL`. If we use the `vim.env` accessor, we can avoid this issue.

Edit: also, this might make testing a bit simpler, since we can just set and clear the env vars. We can make the same change to the Windows `getenv` call for simplicity.

lua/null-ls/loop.lua
```diff
@@ -203,13 +203,37 @@ M.timer = function(timeout, interval, should_start, callback)
...
+    if xdg_runtime_dir then
+        tmp_dir = xdg_runtime_dir .. "/nvim"
+        uv.fs_mkdir(tmp_dir, 700)
```
jose-elias-alvarez, line 220: Out of curiosity, how does this behave if the directory already exists?

lua/null-ls/loop.lua
```diff
@@ -203,13 +203,37 @@ M.timer = function(timeout, interval, should_start, callback)
...
+        return tmp_dir
+    end
 
+    tmp_dir = "/tmp"
+    return tmp_dir
+end
+
+M.revert_tmp_dir = function()
```
jose-elias-alvarez, line 228: I think this is fine, let's just add a leading underscore to the name to make it clearer that it's private.

lua/null-ls/loop.lua
joshuarubin: I don't think it's an issue but I added a `stat` to make sure

test/spec/loop_spec.lua
```diff
@@ -582,20 +582,21 @@ describe("loop", function()
     describe("temp_file", function()
         local mock_content = "write me to a temp file"
         local mock_fd, mock_path = 57, "/tmp/null-ls-123456"
-
+        local xdg_runtime_dir = "/run/user/1000"
         local fs_mkstemp = uv.fs_mkstemp
-        local tmpname
+        local windows_temp_dir = "C:\\temp"
+        local getenv = stub(vim.fn, "getenv").on_call_with("TEMP").returns(windows_temp_dir)
```
jose-elias-alvarez, line 588: One small change we can make to simplify this: instead of stubbing `getenv`, we could just directly set values on the `vim.env` table and clear them between tests.

test/spec/loop_spec.lua
```diff
@@ -604,12 +605,18 @@ describe("loop", function()
...
         end)
 
-        it("should call uv.fs_mkstemp with path + pattern", function()
+        it("should call uv.fs_mkstemp with path + pattern using /tmp", function()
             loop.temp_file(mock_content)
-
             assert.stub(uv.fs_mkstemp).was_called_with("/tmp/null-ls_XXXXXX")
         end)
 
+        getenv.on_call_with("XDG_RUNTIME_DIR").returns(xdg_runtime_dir)
```
jose-elias-alvarez, line 613: I'm actually not sure exactly how Plenary / Busted will handle this, but I think it would be safer to move the setup into the specific test that depends on it.

joshuarubin: I was having a very hard time finding a way to get `getenv` to work in such a way that it could be cleaned up between tests (there were flakes, I think, due to conflicting stubs attaching to the same object) which is why I did it the way I did. If you have a better way, please go ahead and implement it.

test/spec/loop_spec.lua
joshuarubin: ok, I think I've simplified this to death. what do you think?

test/spec/loop_spec.lua
jose-elias-alvarez: I'll take a look tonight and see if I can figure something out, thanks for your work on this!

jose-elias-alvarez:
I updated the tests a bit and also added a mention of the new feature in the documentation. Thanks again for the contribution and for your patience with reviews!

joshuarubin:
Awesome, thanks a bunch!

PR #905, opened 2022-06-09
Add sql-formatter

timbedard:
sql-formatter seems to work decently well for SQL and is _much_ faster than sqlfluff for formatting.


jose-elias-alvarez:
Looks good, thanks for the contribution!

PR #906, opened 2022-06-10
add severities for common flake8 plugins

timbedard:
These are some commonly used plugins for flake8. It's tricky to customize the severities for them, and it doesn't hurt to have them here if they aren't installed, so might as well have them for folks that use them.


jose-elias-alvarez:
Makes sense, thanks!

PR #913, opened 2022-06-12
feat: added default runtime condition for cfn_lint diagnostics source

DeanGaffney:
- Added runtime_condition for cfn_lint diagnostics to only produce cfn-lint diagnostics when the file looks like an AWS Cloudfromation template.
- This prevents diagnostics from being run on all JSON and YAML files being edited.

**NOTE**
- @jose-elias-alvarez from our [discussion](https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/910) I found some behaviour which might not be as expected.

Screenshot of a YAML file where cfn-lint diagnostics are not being produced (the desired result), however the cfn-lint source still appears as attached. Not sure if this is expected.
image

Screenshot of adding `AWSTemplateFormatVersion` to the same YAML file, which then results in cfn-lint starting to produce diagnostics. I believe we were expecting the buffer cache to prevent cfn-lint from producing diagnostics as the cache would have previously marked the buffer as not being a cloud formation template, and we were expecting the user to have to reload the buffer for the cfn-lint diagnostics to be produced.

image

After removing `AWSTemplateFormatVersion` from the file, the file is still considered to be a cloudformation template which is fair enough, a restart of neovim fixes this.

I expect we have to add some documentation around the behaviour to this PR, but I want to verify the behaviour I am seeing with you first, and I can modify the PR as we need.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cfn_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/cfn_lint.lua
@@ -41,6 +41,13 @@ return h.make_builtin({
             end
             return false
         end),
+        notes = {
+            "Once a supported file type is opened null-ls will try and determine if the file looks like an AWS Cloudformation template.",
+            'A file will be considered an AWS Cloudformation template if it contains "Resources" or "AWSTemplateFormatVersion".',
+            "This check will run only once when entering the buffer.",
+            'This means if "Resources" or "AWSTemplateFormatVersion" is added to a file after this check is run, the cfn-lint diagnostics will not be generated.',
+            "To fix this you must restart neovim.",
+        },
     },
     factory = h.generator_factory,
 })
--- a/lua/null-ls/builtins/diagnostics/cfn_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/cfn_lint.lua
@@ -31,6 +31,16 @@ return h.make_builtin({
                 },
             }
         ),
+        runtime_condition = h.cache.by_bufnr(function(params)
+            -- check if file looks like a cloudformation template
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Yes, one limitation that I forgot to mention is that the sources will still show up in the `NullLsInfo` window. This is a pretty fundamental limitation of how sources work - basically, they'll always show up as attached if the filetype matches, but `runtime_condition` effectively prevents them from ever running. I don't personally think this is a huge issue, and it's not specific to this source, so I think we can leave it for now.

Your second issue made me realize that there is a small but significant bug in the current cache implementation, which I've fixed in e1c9b2321ece7adb8c16fb93cabea9381ebae1f5. This should make caching work as expected. I'd almost say we could leave out caching altogether and just run the scan each time, since the Neovim API methods we're using are pretty performant, but I also wouldn't want editing to slow to a crawl when, say, editing a gigantic JSON file with this source enabled, so I think it's safest to keep it on as a default.

I think all we need to do is add a mention to the source documentation about how the check will only run once when first entering the buffer. See [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/64d9e2b58a17cd85138c265ab816198000e126d2/lua/null-ls/builtins/code_actions/eslint_d.lua#L7) for an example of how we can notes like these, which will get picked up by the docgen script.

DeanGaffney:
@jose-elias-alvarez the caching is now working as expected when using cfn-lint, thank you for fixing that! I have updated the source notes for the cfn-lint diagnostics and this should be good for another review.

jose-elias-alvarez:
Great, thanks! I think the documentation should clarify potential issues, and I think this is a good improvement for better out-of-the-box behavior in spite of the edge cases.

Issue #909, opened 2022-06-11
Code Action for cspell: Add Word to Dictonary


yoo:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Program code often contains words / acronyms not known to cspell. cspell supports custom dictionaries https://cspell.org/docs/dictionaries-custom/. null-ls could support adding unknown words to cspells customs dictionary through code actions.  

### Help

Yes

jose-elias-alvarez:
This is a cool idea and is definitely possible. I think you'd have to do some work to get diagnostics specifically under the cursor, but you can get diagnostics for the current line using something like this:

```lua
local get_line_diagnostics = function()
    local lnum = vim.api.nvim_win_get_cursor(0)[1] - 1
    return vim.diagnostic.get(0, { lnum = lnum })
end
```

Appending words to a custom dictionary shouldn't be an issue, but based on the documentation for this feature (I don't actually use cspell, so I can't speak from experience) determining how this should behave seems tricky. It seems like there's a few different cases we'd have to handle:

1. We find a config file and it contains a single custom dictionary entry (this is the happy path, since we should be able to safely append entries to the file).
2. We find a config file, but it doesn't contain a dictionary entry.
3. We don't find a config file at all.

Happy to review and provide help with a PR if this is something you're interested in working on.

kaiuri:
I think, maybe, it'd be easier to write it as a completion source instead.

polyzen:
Implemented in #915 which was merged.

PR #915, opened 2022-06-13
feat: add cspell suggestions to code actions

yoo:
Add a new cspell code action builtin.
When triggering code actions on a line with cspell diagnostic errors
(misspelled words) show spelling suggestions in the code actions menu.

This is the followup to #909.

Example: 
The line `dhis menue` shows the code actions:
```
Fix dhis -> chis
Fix dhis -> dais
Fix dhis -> phis
Fix dhis -> this
Fix menue -> Benue
Fix menue -> mene
Fix menue -> menu
Fix menue -> menus
Fix menue -> venue
```

I'm not happy with the design yet. To generate the suggestions the command 
`cspell sug --num-suggestions=4 dhis menue` is run. Running the process 
takes some time and vim blocks while the command is running.

My preferred solution would be:  Generate the suggestions while the cspell diagnostic
builtin is run and the cspell code action builtin reads the suggestions from some sort of cache.
But I have no idea how to communicate between the two builtins, any ideas?

### DIFF
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -22,7 +22,7 @@ return h.make_builtin({
         notes = {
             "This source depends on the `cspell` built-in diagnostics source, so make sure to register it, too.",
         },
-        usage = "local sources = { null_ls.builtins.diagnostics.cspell, null_ls.builtins.code_actions.cspell }"
+        usage = "local sources = { null_ls.builtins.diagnostics.cspell, null_ls.builtins.code_actions.cspell }",
     },
     method = CODE_ACTION,
     filetypes = {},


jose-elias-alvarez:
I like this idea. If we could generate diagnostics and suggestions together (and assuming it doesn't significantly add to the time it takes to run the command) we could actually store the suggestion data under the `user_data` key (see `:h diagnostic-structure`) and retrieve it by adapting the logic you're already using. 

Since it looks like we'll have to depend on diagnostics in the first place, I think this would be the simplest way to handle it. This would require adapting the existing `cspell` source, and after playing around with it a bit, I wasn't able to get it to generate suggestions and diagnostics in a single command, but if you can figure it out, I think that would be the way to go.

yoo:
Spelling suggestions are now generated by the diagnostics cspell builtin. The code actions cspell builtin retrieves the suggestions from the `user_data` property of the diagnostic. 

The next question is if the code action message format is fine `Fix dhis -> this`?
And adding documentation. 

When this PR is accepted I would close #909 and follow up cspell configuration and custom dictionaries in other PRs.

lua/null-ls/builtins/diagnostics/cspell.lua
```diff
@@ -27,10 +47,13 @@ return h.make_builtin({
             return code <= 1
         end,
         on_output = h.diagnostics.from_pattern(
-            [[.*:(%d+):(%d+)%s*-%s*(.*%((.*)%))]],
-            { "row", "col", "message", "_quote" },
+            ".*:(%d+):(%d+)%s*-%s*(.*%((.*)%))%s*Suggestions:%s*%[(.*)%]",
```
jose-elias-alvarez, line 50: Nice solution - were you able to check if this has a noticeable effect on how long it takes to run `cspell`? Since diagnostics run frequently, small differences can add up (though I think this feature is valuable enough to justify it).

lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -0,0 +1,51 @@
...
+        url = "https://github.com/streetsidesoftware/cspell",
+        description = "cspell is a spell checker for code.",
+    },
+    method = CODE_ACTION,
+    filetypes = {},
+    generator = {
+        fn = function(params)
+            local actions = {}
+            local diagnostics = cspell_diagnostics(params.row - 1)
+            if next(diagnostics) == nil then
```
jose-elias-alvarez, line 28:
```suggestion
            if vim.tbl_isempty(diagnostics) then
```

lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -0,0 +1,51 @@
...
+                return nil
+            end
+            for _, diagnostic in ipairs(diagnostics) do
+                for _, suggestion in ipairs(diagnostic.user_data.suggestions) do
+                    table.insert(actions, {
+                        title = string.format("Fix %s -> %s", diagnostic.user_data.misspelled, suggestion),
```
jose-elias-alvarez, line 34: This looks fine to me - perhaps we can take guidance from the [VS Code cspell extension](https://github.com/streetsidesoftware/vscode-spell-checker).

lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -0,0 +1,51 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local CODE_ACTION = methods.internal.CODE_ACTION
+
+local cspell_diagnostics = function(lnum)
+    local diagnostics = {}
+    for _, diagnostic in ipairs(vim.diagnostic.get(0, { lnum = lnum })) do
```
jose-elias-alvarez, line 8: In practice, I don't think this will make a difference, but for consistency we should pass the buffer number from `params.bufnr` instead of defaulting to 0.

jose-elias-alvarez (commented):
Great! Left a few comments but I think this is a very nice addition. I agree that we should make sure the documentation clarifies the dependency on the diagnostic source.

lua/null-ls/builtins/diagnostics/cspell.lua
yoo: I have done some really quick and dirty testing on my laptop.

For a document with 8300 words and 225 errors, the time goes from 1.3s (no suggestions) to 12.5s (with suggestions).
After removing all but two errors from the text, now 8075 words, the time is 1.2s (no suggestions) and 1.3s (with suggestions). So each suggestion adds about 0.05s to the time.

lua/null-ls/builtins/code_actions/cspell.lua
yoo: I have seen `ltex` and other language servers use `Use <suggestion>`. I will change it to that.

lua/null-ls/builtins/code_actions/cspell.lua
yoo: fixed

lua/null-ls/builtins/code_actions/cspell.lua
yoo: Changed to "Use <suggestion>"

yoo:
I have changed the behavior a little bit.  You need to be on the misspelled word to show the code action, and only the suggestion for the word under the cursor is displayed.

lua/null-ls/builtins/diagnostics/cspell.lua
jose-elias-alvarez: Hmm, this is a pretty steep price to pay, especially for users of the diagnostics source who aren't using code actions. I don't think it's necessarily a deal-breaker, but it's worth considering alternatives if any exist. 

This is maybe a naive suggestion, but now that we're only getting suggestions from the word under the cursor, can we do something like this?

```lua
return h.make_builtin({
    -- other options go here
    generator_opts = {
        command = "cspell",
        args = function()
            -- get suggestions from the current word only
            return {
                "suggestions",
                vim.fn.expand("<cword>"),
            }
        end,
        on_output = function(params)
            -- handle output
        end,
    },
})
```

yoo: About 0.9s seem to be startup time. I think it's worse to wait for a second every time I want to correct a word than to add 0.2s to 0.5s to every diagnostic run, which happens in the background. The wait for the code action to appear seems more noticeable to the user. 

Is it possible to access the null-ls configuration and only generate the suggestions if the cspell code action builtin is enabled?
I will look into cspells caching. Maybe it can reduce the time on subsequent runs.

jose-elias-alvarez: I don't use `cspell` myself, so if you think the tradeoff is acceptable, then that works for me. You could conditionally set the arguments for the diagnostics source based on whether the code action source is also registered:

```lua
return h.make_builtin({
    -- other options removed for brevity
    generator_opts = {
        command = "cspell",
        args = function(params)
            local args = {
                "--language-id",
                params.ft,
                "stdin",
            }

            local should_add_suggestions = require("null-ls").get_source({
                name = "cspell",
                method = methods.internal.CODE_ACTION,
            }) ~= nil
            if should_add_suggestions then
                args = vim.list_extend({ "lint", "--show-suggestions" }, args)
            end

            return args
        end,
    },
})
```

If this works as expected, I think this is probably the solution that adds the least complexity while preserving the status quo for current `cspell` users.

marianozunino:
@yoo I've been using your branch, so far so good. How hard would be to add an option to the suggestion like "Add `word` to cspell.json"? I would love to contribute.

yoo:
The suggestions are now only generated if the cspell code actions builtin is enabled. And added documentation.

yoo:
@marianozunino I would do this in a follow-up PR. 

My thoughts:
Set the path to the `config.json` as a config option. This way, the user can decide if they want 
a single global config file or lookup the current path.

flytaly:
As for adding words, vscode extensions also support dictionaries

Example:

doc/BUILTINS.md
```diff
@@ -8,6 +8,18 @@ See [BUILTIN_CONFIG](BUILTIN_CONFIG.md) to learn how to set up and configure the
 
 ## Code Actions
 
+### [cspell](https://github.com/streetsidesoftware/cspell)
+
+Injects actions to fix typos found by `null_ls.builtins.diagnostics.cspell`.
```
jose-elias-alvarez, line 13: You'll need to add this info to the [source itself](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONTRIBUTING.md#contributing-built-ins) so that it doesn't get wiped out.

lua/null-ls/builtins/diagnostics/cspell.lua
```diff
@@ -14,11 +32,23 @@ return h.make_builtin({
...
+                "lint",
                 "--language-id",
                 params.ft,
                 "stdin",
             }
+
+            local should_add_suggestions = require("null-ls").get_source({
+                name = "cspell",
+                method = methods.internal.CODE_ACTION,
+            }) ~= nil
```
jose-elias-alvarez, line 45:
```suggestion
            local should_add_suggestions = not vim.tbl_isempty(require("null-ls").get_source({
                name = "cspell",
                method = methods.internal.CODE_ACTION,
            }))
```

I am not good at naming API methods (and I need to think about changing the name completely, since this mistake is too easy to make).

jose-elias-alvarez (commented):
Looks good and works well! I left a couple of comments but I think this is almost good to go. 

Edit: I also agree on waiting on a follow-up PR to add more functionality.

PumpedSardines:
Are there any updates on this feature?

lua/null-ls/builtins/diagnostics/cspell.lua
yoo: Rebased the change into the right commit to keep the history clean.

PumpedSardines:
Hurray, thanks a lot for this feature!

juanpprieto:
> @yoo I've been using your branch, so far so good. How hard would be to add an option to the suggestion like "Add `word` to cspell.json"? I would love to contribute.
> 
> Edit: I added the following to your snippet, it does the trick and appends the word to the `.json` file.
> 
> ```lua
>                 -- add word to "words" in cspell.json
>                 table.insert(actions, {
>                     title = "Add to cspell.json",
>                     action = function()
>                         local word = vim.api.nvim_buf_get_text(
>                             diagnostic.bufnr,
>                             diagnostic.lnum,
>                             diagnostic.col,
>                             diagnostic.end_lnum,
>                             diagnostic.end_col,
>                             {}
>                         )[1]
>                         local cspell_json = vim.fn.findfile("cspell.json", vim.fn.getcwd() .. ";")
>                         if cspell_json == "" then
>                             vim.notify("\ncspell.json not found", vim.log.levels.WARN)
>                             return
>                         end
>                         local cspell = vim.fn.json_decode(vim.fn.readfile(cspell_json))
>                         if not cspell.words then
>                             cspell.words = {}
>                         end
>                         table.insert(cspell.words, word)
>                         local json = vim.fn.json_encode(cspell)
>                         vim.fn.writefile({ json }, cspell_json)
> 
>                         -- replace word in buffer to trigger cspell to update diagnostics
>                         vim.api.nvim_buf_set_text(
>                             diagnostic.bufnr,
>                             diagnostic.lnum,
>                             diagnostic.col,
>                             diagnostic.end_lnum,
>                             diagnostic.end_col,
>                             { word }
>                         )
>                     end,
>                 })
> ```

@marianozunino @yoo  was this feature officially implemented? — thanks for this (love cspell)

jose-elias-alvarez:
@juanpprieto There is an open PR - see #1154.

PR #920, opened 2022-06-17
feat: define actionlint runtime_condition

hori-ryota:
add `runtime_condition` to actionlint with cache

resolves #804
resolves #776


hori-ryota:
@jose-elias-alvarez could you review this?

jose-elias-alvarez:
Alright, this looks good - I'm not sure how it'll behave on Windows, but I think this is a nice improvement as-is (and we can deal with Windows issues separately if they arise).

sitiom:
> Alright, this looks good - I'm not sure how it'll behave on Windows

Indeed it doesn't work in Windows, it requires a backward slash for the path (#986)

PR #925, opened 2022-06-20
feat(builtins): rego files diagnostics with opa check

### LINKED ISSUE #924: Add support for linting rego files via opa cli
### Issues

### Feature description

It will be useful to have support for `*.rego` files for [OPA engine](https://www.openpolicyagent.org/) that can be linted with `opa check` command.

### Help

Yes

### Implementation help

_No response_

eshepelyuk:
Resolves #924

### DIFF
--- a/lua/null-ls/builtins/diagnostics/opacheck.lua
+++ b/lua/null-ls/builtins/diagnostics/opacheck.lua
@@ -1,0 +1,45 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local handle_opacheck_output = function(params)
+    local diags = {}
+    if params.output.errors == nil then
+        return diags
+    end
+    for _, d in ipairs(params.output.errors) do
+        if d.location ~= nil then
+            table.insert(diags, {
+                row = d.location.row,
+                col = d.location.col,
+                source = "opacheck",
+                message = d.message,
+                severity = vim.diagnostic.severity.ERROR,
+                filename = d.location.file,
+                code = d.code,
+            })
+        end
+    end
+    return diags
+end
+
+return h.make_builtin({
+    name = "opacheck",
+    meta = {
+        url = "https://www.openpolicyagent.org/docs/latest/cli/#opa-check",
+        description = "Check Rego source files for parse and compilation errors.",
+    },
+    method = methods.internal.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "rego" },
+    generator_opts = {
+        command = "opa",
+        args = { "check", "-f", "json", "--strict", "$ROOT/src/", "$ROOT/test/" },
+        format = "json",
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/opacheck.lua
```diff
@@ -0,0 +1,49 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local handle_opacheck_output = function(params)
+    local diags = {}
+    if params.output.errors == nil then
+        return diags
+    end
+    for _, d in ipairs(params.output.errors) do
+        if d.location ~= nil then
```
jose-elias-alvarez, line 10: I'm curious - what do diagnostics without locations look like? Are they still worth showing? If we leave the location empty, they'll show up at the top of the file (which may or may not be useful).

lua/null-ls/builtins/diagnostics/opacheck.lua
```diff
@@ -0,0 +1,49 @@
...
+        url = "https://www.openpolicyagent.org/docs/latest/cli/#opa-check",
+        description = "Check Rego source files for parse and compilation errors.",
+    },
+    method = methods.internal.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "rego" },
+    generator_opts = {
+        command = "opa",
+        args = { "check", "-f", "json", "--strict", "$ROOT/src/", "$ROOT/test/" },
+        format = "json",
+        to_stdin = false,
```
jose-elias-alvarez, line 37: No need to explicitly specify these `false` options, so we can take them out.

jose-elias-alvarez (commented):
Just a couple of clarifying questions but this looks good, thank you!

lua/null-ls/builtins/diagnostics/opacheck.lua
eshepelyuk: Should i remove all the declarations with false value ?

lua/null-ls/builtins/diagnostics/opacheck.lua
eshepelyuk: i only noticed one case - when incorrect parameters are passed to opa check.

lua/null-ls/builtins/diagnostics/opacheck.lua
jose-elias-alvarez: Yep, not a major issue but I think that would be the cleanest.

eshepelyuk: ok, will do

eshepelyuk: @jose-elias-alvarez done

jose-elias-alvarez:
Thanks!

PR #935, opened 2022-06-28
fix: added stricter pattern matching for cfn-lint runtime_condition

DeanGaffney:
- Added stricter pattern matching for cfn-lint runtime_condition function when determining if a file is a cloud formation template or not.
- Based on the following implementation used for the vs-code extension https://github.com/aws-cloudformation/cfn-lint-visual-studio-code/blob/master/server/src/server.ts#L144
	- This is not an exact match of this regex in the above link, the patterns in this PR have been modified slightly in some cases (shortened to use meta characters, being stricter by looking for an "AWS" prefix etc.)

- Screenshot of a cloud formation file with resource types defined under the Resources key (json).
image

- Screenshot of a cloud formation file with resource types defined under the Resources key (yaml).
image

- Screenshot of AWSTemplateVersionFormat being used to attach cfn-lint to signal that the "Resources" key is invalid (json)
image

- Screenshot of AWSTemplateVersionFormat being used to attach cfn-lint to signal that the "Resources" key is invalid (yaml)
image

- Standard json file with "Resouces" key not producing cfn-lint diagnostics
image

- Standard yaml file with "Resouces" key not producing cfn-lint diagnostics
image

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cfn_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/cfn_lint.lua
@@ -10,7 +10,9 @@ return h.make_builtin({
         description = "Validate AWS CloudFormation yaml/json templates against the AWS CloudFormation Resource Specification",
         notes = {
             "Once a supported file type is opened null-ls will try and determine if the file looks like an AWS Cloudformation template.",
-            'A file will be considered an AWS Cloudformation template if it contains "Resources" or "AWSTemplateFormatVersion".',
+            'A file will be considered an AWS Cloudformation template if it contains a "Resources" or "AWSTemplateFormatVersion" key.',
+            'To prevent cfn-lint running on all YAML and JSON files that contain a "Resources" key.',
+            'The file must contain at least one AWS Cloudformation Resource Type, e.g "Type": "AWS::S3::Bucket"',
             "This check will run only once when entering the buffer.",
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks! This seems reasonable and clearly documented. @jorge-ramirezs Could you verify that this fixes your issue?

jorge-ramirezs:
The screen shots seem reasonable, but I might be doing something wrong when testing locally.

null-ls-info

I cloned Dean's repo and installed the plugin locally using packer. I'm still seeing the false positive.

null-ls-test

I appreciate any help on  how should I test the changes, if you want me to do so.

Thank you

DeanGaffney:
Hey @jorge-ramirezs, it looks like you have everything correct from your screenshots.

Have you checked out the cfn-lint-pattern-fix branch on my fork of the repository? I don't believe I put the pattern fix into my main branch. If you're using the main branch, check out the branch I mentioned above and give it another go.

If the problem still persists I will work with you to try and fix the problem.

jorge-ramirezs:
@DeanGaffney Thanks for your quick response. You are totally right, I switched to "cfn-lint-pattern-fix" in my local and now the false positive in the gitlab-ci.yml file is gone.

null-ls-OK

Thank you

PR #942, opened 2022-07-04
fix: call vim notify from logger

eshepelyuk:
Hello @jose-elias-alvarez 

This is a follow up to #892.

### DIFF
--- a/lua/null-ls/config.lua
+++ b/lua/null-ls/config.lua
@@ -9,16 +9,14 @@ local defaults = {
     default_timeout = 5000,
     diagnostics_format = "#{m}",
     fallback_severity = vim.diagnostic.severity.ERROR,
-    ---@usage setting this to true will enable debug logging
+    ---@usage setting this to nil or false will disable logging completely
     log = {
-        ---@usage disable logging completely
-        enable = true,
-        ---@usage set logging level
+        ---@usage logging level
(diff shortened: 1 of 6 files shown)


lua/null-ls/config.lua
```diff
@@ -19,6 +19,8 @@ local defaults = {
         ---@usage whether to print the output to neovim's
         --- possible values: 'sync','async', false
         use_console = "async",
```
jose-elias-alvarez, line 21: I wonder if it might make sense to add `notify` as a possible value for `use_console` instead of a separate config option, since it doesn't seem to make a lot of sense to enable both, right?

jose-elias-alvarez (commented):
At a glance this looks good, I'll test this out and leave a more thorough review over the weekend.

lua/null-ls/config.lua
eshepelyuk: Then, what about removing `use_console` at all, i.e. setting it explicitely to `false` when setting up plenery logging ?

eshepelyuk: Also, I was thinking about just introducing 2 level options
* one for plenary file logging
" another is for notify logging
again, having i mind removing use_console at all

ievgenii-shepeliuk: Ok will take care.

PR #946, opened 2022-07-09
feat(builtins): add support for pint

smoka7:
[pint](https://github.com/laravel/pint) is a  php code formatter.

### DIFF
--- a/lua/null-ls/builtins/formatting/pint.lua
+++ b/lua/null-ls/builtins/formatting/pint.lua
@@ -1,0 +1,29 @@
+local h = require("null-ls.helpers")
+local u = require("null-ls.utils")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "pint",
+    meta = {
+        url = "https://github.com/laravel/pint",
+        description = "An opinionated PHP code style fixer for minimalists.",
+    },
+    method = FORMATTING,
+    filetypes = { "php" },
+    generator_opts = {
+        command = "./vendor/bin/pint",
+        args = {
+            "--no-interaction",
+            "--quiet",
+            "$FILENAME",
+        },
+        cwd = h.cache.by_bufnr(function(params)
+            return u.root_pattern("pint.json", "composer.json", "composer.lock")(params.bufname)
+        end),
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/formatting/pint.lua
```diff
@@ -0,0 +1,25 @@
...
+return h.make_builtin({
+    name = "pint",
+    meta = {
+        url = "https://github.com/laravel/pint",
+        description = "An opinionated PHP code style fixer for minimalists.",
+    },
+    method = FORMATTING,
+    filetypes = { "php" },
+    generator_opts = {
+        command = "./vendor/bin/pint",
```
jose-elias-alvarez, line 15: So a potential issue here is that this will be resolved relative to the null-ls root directory, which (unless set by the user) will [generally default to a Git root](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#options), which may not always be correct.

Two options that I can think of:
1. If `composer` projects always contain a specific config file at the root level, we can use that as the `cwd` of the spawned process, similar to what we do for [Prettier](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/0d68b5bffa2e15ff06b4dc42b293a0f6e3060da5/lua/null-ls/builtins/formatting/prettier.lua#L45) and `package.json`.
2. We can use [one of our command resolver helpers](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/fbb1929b29beff82e0fc495670f00ef4b3bcbcd3/lua/null-ls/helpers/command_resolver.lua#L25) to find the nearest `pint` executable.

Alternatively, if you think it's fine as-is, I think we should at least add a mention to the documentation to clarify that `pint` should be installed with `composer` at the root level for this to work.

smoka7: Good Point. It's not very common to install composer binaries globally so I'm gonna go with option 1.

jose-elias-alvarez: Perfect, thanks!

jose-elias-alvarez:
I went ahead and rebased this to fix unrelated formatting issues that were causing CI to complain. Thanks for the contribution!

PR #947, opened 2022-07-10
feat: Added csharpier formatter

eugen-meissner:
[csharpier](https://csharpier.com/)

### DIFF
--- a/lua/null-ls/builtins/formatting/csharpier.lua
+++ b/lua/null-ls/builtins/formatting/csharpier.lua
@@ -1,0 +1,22 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "csharpier",
+    meta = {
+        url = "https://csharpier.com/",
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/formatting/csharpier.lua
```diff
@@ -0,0 +1,23 @@
...
+    meta = {
+        url = "https://csharpier.com/",
+        description = "CSharpier is an opinionated code formatter for c#",
+    },
+    method = FORMATTING,
+    filetypes = { "cs" },
+    generator_opts = {
+        command = "dotnet-csharpier",
+        args = {
+            "$FILENAME",
```
jose-elias-alvarez, line 17: Just to verify: this argument structure makes me think that the linter will read from the file on the disk, not from `stdin`. According to [the documentation](https://csharpier.com/docs/CLI#directoryorfile), it looks like we may want to leave this argument out completely in order to use `stdin` as a source. Is that correct?

jose-elias-alvarez:
Also, the failing CI step is unrelated to this PR. If you simply rebase onto `main` you should be good to go.

jose-elias-alvarez:
Thanks for the update!

Issue #859, opened 2022-05-11
Allow calling `setup()` multiple times


AlxHnr:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Most other lua plugins have a `setup()` function which can be called multiple times. This feature is really nice because it allows reloading your `init.lua` at runtime just by running `:so %`. But with null-ls I have to restart neovim on every change I want to try. Each time I want to enable a new linter, or change my existing null-ls settings, I have to restart again and again.

Adding this feature would make null-ls more consistent with the rest of neovim and its ecosystem. Every neovim setting and almost every other plugin can be configured at runtime without a restart.

### Help

No

AlxHnr:
I see, the source API solves my problem. Not sure why I couldn't find it at first glance. Is there a reason why null-ls doesn't ship a vim-help page?

AlxHnr:
If you want to keep `setup()` only callable once, feel free to close this issue. Its up to you to decide on the design philosophy.

For anyone else who insists on hot-reloading their `init.lua`, and happens to stumble upon this discussion, I've pasted a solution below. That way they can for example comment out a linter, run `:so %` and its gone.

```lua
-- init.lua
local null_ls = require('null-ls')
null_ls.setup()
null_ls.disable{}
null_ls.reset_sources()
null_ls.register(null_ls.builtins.diagnostics.shellcheck)
```

`on_attach` could be redefined at runtime by registering a wrapper lambda.

jose-elias-alvarez:
I think what you've posted is a good workaround. As I said, I'm not totally opposed to reviewing a PR that accomplishes this without adding complexity, so I'll keep this open in case anybody wants to pick it up.

jose-elias-alvarez:
I'm closing this since I don't plan to work on it, but anybody who's really interested and is willing to put in the work should feel free to open a PR.

PR #949, opened 2022-07-11
WIP: showcase automatic vimdoc conversion with ts-vimdoc

ibhagwan:
Hi @jose-elias-alvarez,

I'm sure you've heard this before many times, nevertheless, thank you for your wonderful work :-)

Browsing through the issues (to familirize myself better with the plugin and its lesser known options), I noticed this comment from: https://github.com/jose-elias-alvarez/null-ls.nvim/issues/859#issuecomment-1124124737:

> Because it takes a lot of work to write a separate help page + the existing documentation. There's scripts to convert them automatically but I haven't had the time to look into them, and given the structure of our documentation I'm not sure the conversion would produce good results.

Remebering I also had a similar stance against automated conversions (you can read my initial reaction to the idea... https://github.com/ibhagwan/fzf-lua/pull/81) I thought I'd showcase what I was able to achieve with my improved version of what used to be `babelfish.nvim`, now named [`ts-vimdoc`](https://github.com/ibhagwan/ts-vimdoc.nvim).

At the end of the day I didn't see myself maintining a full-on separate vimdoc file (and I don't think you want that either) and the conversion tools are a very decent compromise.

IMHO, the conversion came out quite nice, aside from some very long titles which can be adjusted in the `.md` file I think it's very workable, the file `null-ls.txt` is the conversion of the main `README.md`, the rest of the txt files under `/doc` are self explanatory, all were generated pretty easily by doing this (which can also be turned into github CI action):
```#
git clone https://github.com/jose-elias-alvarez/null-ls.nvim
git clone https://github.com/ibhagwan/ts-vimdoc.nvim
cd null-ls.nvim
../ts-vimdoc.nvim/scripts/docgen.sh README.md ./doc/null-ls.txt null-ls
../ts-vimdoc.nvim/scripts/docgen.sh doc/CONFIG.md doc/CONFIG.txt null-ls
<repeat for all .md files>
```

There's also the much more mature [`panvimdoc`](https://github.com/kdheepak/panvimdoc) by @khdeepak which I highly recommend, test it out and see which output fits your needs better.

If you don't feel this meets the project's goals, I totally understand but I do beleive there's some added value in having the documentation searchable through `:help`.

### DIFF
--- a/.github/workflows/vimdoc.yaml
+++ b/.github/workflows/vimdoc.yaml
@@ -1,0 +1,60 @@
+name: vimdoc
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  vimdocgen:
+    runs-on: [ubuntu-latest]
+    steps:
+      - uses: actions/checkout@v2
+      - run: date +%F > todays-date
+      - name: Restore cache for today's nightly.
+        uses: actions/cache@v2
+        with:
+          path: build
+          key: ${{ runner.os }}-appimage-${{ hashFiles('todays-date') }}
+
+      - name: Setup neovim nightly and install plugins
+        run: |
+          test -d build || {
+            mkdir -p build
+            wget https://github.com/neovim/neovim/releases/download/nightly/nvim.appimage
+            chmod +x nvim.appimage
+            mv nvim.appimage ./build/nvim
+          }
+          mkdir -p ~/.local/share/nvim/site/pack/vendor/start
+          git clone --depth 1 https://github.com/ibhagwan/ts-vimdoc.nvim ~/.local/share/nvim/site/pack/vendor/start/ts-vimdoc.nvim
+          git clone --depth 1 https://github.com/nvim-treesitter/nvim-treesitter ~/.local/share/nvim/site/pack/vendor/start/nvim-treesitter
+      - name: Build parser
+        run: |
+          export PACKPATH=$HOME/.local/share/nvim/site
+          ./build/nvim --headless -u ~/.local/share/nvim/site/pack/vendor/start/ts-vimdoc.nvim/scripts/init.lua -c "TSUpdateSync markdown" -c "TSUpdateSync markdown_inline" -c "qa"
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Wow, this is great! Thanks for putting in the time to set this up. I agree completely - I think it's super valuable to have the documentation in `:help`, even if it's not 100% perfect. 

I will take a closer look at this over the weekend, see how I can adjust the source files to better account for help file generation, and see if I can set up a CI action, too.

ibhagwan:
Glad you liked it :)

> and see if I can set up a CI action, too.

Lmk if you’re interested and I’ll setup the CI action on this PR’s fork, might save you a few minutes?

jose-elias-alvarez:
> Lmk if you’re interested and I’ll setup the CI action on this PR’s fork, might save you a few minutes?

That would be a huge help, thank you!

ibhagwan:
Added the workflow but couldn't really test it as it won't run on my fork or here, I think it should work, I'm sure you'll figure it out :-)

jose-elias-alvarez:
Alright, I tried this out locally and it's super simple. I don't have a lot of bandwidth to pick this up now, but I think the next step is to tweak our Markdown files so that they work better as Vim help files, either manually or by modifying how we generate them in the first place. I also think some of the files are not necessarily relevant, so I want to see how other sources like ALE handle these cases and think about what users might want to see within Neovim. 

Thanks again for the PR! Once I get more time to work on this, I'll follow up on the ts-vimdoc repo about a couple of minor issues I've seen in the generated documentation / potential features.

mskelton:
I'm taking a stab at forking null-ls given the recent announcement that null-ls is being archived, and I've implemented this in https://github.com/mskelton/null-ls.nvim/commit/62336fcd

One thing I noticed is that it doesn't handle the multiple files very well since it results in a lot of conflicting tag names. I think some of this could be resolved by merging the docs somewhat, but for the sources that will still be a problem.

PR #952, opened 2022-07-14
Fix fourmolu formatting builitin

bradsherman:
I upgraded to Fourmolu 0.7.0.1 and it requires the `--stdin-input-file` argument when using stdin.


jose-elias-alvarez:
Thanks for the fix!

PR #954, opened 2022-07-16
fix: diagnostics.stylelint not using local binary

LKummer:
Hi, thanks for making this awesome plugin!

I noticed `builtins.diagnostics.stylelint` does not look up local installations of Stylelint in `node_modules`.
I fixed it by using the `from_node_modules` command resolver like `builtins.formatting.stylelint` already does.


jose-elias-alvarez:
Makes sense, thank you!

PR #959, opened 2022-07-19
feat: add glslc diagnostics

daangoossens22:
Shows the diagnostics provided by glslc when validating a shader.

It needs the shader stage for it to work, which is explained in the notes.


jose-elias-alvarez:
Looks great, thank you! Really appreciate the documentation and tests!

PR #963, opened 2022-07-20
 feat(builtins): add support for blade-formatter

gagansday:
## blade-formatter

An opinionated blade template formatter for Laravel that respects readability


jose-elias-alvarez:
Thank you!

PR #975, opened 2022-07-29
feat: add packer fmt

tom-on-the-internet:
Adds formatting for [Hashicorp Packer](https://www.packer.io/)

Ideally, I would like this to only target `.packer.hcl` files, but I'm unsure how to do this.

Currently, null-ls doesn't target `hcl` files, so this will not cause issues.

Would love feedback.

Thanks for null-ls.

### DIFF
--- a/lua/null-ls/builtins/formatting/packer.lua
+++ b/lua/null-ls/builtins/formatting/packer.lua
@@ -1,0 +1,27 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "packerfmt",
+    meta = {
+        url = "https://www.packer.io/docs/commands/fmt",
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thank you for the PR! We don't have a perfect way to only target files with a specific name, but we could use the [`runtime_condition`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTIN_CONFIG.md#runtime_condition) option to make sure the source only runs on files that match a given pattern. It would look something like this:

```lua
return h.make_builtin({
    -- other options abbreviated
    generator_opts = {
        runtime_condition = function(params)
            -- untested but it should give you the right idea
            return params.bufname:match("%.packer%.hcl") ~= nil
        end,
    },
})
```

I think it would be best to go ahead and include this to avoid potential future conflicts with other sources that work on the same filetype, but let me know what you think.

thomas-humi:
Thanks so much. I've made the change and tested it locally. This works perfectly.

jose-elias-alvarez:
Awesome, thanks!

PR #978, opened 2022-07-31
feature: added goimports_reviser formatting tool

alesbrelih:
It's a similar tool to the goimports, but it works a bit better in some situations.

### DIFF
--- a/lua/null-ls/builtins/formatting/goimports_reviser.lua
+++ b/lua/null-ls/builtins/formatting/goimports_reviser.lua
@@ -1,0 +1,20 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "goimports_reviser",
+    meta = {
+        url = "https://pkg.go.dev/github.com/incu6us/goimports-reviser",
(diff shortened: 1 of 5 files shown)


lua/null-ls/builtins/formatting/goimports_reviser.lua
```diff
@@ -0,0 +1,27 @@
...
+    method = FORMATTING,
+    filetypes = { "go" },
+    generator_opts = {
+        command = "goimports-reviser",
+        args = {
+            "-file-path",
+            "$FILENAME",
+            "-output",
+            "stdout",
+            "-rm-unused",
```
jose-elias-alvarez, line 21: Question: should we always set these last two arguments as defaults, or should we leave it up to the user to add them if necessary?

jose-elias-alvarez (commented):
LGTM! Just one clarifying question about defaults.

lua/null-ls/builtins/formatting/goimports_reviser.lua
alesbrelih: You are completely right, those arguments are of a personal choice, I've removed them because they can be simply added with the with({ extra_args = ... }).

I've amended the existing commit so you won't have any problems with commit squashing if you decide to merge.

And of course.. Thank you for this plugin it really simplified my config 🙇

jose-elias-alvarez:
Great, thank you again!

Issue #790, opened 2022-03-31
BUILTIN: cue_fmt fails to run as diagnostic


spencergilbert:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.6.1 \ Build type: Release \ LuaJIT 2.1.0-beta3

### Operating System

Darwin arm64

### Minimal config

```lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
	local path_sep = on_windows and "\\" or "/"
	local result = table.concat({ ... }, path_sep)
	return result
end

vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir
if on_windows then
	temp_dir = vim.loop.os_getenv("TEMP")
else
	temp_dir = "/tmp"
end

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
	local null_ls = require("null-ls")
	null_ls.setup({
		sources = {
			null_ls.builtins.diagnostics.cue_fmt,
		},
		debug = true,
	})
end

local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				"jose-elias-alvarez/null-ls.nvim",
				requires = { "nvim-lua/plenary.nvim" },
				config = null_ls_config,
			},
		},
		config = {
			package_root = package_root,
			compile_path = compile_path,
		},
	})
end

if vim.fn.isdirectory(install_path) == 0 then
	vim.fn.system({ "git", "clone", "https://github.com/wbthomason/packer.nvim", install_path })
	load_plugins()
	require("packer").sync()
else
	load_plugins()
	require("packer").sync()
end

-- set cue filetype
vim.cmd([[autocmd BufNewFile,BufRead *.cue setf cue]])
```

### Steps to reproduce

[...]

PR #983, opened 2022-08-01
fix: cue diagnostic

johnallen3d:
It would appear that CUE moved from reporting format errors from `cue fmt` to `cue vet`.

Resolves https://github.com/jose-elias-alvarez/null-ls.nvim/issues/790


jose-elias-alvarez:
Thanks for identifying the issue and putting in the fix!

PR #984, opened 2022-08-02
feat: add ocdc formatter

mdwint:
This adds the [ocdc](https://github.com/mdwint/ocdc) formatter.

### DIFF
--- a/lua/null-ls/builtins/formatting/ocdc.lua
+++ b/lua/null-ls/builtins/formatting/ocdc.lua
@@ -1,0 +1,23 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "ocdc",
+    meta = {
+        url = "https://github.com/mdwint/ocdc",
+        description = "A changelog formatter",
+    },
+    method = FORMATTING,
+    filetypes = { "markdown" },
+    generator_opts = {
+        command = "ocdc",
+        args = { "--path", "-" },
+        to_stdin = true,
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/formatting/ocdc.lua
```diff
@@ -0,0 +1,23 @@
...
+        description = "A changelog formatter",
+    },
+    method = FORMATTING,
+    filetypes = { "markdown" },
+    generator_opts = {
+        command = "ocdc",
+        args = { "--path", "-" },
+        to_stdin = true,
+        runtime_condition = function(params)
+            return params.bufname:match("/CHANGELOG.md$")
```
mdwint, line 19: I'm new to Lua, so I'm not sure this is the best way. I want to match all paths that end in a filename of `CHANGELOG.md`. Will `params.bufname` always have a `/` before the filename?

jose-elias-alvarez: Hmm, this will work for Unix, but I'm not sure about Windows. I've heard different things about this (and don't have a Windows computer handy to test it on) but we could be defensive and also handle backslashes by using `[/\\]CHANGELOG.md$` as the pattern.

Also, should this be case-insensitive, or do we only want to match this precise filename?

mdwint: Thanks for your advice!

> Also, should this be case-insensitive, or do we only want to match this precise filename?

I'd only thought about matching the exact filename, but making it case-insensitive seems better.

jose-elias-alvarez:
Great, thank you for the contribution!

PR #986, opened 2022-08-03
fix: properly parse actionlint path for Windows

sitiom:
Use `fnamemodify` to convert the forward slashes to backward slashes in Windows.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -26,7 +26,7 @@ return h.make_builtin({
             },
         }),
         runtime_condition = h.cache.by_bufnr(function(params)
-            return params.bufname:find(vim.pesc(vim.fn.fnamemodify(".github/workflows", ":."))) ~= nil
+            return params.bufname:find("%.github[\\/]workflows") ~= nil
         end),
     },
     factory = h.generator_factory,
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -26,7 +26,7 @@ return h.make_builtin({
             },
         }),
         runtime_condition = h.cache.by_bufnr(function(params)
-            return params.bufname:find(vim.pesc(".github/workflows")) ~= nil
+            return params.bufname:find(vim.pesc(vim.fn.fnamemodify(".github/workflows", ":."))) ~= nil
         end),
     },
     factory = h.generator_factory,


lua/null-ls/builtins/diagnostics/actionlint.lua
```diff
@@ -26,7 +26,7 @@ return h.make_builtin({
             },
         }),
         runtime_condition = h.cache.by_bufnr(function(params)
-            return params.bufname:find(vim.pesc(".github/workflows")) ~= nil
+            return params.bufname:find(vim.pesc(vim.fn.fnamemodify(".github/workflows", ":."))) ~= nil
```
jose-elias-alvarez, line 29: More for my own curiosity, but which part of what we're doing here standardize directory separators? The documentation for `fnamemodify` and `filename-modifiers` doesn't seem to shed any light here, as far as I can tell. 

If this works I'm not opposed to it, but to make it clearer what's happening, I'd either add a comment or lean towards something like this, assuming it works:

```suggestion
            return params.bufname:find("%.github[\\/]workflows") ~= nil
```

sitiom: > More for my own curiosity, but which part of what we're doing here standardize directory separators? The documentation for `fnamemodify` and `filename-modifiers` doesn't seem to shed any light here, as far as I can tell.

Perhaps you are looking for the `shellslash` option:
```
'shellslash' 'ssl'	boolean	(default off)
			global
			{only for MS-Windows}
	When set, a forward slash is used when expanding file names.  This is
	useful when a Unix-like shell is used instead of cmd.exe.  Backward
	slashes can still be typed, but they are changed to forward slashes by
	Vim.
	Note that setting or resetting this option has no effect for some
	existing file names, thus this option needs to be set before opening
	any file for best results.  This might change in the future.
	'shellslash' only works when a backslash can be used as a path
	separator.  To test if this is so use:
		if exists('+shellslash')
```

> If this works I'm not opposed to it, but to make it clearer what's happening, I'd either add a comment or lean towards something like this, assuming it works:

That works too, up to you what looks simple to you 🙂

jose-elias-alvarez: I see, interesting! I'd lean towards putting this into the pattern, since it's more Lua-like (and theoretically a little bit faster) but I'm fine either way. Let me know when you think this is ready and I'll merge it.

jose-elias-alvarez:
Alright, let's try this and see how it works. Thanks again!

PR #987, opened 2022-08-03
docs: fix protolint url

mercxry:


jose-elias-alvarez:
Nice catch, thank you! Would you mind taking a second to fix this in the [source definition](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/protolint.lua)? The documentation is automatically generated from there, so that'll fix it once and for all.

mercxry:
Updated the source definition, let me know if there is something else to be done 😃

jose-elias-alvarez:
Should be good, thanks again!

PR #998, opened 2022-08-08
:wrench: add dprint to builtins formatter

hemedani:


hemedani:
what can I do to pass Check codestyle ?

lua/null-ls/builtins/formatting/dprint.lua
```diff
@@ -0,0 +1,35 @@
...
+		"toml",
+		"rust",
+		"roslyn",
+	},
+	generator_opts = {
+		command = "dprint",
+		args = {
+			"fmt",
+			"--config",
+			"https://raw.githubusercontent.com/hemedani/dotfiles/main/dprint/dprint.json",
```
jose-elias-alvarez, line 29: Are you sure that we want to include this into the built-in's definition? It seems like this should be left up to the user.

jose-elias-alvarez (commented):
Thanks for the PR! I had one clarifying question.

> what can I do to pass Check codestyle ?

Did you format the files with `stylua` 14.2 using this project's config? When I ran your file through it locally, it did something to the whitespace.

lua/null-ls/builtins/formatting/dprint.lua
hemedani: I'm using `stylua` 0.10.1,
yes it is better to leave 
```
"
            --config",
            "https://raw.githubusercontent.com/hemedani/dotfiles/main/dprint/dprint.json",
```
to the user.
but users should initialize dprint in every project with `dprint init`.

hemedani: I'm fixing these and pushing it again

jose-elias-alvarez: Sure, we can mention that users should run `dprint init` in the documentation.

jose-elias-alvarez:
Looks good! Based on what you mentioned, I think we should add a mention of the need to run `dprint init` in a project before running this, but let me know what you think.

hemedani:
I'm adding a note for that

jose-elias-alvarez:
Great, thank you!

PR #1008, opened 2022-08-14
feat(commitlint): add diagnostics for commit messages

wassimk:
[commitlint](https://commitlint.js.org/) checks if your commit messages meet the conventional commit format.

this is a cli tool that doesn't return line numbers. so, i've done some formatting to try and prevent all warnings from showing up on line 0, column 0.

you'll need commitlint and the json formatter packages:

``` shell
yarn global add @commitlint/{config-conventional,cli} 
yarn global add commitlint-format-json
```

then create a `commitlint.config.js` in the root of your project:

``` javascript
const Configuration = {
  extends: ["@commitlint/config-conventional"],
  formatter: "commitlint-format-json",
  rules: {
    "header-max-length": [2, "always", 72],
    "body-max-line-length": [2, "always", 72],
  },
};

module.exports = Configuration;
```

also, you can override the config file with an `extra_args` of `--config` to null-ls. if you don't have a config file, the first lint warning is a helpful guide to creating one.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -55,7 +55,7 @@ return h.make_builtin({
         command = "commitlint",
         args = { "--format", "commitlint-format-json" },
         to_stdin = true,
-        format = "json_raw",
+        format = "json",
         check_exit_code = function(code)
             return code <= 1
         end,
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -48,6 +48,10 @@ return h.make_builtin({
     meta = {
         url = "https://commitlint.js.org",
         description = "commitlint checks if your commit messages meet the conventional commit format.",
+        notes = {
+            "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
+            "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
+        },
     },
     method = DIAGNOSTICS,
     filetypes = { "gitcommit" },
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -57,7 +57,7 @@ return h.make_builtin({
         command = "commitlint",
         args = { "--format", "commitlint-format-json" },
         to_stdin = true,
-        format = "json",
+        format = "json_raw",
         check_exit_code = function(code)
             return code <= 1
         end,
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -1,0 +1,67 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local violation_to_diagnostic = function(violation)
+    local diagnostic = nil
+
+    diagnostic = {
+        ruleId = violation.name,
+        message = violation.message,
+        level = violation.level,
+    }
+
+    if violation.name == "body-leading-blank" then
+        diagnostic = vim.tbl_extend("force", diagnostic, { line = 2 })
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -0,0 +1,67 @@
...
+        url = "https://commitlint.js.org",
+        description = "commitlint checks if your commit messages meet the conventional commit format.",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "gitcommit" },
+    generator_opts = {
+        command = "commitlint",
+        args = { "--format", "commitlint-format-json" },
+        to_stdin = true,
+        format = "json_raw",
```
jose-elias-alvarez, line 60:
```suggestion
        format = "json",
```

We should only use `json_raw` if we are planning on handling errors ourselves (in this case, this suppressed an error I was getting from the CLI, making it harder to figure out what was going on).

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -0,0 +1,67 @@
...
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local violation_to_diagnostic = function(violation)
+    local diagnostic = nil
+
+    diagnostic = {
+        ruleId = violation.name,
+        message = violation.message,
+        level = violation.level,
+    }
```
jose-elias-alvarez, line 13:
```suggestion
    local diagnostic = {
        ruleId = violation.name,
        message = violation.message,
        level = violation.level,
    }
```

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -0,0 +1,67 @@
...
+
+    if violation.name == "body-leading-blank" then
+        diagnostic = vim.tbl_extend("force", diagnostic, { line = 2 })
```
jose-elias-alvarez, line 16:
```suggestion
        diagnostic.line = 2
```

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -0,0 +1,67 @@
...
+    elseif vim.startswith(violation.name, "body") then
+        diagnostic = vim.tbl_extend("force", diagnostic, { line = 3 })
```
jose-elias-alvarez, line 18:
```suggestion
        diagnostic.line = 3
```

jose-elias-alvarez (commented):
Thanks for the PR! I put in a few suggestions, but this is working well. I also ran into a couple of (external) issues testing this out locally that I think we should consider mentioning in the documentation:

- I installed the CLI using [these instructions](https://commitlint.js.org/#/?id=getting-started), but I ran into [this issue](https://github.com/conventional-changelog/commitlint/issues/613). I'm not sure what your local setup is like, but I think we can at least mention that users may run into issues with global modules.
- It looks like this implementation depends on [commitlint-format-json](https://www.npmjs.com/package/commitlint-format-json), right? If so, we should mention the dependency.

lua/null-ls/builtins/diagnostics/commitlint.lua
wassimk: Thanks! I got these changes made. This is how I originally wrote it, but Sumneko gave me this warning, which I still don't understand.

image

lua/null-ls/builtins/diagnostics/commitlint.lua
wassimk: I initially switched it to `json_raw` because when I was typing a commit, then backspaced it to an empty commit, it would throw a JSON parsing error. But I can't even reproduce that now, so likely I was doing something else wrong. It does make more sense to let the errors bubble up. I got this changed! I appreciate your help.

wassimk:
@weilbith, that sounds interesting, but I don't see that it can do that. Can you point me to what you're seeing?

@jose-elias-alvarez, I made new commits addressing your suggestions. So this is ready to look at again.

weilbith:
@wassimk I'm very sorry. I saw [this](https://github.com/conventional-changelog/commitlint/tree/master/@commitlint/format) and that there is `commitlint --format`, but that is just the format of the linting output. 🤕

kaiuri:
Commitlint can also read the current buffer with the `-e` option, provided with a config file `without` a `"parserPreset"` option, thou the output isn't `.json` we could parse it. Should we provide such a feature?

PR #1009, opened 2022-08-14
fix(rubocop): misc fixes and improvements to parsing offenses

wassimk:
We started using null-ls for RuboCoc diagnostics and ran into some behaviors that we feel are worth improving here. There are two improvements to the diagnostics parsing and a change to severity assignments to better align with similar tools:

**1) don't underline whole files/methods**

Some RuboCop offenses, such as Metrics/ClassLength, return an offense start and end line/column range of the entire class. This means if you have Neovim underline diagnostics enabled, the whole class will get underlined, which is a lot of noise and unhelpful.

I looked up how Solargraph, the most popular Ruby LSP which uses RuboCop, [handles this situation](https://github.com/castwide/solargraph/blob/81bd810e7578040cbdb7650b71abb5c8a10962ab/lib/solargraph/diagnostics/rubocop.rb#L87-L95). If the end line is different from the start line, it says the offense ends on the start line at column 0. This has the effect of underlining just the class declaration instead of the whole thing.

**Before: noisy underlines with convention as warning**
image

**After: no more noisy underline and conversion to information (change 3)**
image

**2) honor rubocop configured excludes**

You can configure RuboCop to exclude files from diagnostics. However, using the RuboCop command line on those files will not honor those excludes without the [`--force-excludes`](https://docs.rubocop.org/rubocop/configuration.html) command line argument.

We need this to prevent null-ls, through RuboCop, from returning diagnostics for files we asked RuboCop to ignore.

**3) tweak severity assignments**

I started using RuboCop through null-ls on some personal projects and noticed I was only getting one type of diagnostic severity. At first, I thought the null-ls built-in for RuboCop had a problem. But on further inspection, it was more about default severity levels and what I'm used to on work projects, where we use Solargraph.

This change [matches what Solargraph](https://github.com/castwide/solargraph/blob/81bd810e7578040cbdb7650b71abb5c8a10962ab/lib/solargraph/diagnostics/rubocop.rb#L13-L19), the most popular Ruby LSP that provides diagnostics through RuboCop, does.


jose-elias-alvarez:
Thank you for the PR! All of the changes sound reasonable, and I think matching Solargraph is a good heuristic.

PR #1013, opened 2022-08-15
feat(builtin): Add `textlint` formatting

uhooi:
I implemented textlint formatting with reference to #532 and [markdownlint](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/8c90ccf7ffbdeb49da415837ec45e6ac457d5c60/lua/null-ls/builtins/formatting/markdownlint.lua).


jose-elias-alvarez:
LGTM, thank you!

PR #1014, opened 2022-08-15
feat: add markdown-toc

danilshvalov:


jose-elias-alvarez:
Thanks for the PR! I'm having a little trouble getting this to work locally, though I think I might be misunderstanding what the tool does. When I run the formatter via null-ls or run it directly using `markdown-toc -i $FILENAME`, it doesn't seem to do anything. I see that the documentation says that the `-i` flag should do this:

> Edit the <input> file directly, injecting the TOC at ;

I think the tool is correctly parsing files, though, since when I make these changes, it replaces the entire file with a table of contents:

```diff
diff --git a/lua/null-ls/builtins/formatting/markdown_toc.lua b/lua/null-ls/builtins/formatting/markdown_toc.lua
index b2a9c1b..9160335 100644
--- a/lua/null-ls/builtins/formatting/markdown_toc.lua
+++ b/lua/null-ls/builtins/formatting/markdown_toc.lua
@@ -13,8 +13,8 @@ return h.make_builtin({
     filetypes = { "markdown" },
     generator_opts = {
         command = "markdown-toc",
-        args = { "-i", "$FILENAME" },
-        to_temp_file = true,
+        args = { "-" },
+        to_stdin = true,
     },
     factory = h.formatter_factory,
 })
```

Am I misunderstanding what this is supposed to do?

danilshvalov:
Have you added `<!-- toc -->`?

danilshvalov:
Example:

**Before**:
```markdown
<!-- toc -->

# A
# B
# C
```

**After**:

```markdown
<!-- toc -->

- [A](#a)
- [B](#b)
- [C](#c)

<!-- tocstop -->

# A
# B
# C
```

jose-elias-alvarez:
> Have you added `<!-- toc -->`?

Nope, I hadn't, but after doing so it works well. That wasn't clear to me from the CLI documentation, so maybe we should add a mention of that in our documentation here, so that users don't conclude that it's broken.

Also, it seems to add 2 consecutive blank lines to the top of the file, which conflicts with `markdownlint`. Is that configurable, or is it something users will have to live with?

danilshvalov:
It doesn't seem like a big problem. I think usually people add a header for ToC. For example:

```markdown
# Table of Contents

<!-- toc -->

- [A](#a)
- [B](#b)
- [C](#c)

<!-- tocstop -->

# A

# B

# C
```

And it doesn't conflict with `markdownlint` as I understand (I use both of them and here is the result).

jose-elias-alvarez:
Ah I see, the behavior probably assumes that you have a header, which makes sense. 

Let me know if you think it's sensible to mention that the file needs to contain the line `<!-- toc -->` to our documentation. Otherwise I think this is good to go.

danilshvalov:
Is the note clear enough?

jose-elias-alvarez:
Perfect, thanks!

PR #1015, opened 2022-08-15
Use stdin for npm-groovy-lint and add filetype

mosheavni:
Following couple of issues I created for npm-groovy-lint to accept stdout and single filenames:
https://github.com/nvuillam/npm-groovy-lint/pull/236
https://github.com/nvuillam/npm-groovy-lint/issues/227
npm-groovy-lint now works better with stdin and all.
Also, `Jenkinsfile` is a valid filetype for groovy formatting, this should be enabled by default.
Thanks!


jose-elias-alvarez:
Great, thanks for putting this in!

PR #1017, opened 2022-08-16
fix(utils): resolve symlinks before testing executable bit

jceb:


jose-elias-alvarez:
Nice catch, thank you!

PR #1021, opened 2022-08-18
feat: buf format via stdin

choplin:
[buf](https://buf.build/) now support format files via stdin.

This would be more efficient than the current temp_file settings. (Additionally, temp_file settings don't work out-of-the-box in my environment)

### DIFF
--- a/lua/null-ls/builtins/formatting/buf.lua
+++ b/lua/null-ls/builtins/formatting/buf.lua
@@ -10,14 +10,12 @@ return h.make_builtin({
         description = "A new way of working with Protocol Buffers.",
     },
     method = FORMATTING,
-    to_temp_file = true,
-    from_temp_file = true,
+    to_stdin = true,
     filetypes = { "proto" },
     generator_opts = {
         command = "buf",
         args = {
             "format",
-            "-w",
             "$FILENAME",
         },
     },


stefanvanburen:
I don't think it's supported quite yet - there's an [open PR](https://github.com/bufbuild/buf/pull/1345) for it though. I'd assume it'll make it out in the next release.

jose-elias-alvarez:
Thanks! Let's wait until the upstream PR has been merged, then we can go ahead and merge this, since I agree it's a big improvement.

lua/null-ls/builtins/formatting/buf.lua
```diff
@@ -10,14 +10,12 @@ return h.make_builtin({
...
     },
     method = FORMATTING,
-    to_temp_file = true,
-    from_temp_file = true,
+    to_stdin = true,
     filetypes = { "proto" },
     generator_opts = {
         command = "buf",
         args = {
             "format",
```
amckinney, line 18: Depending on whether https://github.com/bufbuild/buf/pull/1345 gets merged in as-is, you'll need to replace the `-w` flag with `--stdin` so that the `buf format` actually reads the input.

choplin:
Thanks, everyone! Right, I misunderstood the buf behavior. Although this PR worked for me as expected, it didn't format input from stdin but format all proto files on the project defined with `buf.yaml` and `buf.work.yaml`.  Since the rest proto files were already formatted, it looked to format only the file being edited.

By the way, I added the `--stdin` flag for https://github.com/bufbuild/buf/pull/1345. I'd be happy if this would be merged when https://github.com/bufbuild/buf/pull/1345 is merged and released.

lua/null-ls/builtins/formatting/buf.lua
choplin: Right. We need that flag to support stdin after that PR is merged.

choplin:
I'll watch https://github.com/bufbuild/buf/pull/1345, and let you know when it's released.

Issue #686, opened 2022-02-14
Diagnostic source for TODO, FIXME, etc.


andrewferrier:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

One of the things I have which I've implemented locally is a null-ls diagnostic source which highlights `TODO` and `FIXME` in code as diagnostic hints. I find this useful for finding things I need to fix.

I was thinking about making this a generic built-in a bit like the other pull req I have going for trailing whitespace (https://github.com/jose-elias-alvarez/null-ls.nvim/pull/665). Do you think that would be useful/interesting?

One of the things that makes it slightly more complicated is what terms to highlight. There appears to be no standard that covers this so I would propose a good initial list is that mentioned in [this](https://stackoverflow.com/a/7284583/27641) SO link. Or perhaps it should be configurable by passing in a list of terms (maybe with a conservative default list?).

Also again some LSPs/sources already highlight these, so like the trailing space one, the user would probably need to figure out what they need to add to `disabled_filetypes`.

### Help

Yes

### Implementation help

I think at this stage just approval it would be worth merging is sufficient.

jose-elias-alvarez:
This would overlap a bit with [todo-comments.nvim](https://github.com/folke/todo-comments.nvim) but I'd still be up for including it here. Per-source configuration is still a work in progress, so I think going with the list in the link for now sounds good.

andrewferrier-ibm:
Hmm, well I wasn't aware of `todo-comments`. I've tried it out and actually I have to say I don't really like it personally. The main reason is that it actually doesn't seem to integrate with the diagnostics ecosystem in NeoVim. If memory serves, that was only factored out from LSP in the 0.6 timeframe and so I guess maybe that's why, but it integrates with a bunch of things that personally I don't use (Telescope/Trouble), but won't actually display found items as an NeoVim diagnostic.

So I would argue that there isn't actually that much overlap - I'm solving the same problem, but in quite a different way architecturally. 

So if you're up for it, I'll work on this. I would like to make the list of search terms configurable so it might take a little longer than before.

jose-elias-alvarez:
Closing this as #1028 adds support. Per-source configuration is still not possible with our current architecture, but I think the implementation there has a good amount of flexibility for additional patterns.

andrewferrier:
[...]

PR #1028, opened 2022-08-20
feat: add todo-comments diagnostics

danilshvalov:
Refers to #686.

I've implemented diagnostics, but I have a question.  Is there a way to get keywords by configuration?  Or can it only be hard-coded?


danilshvalov:
It depends on `nvim-treesitter` because it expects a treesitter parser

jose-elias-alvarez:
Got it, thanks! This is working well for me. An interesting future improvement might be to use the [`multiple_files`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#multiple_files) option to scan a project, kind of like [todo-comments.nvim](https://github.com/folke/todo-comments.nvim).

PR #1029, opened 2022-08-21
feat(builtins): add support for spectral lint

luizcorreia:
# Spectral

> A flexible JSON/YAML linter for creating automated style guides, with baked in support for OpenAPI v3.1, v3.0, and v2.0.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/spectral.lua
+++ b/lua/null-ls/builtins/diagnostics/spectral.lua
@@ -13,14 +13,14 @@ return h.make_builtin({
     method = DIAGNOSTICS,
     filetypes = { "yaml", "json" },
     generator_opts = {
-        command = 'spectral',
+        command = "spectral",
         args = {
             "lint",
             "--stdin-filepath",
             "-f",
             "json",
         },
-        format = 'json',
+        format = "json",
         to_stdin = true,
         ignore_stderr = true,
         use_cache = true,
@@ -44,5 +44,5 @@ return h.make_builtin({
             end
             return diags
         end,
-    }
+    },
 })
--- a/lua/null-ls/builtins/diagnostics/spectral.lua
+++ b/lua/null-ls/builtins/diagnostics/spectral.lua
@@ -1,0 +1,48 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+local severities = { 1, 2, 3, 4 }
+
+return h.make_builtin({
+    name = "spectral",
+    meta = {
+        url = "https://github.com/stoplightio/spectral",
+        description = "A flexible JSON/YAML linter for creating automated style guides, with baked in support for OpenAPI v3.1, v3.0, and v2.0.",
+    },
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/spectral.lua
```diff
@@ -0,0 +1,48 @@
...
+        url = "https://github.com/stoplightio/spectral",
+        description = "A flexible JSON/YAML linter for creating automated style guides, with baked in support for OpenAPI v3.1, v3.0, and v2.0.",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "yaml", "json" },
+    generator_opts = {
+        command = "spectral",
+        args = {
+            "lint",
+            "--stdin-filepath",
```
jose-elias-alvarez, line 19: According to the [pull request](https://github.com/stoplightio/spectral/pull/1712) that added support for this, it looks like we should add the filepath after this flag, which we can do using `$FILENAME`.

lua/null-ls/builtins/diagnostics/spectral.lua
```diff
@@ -0,0 +1,48 @@
...
+            "-f",
+            "json",
+        },
+        format = "json",
+        to_stdin = true,
+        ignore_stderr = true,
+        use_cache = true,
```
jose-elias-alvarez, line 26: This won't do anything here - it would only work if we had another source using the same output.

jose-elias-alvarez (commented):
Thanks for the PR! I won't have a chance to test this for a bit but I left a couple of comments for now.

lua/null-ls/builtins/diagnostics/spectral.lua
luizcorreia: Hi @jose-elias-alvarez I made some tests, the --stdin-filepath with $FILENAME don't work. 
but is possible using that way:
```
    args = {
      'lint',
      '$FILENAME',
      '-f',
      'json',
    },
```
What is the recommended method?

lua/null-ls/builtins/diagnostics/spectral.lua
luizcorreia: Without 
```
format = "json",
to_stdin = true,
```
I received a error.

It's possible don't use to_stdin using the solution with $FILENAME.

jose-elias-alvarez: GitHub makes this unclear, but the option I meant to highlight was `use_cache = true`, which is not doing anything here.

lua/null-ls/builtins/diagnostics/spectral.lua
jose-elias-alvarez: I tried this locally, and this seems to work:

```sh
cat file.yaml | spectral lint --stdin-filepath file.yaml -f json
```

Does that not work on your end? We should be able to replicate the same structure like so:

```lua
    args = {
      'lint',
      '--stdin-filepath',
      '$FILENAME',
      '-f',
      'json',
    },
```

jose-elias-alvarez:
Looks good, thanks again!

PR #1033, opened 2022-08-22
Add support for ts-standard

jamietanna:


lua/null-ls/builtins/formatting/tsstandard.lua
```diff
@@ -0,0 +1,22 @@
...
+        url = "https://github.com/standard/ts-standard",
+        description = "Typescript style guide, linter, and formatter using StandardJS",
+    },
+    method = FORMATTING,
+    filetypes = { "typescript", "typescriptreact" },
+    generator_opts = {
+        command = "ts-standard",
+        args = { "--stdin", "--fix" },
+        to_stdin = true,
+        dynamic_command = cmd_resolver.from_node_modules,
```
jamietanna, line 19: Not sure this is working, in the repo I'm testing with, I'm using `yarn` and it doesn't seem to be running 🤔

Issue #868, opened 2022-05-15
How can I enable the clang-check linter?


jose-elias-alvarez:
### Discussed in https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/867

<div type='discussions-op-text'>

<sup>Originally posted by **krishnakumarg1984** May 15, 2022</sup>
clangcheck (available as the executable name `clang-check`) is a helpful linter that better understands the semantics of the C++ language. How can we enable this linter with `null-ls`? </div>

Converted this to an issue for visibility. To add this as a built-in source, we could try porting [the logic from ALE](https://github.com/dense-analysis/ale/blob/master/ale_linters/cpp/clangcheck.vim), though it seems like we'd have to do some extra work to find and use a compilation database if it's available.

PR #1034, opened 2022-08-23
feat(diagnostics): add clang-check

danilshvalov:
Refers to #868

### Example

Usage example

### How to reproduce it?

1. Put these files in the same directory:
```cpp
// main.cpp
#include <iostream>

int main() {
  int *i;
  int j = *i;

  std::cout << j << std::endl;

  return 0;
}
```
```cmake
# CMakeLists.txt
add_executable(main main.cpp)
```
2. Run in the terminal (in the same directory):
```shell
cmake -S . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
```
3. Open neovim:
```shell
nvim main.cpp
```


lua/null-ls/builtins/diagnostics/clang_check.lua
```diff
@@ -0,0 +1,41 @@
...
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "clang_check",
+    meta = {
+        url = "https://releases.llvm.org/14.0.0/tools/clang/docs/ClangTools.html",
+        description = "ClangCheck combines the LibTooling framework for running a Clang tool with the basic Clang diagnostics by syntax checking specific files in a fast, command line interface.",
+        notes = {
+            "`clang-check` will be run only when files are saved to disk, so that `compile_commands.json` files can be used. It is recommended to use this linter in combination with `compile_commands.json` files.",
```
jose-elias-alvarez, line 12: If this is the case, should we instead use `DIAGNOSTICS_ON_SAVE` as this source's method? Otherwise it's easy for diagnostics to get out of sync during normal editing.

danilshvalov:
Done

jose-elias-alvarez:
Made a similar change here to replace `from_patterns` with `from_pattern`. Thanks again!

PR #1049, opened 2022-08-26
Enforce absolute paths in phpcs output 

greg0ire:
When using a configuration file to configure the basepath, like for instance here: https://github.com/doctrine/collections/blob/c2a8c5f0f806cc467b1103a418db3e73815e83b6/phpcs.xml.dist#L3, no diagnostic appears.
Overriding it again in the CLI allows to get absolute paths, and fixes the issue.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/phpcs.lua
+++ b/lua/null-ls/builtins/diagnostics/phpcs.lua
@@ -28,7 +28,8 @@ return h.make_builtin({
             "1",
             -- process stdin
             "--stdin-path=$FILENAME",
-            "-",
+            -- get absolute paths under params.output.files
+            "--basepath=",
         },
         format = "json_raw",
         to_stdin = true,


greg0ire:
@weeman1337 can you please review this and maybe test that it does not break your setup?

weeman1337:
Tested @greg0ire . With your branch it doesn't map the files correctly any more → I don't see the diagnostics in the editor.

greg0ire:
Thanks! I think what this means is that `phpcs` is returning a report with absolute paths to you.

Do you use `phpcs` as a global tool, or through `vendor/bin/phpcs`, like this:

```
        null_ls.builtins.diagnostics.phpcs.with({
          prefer_local = 'vendor/bin'
        }),
```

?

greg0ire:
@weeman1337 I found a far better fix, can you please test again :pray: ?

lua/null-ls/builtins/diagnostics/phpcs.lua
```diff
@@ -28,7 +28,8 @@ return h.make_builtin({
             "1",
             -- process stdin
             "--stdin-path=$FILENAME",
-            "-",
+            -- get absolute paths under params.output.files
+            "--basepath=",
```
greg0ire, line 32: I've tried using
```suggestion
            '--basepath=""',
```

but that seems to result in the tool looking for a directory literally named `""`

weeman1337:
@greg0ire you can find my setup here: https://pubcode.weimann.digital/projects/11/files/ea7982c05e4caccdeb0810adfb091cc8d18e9eaf/roles/neovim/files/config/lua/weeman/plugins.lua?position=source-336.1-395.6-1

Tested again, works now :+1:

weeman1337:
Still works :+1:

greg0ire:
Thank you :tada:

jose-elias-alvarez:
Great, thanks all!

Issue #508, opened 2021-12-28
Please provide documentation to configure generic (non-builtin) linters and formatters


krishnakumarg1984:
First of all, thanks a lot for `null-ls`. This is indeed such an awesome plugin that I am migrating my previous linting and formatting setup from ALE during this winter break.

The built-in diagnostics and formatting for C++ works well. However, I'd like to know how to configure a few other useful linters and formatters such as:

- [flawfinder](https://github.com/david-a-wheeler/flawfinder)
- [cpplint](https://github.com/google/styleguide/tree/gh-pages/cpplint)
- [clang-tidy](https://clang.llvm.org/extra/clang-tidy/)

 for C++ filetype. A wiki entry in the project's documentation to configure generic linters and formatters shall be much appreciated.

jose-elias-alvarez:
The documentation is a work in progress (see #364). At the moment, the best way to add a new source is to look at existing built-ins and modify them as necessary. #69 contains an implementation for clang-tidy that may work as a starting point.

krishnakumarg1984:
Okay. Thank you.  It would be good to have `clang-tidy` to be available in the official supported list of linters at some point in the future

jose-elias-alvarez:
This is essentially a duplicate of #364, so I'm closing it in favor of that issue. I'm still open to PRs to improve the documentation situation, though I'd ideally like to make some changes to the core null-ls source API before doing it myself.

PR #1051, opened 2022-08-26
feat(diagnostics): add cpplint

VidocqH:
[cpplint](https://github.com/cpplint/cpplint) support, which mentioned at https://github.com/jose-elias-alvarez/null-ls.nvim/issues/508

Tested with `cpplint` that installed by `brew` and [`Mason`](https://github.com/williamboman/mason.nvim)

Preview:
image

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cpplint.lua
+++ b/lua/null-ls/builtins/diagnostics/cpplint.lua
@@ -1,0 +1,37 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "cpplint",
+    meta = {
+        url = "https://github.com/cpplint/cpplint",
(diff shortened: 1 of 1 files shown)


lvim-tech:
LGTM

lua/null-ls/builtins/diagnostics/cpplint.lua
```diff
@@ -0,0 +1,37 @@
...
+            severities = {
+                build = h.diagnostics.severities["warning"],
+                whitespace = h.diagnostics.severities["hint"],
+                runtime = h.diagnostics.severities["warning"],
+                legal = h.diagnostics.severities["information"],
+                readability = h.diagnostics.severities["information"]
+            },
+        }),
+        check_exit_code = function(code)
+            return code >= 1
```
jose-elias-alvarez, line 33: Just to check but is this okay? This means that a `0` exit code will be handled as an error.

lvim-tech: ```cppcheck``` has the same setting.

jose-elias-alvarez:
Thank you for the contribution!

PR #1056, opened 2022-08-31
feat(doc):add index in doc/BUILTINS.md

lunarxlark:
BUILTINS.md is so large. So i add index to search easily nullls.buitinins.

```md
<!-- THIS FILE IS AUTOMATICALLY GENERATED. DO NOT EDIT IT MANUALLY -->

# Built-in Sources

This is an automatically generated list of all null-ls built-in sources.

See [BUILTIN_CONFIG](BUILTIN_CONFIG.md) to learn how to set up and configure these sources.

## Index
- [Code Actions](#code-actions)
- [Completion](#completion)
- [Diagnostics](#diagnostics)
- [Formatting](#formatting)
- [Hover](#hover)

## Code Actions
...
```

### DIFF
--- a/scripts/autogen.lua
+++ b/scripts/autogen.lua
@@ -61,6 +61,19 @@ end
 local markdown_content = {}
 
 -- helpers
+local generate_method_index = function()
+    local index = {}
+    for _, method in ipairs(methods) do
+        local iname = method
+        local link = method:gsub("_", "-")
+        for _, component in ipairs(vim.split(method, "_")) do
+            iname = component:sub(1, 1):upper() .. component:sub(2)
+        end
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thank you! Could we update this so that it handles methods with multiple words in their name correctly? For example, I'd prefer `Code Actions` over `Actions`. 

Now that I look at it, we might even be able to reuse some of the logic from `generate_method_header` to avoid duplication.

lunarxlark:
@jose-elias-alvarez thanks for reply. I fix generate_method_index to rename method `Actions` -> `Code Actions`.
I was able to achieve this using same function in `generate_method_header `.

jose-elias-alvarez:
Excellent, thank you!

PR #1060, opened 2022-09-02
feat(diagnostics): add support for pycodestyle

bydlw98:
Add support for [pycodestyle](https://github.com/PyCQA/pycodestyle) which is is a tool to check your Python code against some of the style conventions in [PEP 8](http://www.python.org/dev/peps/pep-0008/).


jose-elias-alvarez:
LGTM, thank you!

PR #1066, opened 2022-09-04
feat(hover): add support for printenv

bydlw98:
Shows the value for the current environment variable under the cursor.

This source is similar in function to `printenv` where it shows value of environment variable, however this source uses `vim.fn.environ()` instead of `printenv` thus making it cross-platform.

### DIFF
--- a/lua/null-ls/builtins/hover/printenv.lua
+++ b/lua/null-ls/builtins/hover/printenv.lua
@@ -12,16 +12,14 @@ return h.make_builtin({
             -- Get word under cursor
             local cword = vim.fn.expand("<cword>")
 
-            -- Gets table of environment variables
-            local env_vars = vim.fn.environ()
-
-            -- Checks if cword is in table of environment variables
-            -- If not in table of environment variables, show in hover window "Error! `cword` is not an environment variable!"
-            -- Else show in hover window value of environment variable
-            if env_vars[cword] == nil then
-                done({ "Error! " .. cword .. " is not an environment variable!" })
+            -- Checks if cword is an environment variable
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/hover/printenv.lua
```diff
@@ -0,0 +1,35 @@
...
+    name = "printenv",
+    method = HOVER,
+    filetypes = { "sh", "dosbatch", "ps1" },
+    generator = {
+        fn = function(_, done)
+            -- Get word under cursor
+            local cword = vim.fn.expand("<cword>")
+
+            -- Gets table of environment variables
+            local env_vars = vim.fn.environ()
```
jose-elias-alvarez, line 16: I don't think this will realistically make a difference, but any thoughts on using the libuv equivalents? [`os_environ`](https://github.com/luvit/luv/blob/master/docs.md#uvos_environ) should be a drop-in replacement, and [`os_getenv`](https://github.com/luvit/luv/blob/master/docs.md#uvos_getenvname-size) might be even better.

bydlw98:
Hi. I have made the changes to use [os_get_env](https://github.com/luvit/luv/blob/master/docs.md#uvos_getenvname-size) instead of `vim.fn.environ()`. Also I would just like to ask on why using libuv equivalents will be better? Is the libuv equivalents async and faster than `vim.fn.environ()`?

jose-elias-alvarez:
Theoretically it'll be faster, but realistically it won't make a difference. I do think `os_getenv` is a better fit for this case, since we're only interested in the value of one variable. Thanks for the contribution!

PR #1067, opened 2022-09-04
feat(diagnostics): add cmake_lint to go alongside cmake_format

MitchAllain:
Add support for `cmake_lint` diagnostics support from the `cmakelang` Python, part of which is already supported by the `cmake_format` builtin.

`cmake_lint` does not provide an `end_col`, so this currently goes to EOL, making underlines a little messy.


lua/null-ls/builtins/diagnostics/cmake_lint.lua
```diff
@@ -0,0 +1,38 @@
...
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "cmake" },
+    generator_opts = {
+        command = "cmake-lint",
+        args = {
+            "$FILENAME",
+        },
+        format = "line",
+        to_stdin = false,
```
jose-elias-alvarez, line 20: So as currently written, the linter will read the file on disk, which means that diagnostics won't update properly until the file is actually written. It does seem that `cmake-lint` has `stdin` support, but [this PR](https://github.com/cheshirekow/cmake_format/pull/285) seems to imply that it doesn't currently work.

Using `stdin` would be ideal, but other options that we have are [writing the buffer's contents to a temp file](https://github.com/cheshirekow/cmake_format/pull/285) or [only generating diagnostics on save](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/10d621fd6efbcdead63ba570871223e255e0eeb5/lua/null-ls/methods.lua#L23).

MitchAllain: I see, thanks for the feedback. Personal preference would be to switch to `DIAGNOSTICS_ON_SAVE`. Judging by the age of that PR in `cmake_format`, stdin may not be fixed soon.

jose-elias-alvarez:
Thanks!

PR #1069, opened 2022-09-05
fix(builtins): allow use of local .sqlfluff file

andrewferrier:
The previous behaviour had null-ls pass in the contents of a SQL file to stdin. This doesn't work if the user has created a local .sqlfluff file (for example to specify the dialect), as sqlfluff cannot tell where the file is on the filesystem. This commit changes that behaviour.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1455,11 +1455,11 @@ local sources = {
 - Filetypes: `{ "sql" }`
 - Method: `diagnostics`
 - Command: `sqlfluff`
-- Args: `{ "lint", "-f", "github-annotation", "-n", "--disable_progress_bar", "-" }`
+- Args: `{ "lint", "-f", "github-annotation", "-n", "--disable_progress_bar" }`
 
 #### Notes
 
-- SQLFluff needs a mandatory `--dialect` argument. Use `extra_args` to add yours. `extra_args` can also be a function to build more sophisticated logic.
+- SQLFluff needs a mandatory `--dialect` argument. Use `extra_args` to add yours, or create a .sqlfluff file in the same directory as the SQL file to specify the dialect (see the sqlfluff docs for details). `extra_args` can also be a function to build more sophisticated logic.
 
 ### [standardjs](https://standardjs.com/)
 
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1455,7 +1455,7 @@ local sources = {
 - Filetypes: `{ "sql" }`
 - Method: `diagnostics`
 - Command: `sqlfluff`
-- Args: `{ "lint", "-f", "github-annotation", "-n", "--disable_progress_bar" }`
+- Args: `{ "lint", "-f", "github-annotation", "-n", "--disable_progress_bar", "$FILENAME" }`
 
 #### Notes
 
--- a/lua/null-ls/builtins/diagnostics/sqlfluff.lua
+++ b/lua/null-ls/builtins/diagnostics/sqlfluff.lua
@@ -26,9 +26,11 @@ local sources = {
             "github-annotation",
             "-n",
             "--disable_progress_bar",
+            "$FILENAME"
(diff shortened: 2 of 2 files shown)


jose-elias-alvarez:
Hmm, the issue here is that we will lose the ability to lint the contents of the buffer vs. linting the file on disk, which is a worse experience for users. In other cases we'd be able to use [to_temp_file](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#to_temp_file), but I think we'll run into the same issue here, since we won't be able to find the config file from the temporary directory.

I found [this issue](https://github.com/sqlfluff/sqlfluff/issues/3640) where the conclusion seems to be that specifying a path to a config file is the right solution for this case. That's possible to wire together with null-ls, though it's a little less user-friendly. What do you think?

andrewferrier:
OK. Thanks for your feedback, great points. So you definitely can specify the config file using `--config`, and that works for `sqlfluff` itself if run from the command line. The problem, though, is that the intention here is that the config file is local to (i.e. in the directory of, or a parent directory of) the SQL file being checked. For `sqlfluff`, this is important as different SQL files can have different dialects (e.g Postgres/MySQL/etc.). On my system I have both SQLite and Postgres files, for example. So specifying `--config` in, say, `extra_args` isn't really going to work. Basically, I'm then "forced" to have the same config for all files on my system I might want to check.

I guess there's a more general problem here - as you say, the filename approach doesn't work for checking the buffer in NeoVim until it's saved, but the stdin approach doesn't find local config. I suspect it's not only `sqlfluff` which ultimately might be affected by this.

Suggestion (not specific to sqlfluff) - what do you think of the idea of `to_temp_file` at least creating the temp file in the same directory as the file in the buffer? This would mean for any tool like `sqlfluff` which uses config files in the same dir (I think many linter style tools look through the dir and parent dirs for config like this), they would still find it. Thoughts?

jose-elias-alvarez:
Sure, I think it's worth considering. It's come up in the past, and I was a bit resistant, since I think it could cause bad interactions with other processes that are watching the file system, but I think cases like this are ultimately unavoidable, so I'd be open to giving it a shot.

[Our current implementation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/4b403d2d724f48150ded41189ae4866492a8158b/lua/null-ls/loop.lua#L234) has quite a bit of complexity that would be solved by just creating a file in the same directory as the original file, so it might even make a good default. It looks like [formatter.nvim](https://github.com/mhartington/formatter.nvim/blob/465fd2fa4f6c32753d2fcae3de39858055d22b7e/lua/formatter/tempfile.lua#L29) already defaults to using the buffer's directory, so it shouldn't be a big deal to change how we do it, too.

andrewferrier:
So the honest answer is I'm probably not going to have too much time to look at this over the coming week, and even if I did, I'd probably need quite a lot of guidance! So if you have the time, it would be wonderful if you could work on the fix for this.

andrewferrier:
I see you've merged the temp file PR now. I think this PR should still be good to go unchanged now that's in.

jose-elias-alvarez:
I might be missing something, but I don't think the code is using `to_temp_file` yet? I also imagine we'd need to pass the filename in the arguments.

andrewferrier:
D'oh, I forgot about that :) You're not missing something! Let me fix and re-push.

andrewferrier:
OK, I've pushed those changes. I also did some testing and I think this is all working. If I change `dialect` in a `.sqlfluff` file next to the `.sql` file, it has effects on the linting, so it's clearly picking it up.

Hopeful that this new `to_temp_file` technique can prove useful with some other diagnostic sources in the same way...

jose-elias-alvarez:
> Hopeful that this new to_temp_file technique can prove useful with some other diagnostic sources in the same way...

I definitely think so, the issue has come up before. Thanks for the PR and for all your feedback!

PR #1070, opened 2022-09-05
Revert "fix(utils): resolve symlinks before testing executable bit"

jose-elias-alvarez:
Reverts jose-elias-alvarez/null-ls.nvim#1017. @jceb 

I believe this is the cause of the issues described in #1047, though I don't know exactly why. If I had to guess, I'd say that `vim.fn.resolve` and `vim.fn.exepath` have different behavior on different platforms, so one potential alternative  would be to use libuv's [fs_realpath](https://github.com/luvit/luv/blob/master/docs.md#uvfs_realpathpath-callback). Either way, I think it's safest to revert this for now.

### DIFF
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -85,7 +85,7 @@ end
 ---@param cmd string? command to check
 ---@return boolean, string is_executable, string|nil error_message
 M.is_executable = function(cmd)
-    if cmd and vim.fn.executable(vim.fn.resolve(vim.fn.exepath(cmd))) == 1 then
+    if cmd and vim.fn.executable(cmd) == 1 then
         return true
     end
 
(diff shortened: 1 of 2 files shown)


jceb:
Interesting, is this a neovim bug?

jose-elias-alvarez:
Hard to say, but from the linked discussion it looks like the return value of either `vim.fn.resolve` or `vim.fn.exepath` was somehow causing `vim.fn.executable` to return `1` for nonexistent files. If I had to guess, I'd say `resolve` is the culprit, since [we use `exepath` elsewhere](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/4b403d2d724f48150ded41189ae4866492a8158b/lua/null-ls/loop.lua#L147) without issues. I'd say `fs_realpath` is worth a shot if we can get someone who originally ran into this issue to test it out.

PR #1075, opened 2022-09-06
feat(loop): create temp files at source location

jose-elias-alvarez:
Discussed in #1069. @andrewferrier 

Our current temp file creation mechanism uses a dedicated temporary directory (`$TEMP` on Windows, `$XDG_RUNTIME_DIR` if set, and `/tmp` otherwise). The issue, which has come up before, is that some linters / formatters use file paths to resolve config files. Well-behaved CLIs like [`black`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/black.lua) use an additional flag to specify the "actual" filepath, even when using `stdin`, but we can't expect this behavior for all CLIs (using a temp file is itself a hack to improve behavior when `stdin` is not supported).

I resisted making this change in the past because I didn't like the idea of writing to an actual project, even temporarily, and I worried about potential bad interactions with file watchers. I've since learned that [formatter.nvim](https://github.com/mhartington/formatter.nvim/blob/master/lua/formatter/tempfile.lua) already does something similar, and I haven't seen any related issues there, so I think this should be okay. The change also removes a decent amount of complexity related to getting the right temp directory and adding a file extension to the generated path.

I think this is worth trying out and potentially reverting / making it opt-in if we get reports of issues, but I'm happy to hear thoughts from anyone who happens to see this.

### DIFF
--- a/lua/null-ls/helpers/generator_factory.lua
+++ b/lua/null-ls/helpers/generator_factory.lua
@@ -18,7 +18,7 @@ local get_content = function(params)
     end
 
     -- otherwise, get content directly
-    return u.buf.content(params.bufnr, true)
+    return u.buf.content(params.bufnr, true) --[[@as string]]
 end
 
 local parse_args = function(args, params)
@@ -306,23 +306,13 @@ return function(opts)
             }
 
             if to_temp_file then
-                local filename = vim.fn.fnamemodify(params.bufname, ":e")
-                local temp_path, cleanup = loop.temp_file(get_content(params), filename)
+                local content = get_content(params)
+                local temp_path, cleanup = loop.temp_file(content, params.bufname)
 
                 spawn_opts.on_stdout_end = function()
                     if from_temp_file then
-                        -- wrap to make sure temp file is always cleaned up
-                        local ok, err = pcall(function()
-                            local fd = vim.loop.fs_open(temp_path, "r", 438)
-                            local stat = vim.loop.fs_fstat(fd)
-                            params.output = vim.loop.fs_read(fd, stat.size, 0)
-                            vim.loop.fs_close(fd)
-                        end)
-                        if not ok then
-                            log:warn("failed to read from temp file: " .. err)
-                        end
+                        params.output = loop.read_file(temp_path)
                     end
-
                     cleanup()
                 end
                 params.temp_path = temp_path
--- a/lua/null-ls/loop.lua
+++ b/lua/null-ls/loop.lua
@@ -48,7 +48,8 @@ M.spawn = function(cmd, args, opts)
     local handler, input, check_exit_code, timeout, on_stdout_end, env =
         opts.handler, opts.input, opts.check_exit_code, opts.timeout, opts.on_stdout_end, opts.env
 
-    local output, error_output = "", ""
+    local output, error_output
+    output, error_output = "", ""
     local handle_stdout = function(err, chunk)
         if err then
             on_error("stdout", err)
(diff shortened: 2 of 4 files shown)


andrewferrier:
So I've tested this out (in combination with my pull request of #1069 specifically for sqlfluff) and it seems to work very well! The diagnostics now show up correctly.

I suppose you could consider making this opt-in on a per-diagnostic config basis (e.g. having a `temp_file_next_to_file = true` for `sqlfluff` - I'm sure there's a better name...) but right now I can't think even why that would be necessary, and it's just another option for you to maintain.

jose-elias-alvarez:
Glad to hear it's working! I'd definitely rather avoid adding another option, since we'd not only have to maintain it but also choose whether to use it each time we add a new source that uses temp files. I'm going to sit on this for another day or so and see if I can think of any pitfalls, but otherwise I think we can merge this and see how it goes.

andrewferrier:
> Glad to hear it's working! I'd definitely rather avoid adding another option, since we'd not only have to maintain it but also choose whether to use it each time we add a new source that uses temp files. I'm going to sit on this for another day or so and see if I can think of any pitfalls, but otherwise I think we can merge this and see how it goes.

Agreed, sounds good to me. Thanks...

vlada-dudr:
Hello, found this pull request after I recently updated null-ls and nvim-tree started to jump up and down as null-ls is creating files in the project hierarchy. It is bit annoying. I have some ideas:

andrewferrier:
Personally I think (1) is a good approach. Temp files are often considered "hidden" anyway

jose-elias-alvarez:
Agreed, if the issue is purely visual then I would rather go for a simpler fix before adding more configuration. I’ll put in the change later today.

vlada-dudr:
Yes, the issue is mostly visual. Also some of the temp files show in lazygit which I use, but that is also not a big deal. Both of them are solvable by adding a line to .gitignore, but I don't think is good to make people doing it. Feels too much like hack/workaround.

Thank you

soifou:
Personally I'm in favor of option 2.

I have neo-tree installed with `libuv_file_watcher` and `git_status` enabled. With this PR, PHPStan creates temp files like crazy whenever I go back to normal mode or doing normal operations. The dot prefix is not viable for me since I like to see how many files are hidden (now the count changes) and the git status of the working tree is updated in live, as you can see below., It is now a christmas tree :) 

https://user-images.githubusercontent.com/7377369/190828424-6e38ee57-46e5-4029-9bd9-9deb7de5321a.mp4

jose-elias-alvarez:
I see what you mean, but I'm not sure it's really a huge problem, to be honest. According to [this issue](https://github.com/nvim-neo-tree/neo-tree.nvim/issues/266), it seems that neo-tree has an option to ignore files matching a specific pattern, so assuming that works as expected, I think that should solve this case.

In the absence of larger, non-aesthetic issues, I am resistant to making this configurable because we are adding sources that depend on this feature. If we enable the option for some sources and not for others, then we'll get reports from users who assume it's a bug, and either way they'll either have to live with it or depend on their always-on file tree to provide some kind of filtering behavior.

soifou:
Just to be clear, what I meant for "option 2" was to let the user override the temp dir for a particular source, something like 

```lua
        b.diagnostics.phpstan.with({
                temp_dir = "/tmp",
        }),
```

So that's entirely up on him if something goes wrong. The default will still set the buffer dir. But I probably miss the point.

And thanks for looking at the possible options/issues of neo-tree. 
In fact in the demo, I already set `hide_by_pattern = { "_null-ls_*" }` but, as the name suggests, that only seems to hide not ignore. I guess I'll fill an issue on the repo then.

vlada-dudr:
Maybe configurability is not the best option. But making the files hidden is good idea, I think.

andrewferrier:
I'm still in favor of option 1. Option 2 would work somewhat if it were *per-formatter* (although I think allowing the user to configure it as you are showing @soifou isn't how I would see it being used), but the problem with that is that it doesn't solve the flickering issue for formatters that *do* have to use the current directory of the file, because if you apply it to them the issue with not finding configuration reoccurs (see the previous discussion for `sqlfluff`). I guarantee that will cause confusion.

It's not ideal, but I would compromise on option 1 - the dot will hide it for most people (you could even specify a 'known pattern' for the filename to make it easier for folks using file manager plugins like NeoTree to explicitly exclude, even if they show hidden files otherwise), and for those that don't you are trading visual unpleasantness for linter correctness.

Really, I think the proper fix for this is for tools like sqlfluff to allow a CLI option to specify where the file 'logically' is - then you can use stdin - but until they do that I think this is a good compromise.

vlada-dudr:
I completely agree

vlada-dudr:
I completely agree.

vlada-dudr:
I have another, but maybe ridiculously overkill solution. If available, null-ls could launch linter via bubblewrap, which can be used to overlay current file directly. So, something like `bwrap --bind / / --bind /tmp/tmpfile /path/to/original file /path/to/linter`. Also it won't work anywhere else then on Linux.

cseickel:
Hey guys, author of Neo-tree here. I'm personally not opposed to putting special handling on my file watchers to ignore events from these temp files. I already have special handling in .git directories so I can add one more filter.

Am I correct in my understanding that all of these temp files will start with `_null-ls_` or `._null-ls_` and will only be around for a few seconds?

jose-elias-alvarez:
@cseickel Your understanding is correct! As of #1114, null-ls will create temp files that start with `._null-ls_` that will persist for as long as the underlying command takes to run (usually less than a second, but it can be slower).

Rather than special-casing our temp files, though, I think we could maintain flexibility on both sides if neo-tree users had the ability to completely ignore files, not just treat them as hidden files.

cseickel:
> Rather than special-casing our temp files, though, I think we could maintain flexibility on both sides if neo-tree users had the ability to completely ignore files, not just treat them as hidden files.

My thinking on this is that it's actually a problem because when using the file watcher, it will refresh the tree every time these files are written or deleted, which can be a bigger problem than just the visual noise of hidden files. That's why I want to put in an exclusion in the "file changed" event itself. It's also an argument against creating these temp files within the project, because there are likely a lot of other plugins out there that utilize file watchers that will start to eat a lot of cpu cycles.

It would also be a good idea to add a `never_show_by_pattern` option, in a general sense, but for this specific case I think it's more important to skip all of the unnecessary "file changed" events.

jose-elias-alvarez:
> It's also an argument against creating these temp files within the project, because there are likely a lot of other plugins out there that utilize file watchers that will start to eat a lot of cpu cycles.

I agree that this is a possibility, and this is why it took me a while to agree to make this change (ultimately the fact that it enables sources that we otherwise would not be able to use won out). I am open to making this configurable if we can show that this is causing issues of this nature, especially performance issues. 

We are not the only plugin doing this (see [formatter.nvim](https://github.com/mhartington/formatter.nvim)) and I'm sure there are other programs in the wild that do the same, so I would imagine that giving users some amount of flexibility in determining what constitutes a change event is desirable, but I'm not a neo-tree user, so I can't say what's best for the plugin.

cseickel:
Yeah, I'm sure there's all sorts of software out there that creates temp files. I think it's fine to be the tree's responsibility to ignore junk files. 

Now that I've thought more about it, I'll add the `never_show_by_pattern` and then re-use that filter in the file changed event to prevent tree refreshes altogether if the cause was a "never show" file. That covers both sides of the issue. For anyone that stumbled on this conversation, I'll track that work in https://github.com/nvim-neo-tree/neo-tree.nvim/issues/527.

soifou:
@andrewferrier  I'm sure there are very good reasons behind this and honestly  I did not dug that much on the subject so I leave it up to you for the best solution.

I should report that another consequence of the change I observe here, I have now untracked files on my git tree when I exit **before** the source finishes its work. That's a edge case but it sometimes happen. 
A simple git clean would clear the mess, but that's add a chore that `/tmp` or `$XDG_RUNTIME_DIR` was doing on their own.

jose-elias-alvarez:
Yes, that’s a good point. I’ll make sure we handle that case correctly.

p5a0u9l:
This one change introduces some pain points into my work flow. I'm curious if other people are mentioning that? 

* The temporary files cause issues with nvim-tree.lua - fixable with a workaround
* They show up in git, even if temporarily - fixable with a workaround

I feel like an editor tool should leave a working directory completely untouched. I understand the reasoning used, but it seems unjustified imo.

jose-elias-alvarez:
You are not the only one to bring up the same issues (see: the comments here). As I said above, I believe that the benefits of the change outweigh the downsides that have been presented thus far, but I am open to changing my mind on this if there is evidence of more serious consequences.

Using temp files in the first place is a hack necessitated by tools that don't support `stdio`, and if you are truly bothered by this, I would recommend directing your attention upstream and seeing what can be done to improve support for editor integrations in the tools you use. 

I'm going to lock this issue because I don't believe it's productive to continue discussing the same topics. For anyone who finds this in the future: my conclusion is that flickering in file tree plugins / lazygit / other programs is not enough of a justification for me to revert this change. If you run into **non-visual** problems when using temp files, please feel free to open an issue.

jose-elias-alvarez:
#1223 raised a non-visual problem, and so #1224 makes this configurable. Please feel free to leave any feedback there.

PR #1077, opened 2022-09-06
feat(loop): check if command is executable by handling spawn failure

### LINKED ISSUE #912: `:checkhealth` is returning incorrect executable information when using `prefer_local`
### FAQ

### Issues

### Neovim Version

v0.8.0-dev

### Operating System

Linux

### Minimal config

```lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
	local path_sep = on_windows and "\\" or "/"
	local result = table.concat({ ... }, path_sep)
	return result
end

vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir
if on_windows then
	temp_dir = vim.loop.os_getenv("TEMP")
else
	temp_dir = "/tmp"
end

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
	-- add only what you need to reproduce your issue
	null_ls.setup({
		sources = {
			null_ls.builtins.diagnostics.eslint.with({
				prefer_local = "node_modules/.bin",
			}),
			null_ls.builtins.formatting.prettier.with({
				prefer_local = "node_modules/.bin",
			}),
		},
		debug = true,
	})
end

local function load_plugins()
	-- only add other plugins if they are necessary to reproduce the issue
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				"jose-elias-alvarez/null-ls.nvim",
				requires = { "nvim-lua/plenary.nvim" },
				config = null_ls_config,
			},
		},
		config = {
			package_root = package_root,
			compile_path = compile_path,
		},
	})
end

if vim.fn.isdirectory(install_path) == 0 then
	vim.fn.system({ "git", "clone", "https://github.com/wbthomason/packer.nvim", install_path })
	load_plugins()
	require("packer").sync()
else
	load_plugins()
	require("packer").sync()
end
```

### Steps to reproduce

- Open `nvim` within a JavaScript/TypeScript project. The underlying `null-ls` configuration should specify `prefer_local` for both `eslint `and `prettier`.
- Ensure that `prettier` and `eslint` are included in the project's dependencies (`node_modules`)
- Run `:checkhealth` and note that the health test shows that `eslint` and `prettier` are not executable
- Examine files within the project and note that `eslint` and `prettier` are in fact executable as lint diagnostics exist (via `eslint`) and formatting functionality (via `prettier`) works correctly.

### Expected behavior

`:checkhealth` should show that `eslint` and `prettier` are executable.

### Actual behavior

`:checkhealth` shows that `eslint` and `prettier` are not executable.

### Debug log

N/A

### Help

Yes

### Implementation help

_No response_

### Requirements

jose-elias-alvarez:
Supersedes #892. The purpose of that PR was to improve messaging in dynamically-resolved `node_modules` commands. This PR implements the approach I described in a comment:

> If you're up for it, a more robust solution might be to determine the ability to execute a command based on the result of vim.loop.spawn, since vim.fn.executable doesn't account for all cases. Neovim switched to this approach for LSP servers in https://github.com/neovim/neovim/pull/16430, and I've been meaning to switch over to it, too (except for healthchecks).

This should also improve the situation that #1017 was intended to fix, since null-ls doesn't have to check whether a command is executable before attempting to spawn it. @jceb Could you check if this works for you when using symlinks?

Unfortunately, this approach only works if we can actually spawn the command on success. We're not able to do so in health checks, so this PR also improves messaging and closes #912. Health checks are really for informational purposes, so as long as users are aware that the result of the checks doesn't affect how null-ls works, I think we're in the clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,15 @@ plugin before sending anything upstream.
    feature, open an issue and provide the information requested in the issue
    template.
 
+### My `:checkhealth` output is wrong! What do I do?
+
+Checking whether a given command is executable is tricky, and null-ls' health
+check doesn't handle all cases. null-ls' internal command resolution is
+independent of its health check output, which is for informational purposes.
+
+If you're not sure whether a given command is running as expected, enable debug
+mode and check your log (see below).
+
 ### How do I format files?
 
 null-ls formatters run when you call `vim.lsp.buf.formatting()` or
--- a/lua/null-ls/health.lua
+++ b/lua/null-ls/health.lua
@@ -11,8 +11,8 @@ local messages = {
 
 local function report(source)
     local name = source.name
-    local only_local = source.generator.opts and source.generator.opts.only_local
-    local command = source.generator.opts and source.generator.opts.command
+    local opts = source.generator.opts or {}
+    local command, only_local, prefer_local = opts.command, opts.only_local, opts.prefer_local
 
     if type(command) ~= "string" then
         health.report_info(string.format(messages["unable"], name, command))
@@ -24,7 +24,7 @@ local function report(source)
         return
     end
 
-    if only_local then
+    if only_local or prefer_local then
         health.report_warn(string.format(messages["executable-local"], name, command))
         return
     end
--- a/lua/null-ls/helpers/generator_factory.lua
+++ b/lua/null-ls/helpers/generator_factory.lua
@@ -174,14 +174,6 @@ return function(opts)
             opts.command = command
         end
 
-        if not dynamic_command then
-            local is_executable, err_msg = u.is_executable(command)
-            if not is_executable then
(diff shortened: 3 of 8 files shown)


jceb:
@jose-elias-alvarez the solution works well for me! Thank you for the fix.

jose-elias-alvarez:
Great, thanks for confirming!

Issue #1085, opened 2022-09-07
todo_comments built-in doesn't always work


heygarrett:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.8.0-dev-2499-g9d1d3a670-dirty Build type: Release LuaJIT 2.1.0-beta3 Compiled by garrett@Garretts-MacBook-Pro.local  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-9d1d3a6_1/share/nvim"  Run :checkhealth for more info

### Operating System

macOS 12.5.1

### Minimal config

(Updated with fixed minimal config: 2022-09-07 22:53 ET)

```lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
    local null_ls = require("null-ls")
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = { null_ls.builtins.diagnostics.todo_comments },
        debug = true,
    })
end

local function load_plugins()
    -- only add other plugins if they are necessary to reproduce the issue
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {
                "jose-elias-alvarez/null-ls.nvim",
                requires = {
			"nvim-lua/plenary.nvim",
			{
				"nvim-treesitter/nvim-treesitter",
[...]

PR #1088, opened 2022-09-08
fix(diagnostics): parse all file for todo-comments

danilshvalov:
Refers to #1085

### DIFF
--- a/lua/null-ls/builtins/diagnostics/todo_comments.lua
+++ b/lua/null-ls/builtins/diagnostics/todo_comments.lua
@@ -87,7 +87,7 @@ return h.make_builtin({
                 local content = vim.treesitter.get_node_text(node, params.bufnr):match("^%s*(.*)")
 
                 for kw, _ in pairs(keyword_by_name) do
-                    if content:match("%f[%a]" .. kw .. "%f[%A]") then
+                    if content:match("%f[%a]" .. kw .. "%f[%A]") and node:start() then
                         local row, _, _ = node:start()
 
                         table.insert(result, {
--- a/lua/null-ls/builtins/diagnostics/todo_comments.lua
+++ b/lua/null-ls/builtins/diagnostics/todo_comments.lua
@@ -18,18 +18,14 @@ local function get_document_root(bufnr)
     return tree:root()
 end
 
-local function parse_comments(root)
-    local output = {}
-
+local function parse_comments(root, output)
     for node in root:iter_children() do
-        if node:type() == "comment" then
-            table.insert(output, parse_comments(node))
-        elseif node:type() == "comment_content" then
-            return node
+        if node:type() == "comment_content" then
+            table.insert(output, node)
+        else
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez (approved):
LGTM, and this fixes the example from the original issue. Let's wait for feedback from the original reporter.

heygarrett:
This fixes the issue for Lua, but I'm still not seeing any diagnostics for other languages (I tried TypeScript and Python). Could I be holding it wrong?

danilshvalov:
Have you installed treesitter parser for these languages?

heygarrett:
Yeah, I even tried uninstalling and reinstalling the TypeScript parser. No luck. Any other information I can provide?

Also should I be leaving these comments on the issue rather than the PR?

jose-elias-alvarez:
For TypeScript, it looks like the node's type is `comment`. I am not a Treesitter expert, but does this diff make sense? It seems to make things work on my end:

```diff
diff --git a/lua/null-ls/builtins/diagnostics/todo_comments.lua b/lua/null-ls/builtins/diagnostics/todo_comments.lua
index cfa747f..82682bd 100644
--- a/lua/null-ls/builtins/diagnostics/todo_comments.lua
+++ b/lua/null-ls/builtins/diagnostics/todo_comments.lua
@@ -4,6 +4,8 @@ local methods = require("null-ls.methods")
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 local SEVERITIES = h.diagnostics.severities
 
+local comment_types = { "comment", "comment_content" }
+
 local function get_document_root(bufnr)
     local has_parser, parser = pcall(vim.treesitter.get_parser, bufnr)
     if not has_parser then
@@ -20,7 +22,7 @@ end
 
 local function parse_comments(root, output)
     for node in root:iter_children() do
-        if node:type() == "comment_content" then
+        if vim.tbl_contains(comment_types, node:type()) then
             table.insert(output, node)
         else
             parse_comments(node, output)
```

heygarrett:
@jose-elias-alvarez Yup, that works for me as well!

heygarrett:
Maybe unrelated, but the diagnostic is applied to the whole line (including preceding whitespace) rather than just the comment content. Is that expected?

danilshvalov:
Are you talking about a message after the TODO keyword?

heygarrett:
Screen Shot 2022-09-09 at 16 18 01

My diagnostics use underlines, and the whole line is getting underlined. I suppose I'd only expect the keyword itself (in this screenshot `TODO`) to get underlined (or maybe the keyword and the text that follows it).

danilshvalov:
Can you try the latest patch?

heygarrett:
That's great! Thanks for the quick fix. I'd say this is ready to merge. 👍

jose-elias-alvarez:
Excellent, thank you @danilshvalov for the improvements!

PR #1095, opened 2022-09-11
feat(prettier): Add MDX filetype

briandipalma:


jose-elias-alvarez:
Thanks! I think the CI error is caused by upstream changes, so I'll solve it separately.

PR #1099, opened 2022-09-12
Improve readme

mgred:
1. fix typo:  _null-is_ -> _null-ls_
2. use anchor links, so users do not need to scroll to the debug section and can jump right in


jose-elias-alvarez:
Thanks!

PR #1101, opened 2022-09-13
feat(builtins): allow configuring diagnostics display

### LINKED ISSUE #1094: Set `vim.diagnostic` per server
### Issues

### Feature description

Small feature request.

In standard LSP setups, you may set custom LSP handlers as so:
```lua
local format_handlers = {
    ['textDocument/publishDiagnostics'] = vim.lsp.with(vim.lsp.diagnostic.on_publish_diagnostics, {
        underline = true,
        virtual_text = false,
        signs = true,
        update_in_insert = false,
        severity_sort = true,
    }),
    ['window/showMessageRequest'] = function(_, result)
        return result
    end,
}

require('lspconfig').server_name.setup({ handlers = format_handlers })
```

Browsing previous issues (https://github.com/jose-elias-alvarez/null-ls.nvim/issues/525), it seems like null-ls uses the new `vim.diagnostic` API. We can replicate the above behavior for null-ls via namespaces (shown below), but I'd love a way to configure diagnostic display on a per-server basis. (I frequently turn off virtual text for noisy linters).

Supporting `handlers` in the `.with({})` config for sources seems like it might be a simple fix?

Alternatively, if this capability already exists, would love for someone to point me to the way.

Thanks!

```lua
local namespaces = vim.diagnostic.get_namespaces()
for id, tbl in pairs(namespaces) do
    local name = string.lower(tbl.name)
    local idx = string.find(name, 'null')
    if idx ~= nil then
        vim.diagnostic.config({
            underline = true,
            virtual_text = false,
            signs = true,
            update_in_insert = false,
            severity_sort = true,
        }, id)
    end
end
```

### Help

No

### Implementation help

_No response_

jose-elias-alvarez:
Closes #1094. @mikatpt

null-ls uses separate diagnostic namespaces per source, which is a prerequisite for multiple-file diagnostics but makes configuration annoying. One option is to allow configuring source namespace generation (or expose existing namespaces more readily), but this increases the chances of collisions and unexpected behavior. 

This PR instead configures source namespaces using the `diagnostic_config` key when the namespace is first generated. The option is available per-source and globally, since null-ls' unique behavior means that there is currently no good way to configure the display of all null-ls diagnostics.

### DIFF
--- a/doc/BUILTIN_CONFIG.md
+++ b/doc/BUILTIN_CONFIG.md
@@ -191,6 +191,29 @@ local sources = {
 }
 ```
 
+### Diagnostic config
+
+You can configure how Neovim displays source diagnostics by setting
+`diagnostic_config`:
+
+```lua
+local sources = {
+    null_ls.builtins.diagnostics.shellcheck.with({
+        diagnostic_config = {
+            -- see :help vim.diagnostic.config()
+            underline = true,
+            virtual_text = false,
+            signs = true,
+            update_in_insert = false,
+            severity_sort = true,
+        },
+    }),
+}
+```
+
+- Specifying `diagnostic_config` for a built-in will override your global
+  `diagnostic_config` for that source.
+
 ### Diagnostics format
 
 For diagnostics sources, you can change the format of diagnostic messages by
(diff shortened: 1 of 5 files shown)


mikatpt (commented):
looks great! TYSM  🎉

PR #1107, opened 2022-09-14
feat(commitlint): remove external dependencies from source `commitlint`

kaiuri:
Previous version of this source depended on `@commitlint/config-conventional` and `commitlint-format-json`.

Signed-off-by: Uriel C. A. Fernandes <uriel.acioli@gmail.com>

### DIFF
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -94,7 +94,7 @@ return h.make_builtin({
         notes = {
             -- "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
             -- "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
-            [=[Needs a `commitlintrc.json` configuration file, [schema](https://json.schemastore.org/commitlintrc.json).]=],
+            -- [=[Needs a `commitlintrc.json` configuration file, [schema](https://json.schemastore.org/commitlintrc.json).]=],
         },
     },
     method = DIAGNOSTICS,
@@ -103,7 +103,7 @@ return h.make_builtin({
         command = "commitlint",
         -- args = { "--format", "commitlint-format-json" },
         -- NOTE: Maybe add a function for resolving the configuration path?
-        args = { "-g", vim.fs.normalize("~/.commitlintrc.json"), "--edit", "$FILENAME" },
+        args = { "--edit", "$FILENAME" },
         -- to_stdin = true,
         to_temp_file = true,
         format = "line",
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -82,7 +82,6 @@ local function on_output(params, done)
     local messages = lex(params.output)
     for i = 1, #messages do
         local message = messages[i]
-        local sev = severity(message)
         local diagnostic = {
             row = 1,
             col = 1,
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -3,7 +3,8 @@ local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
-local violation_to_diagnostic = function(violation)
+-- NOTE: Unused
+--[==[ local violation_to_diagnostic = function(violation)
     local diagnostic = {
         ruleId = violation.name,
         message = violation.message,
@@ -18,8 +19,10 @@ local violation_to_diagnostic = function(violation)
 
     return diagnostic
 end
+--]==]
 
-local handle_commitlint_output = function(params)
+-- NOTE: Unused
+--[==[ local handle_commitlint_output = function(params)
     if params.output and params.output.results then
         local output = params.output.results[1]
 
@@ -42,6 +45,46 @@ local handle_commitlint_output = function(params)
         return parser({ output = violations })
     end
 end
+--]==]
+
+---@param line string
+local function get_severity(line)
+    local severities = {
+        ["✖"] = vim.diagnostic.severity.ERROR,
+        ["⚠"] = vim.diagnostic.severity.WARN,
+        -- NOTE: ↓ INFO is currently disable since it doesn't seem to produce anything but the same link
+        -- ['^ⓘ'] = vim.diagnostic.severity.INFO
+    }
+    for pattern, severity in pairs(severities) do
+        -- NOTE: Can be  used to futher improve parsing.
+        local pos = { vim.regex("^" .. pattern):match_str(line) }
+
+        return pos ~= nil and severity
+    end
+end
+
+---@param line string: line from command output
+---@param _ {}: `params`
+local function on_output(line, _)
+    local valid_line = line:match("]$") and line
+
+    if not valid_line then
+        return nil
+    else
+        local severity = get_severity(valid_line)
+        return severity
+                and {
+                    row = 1,
+                    col = 1,
+                    source = "commitlint",
+                    message = valid_line:match("%s%s%s(.*)%[.*"),
+                    code = valid_line:match("%[.*%]"),
+                    severity = severity,
+                    filename = "",
+                }
+            or nil
+    end
+end
 
 return h.make_builtin({
     name = "commitlint",
@@ -49,21 +92,25 @@ return h.make_builtin({
         url = "https://commitlint.js.org",
         description = "commitlint checks if your commit messages meet the conventional commit format.",
         notes = {
-            "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
-            "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
+            -- "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
+            -- "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
+            [=[Needs a `commitlintrc.json` configuration file, [schema](https://json.schemastore.org/commitlintrc.json).]=],
         },
     },
     method = DIAGNOSTICS,
     filetypes = { "gitcommit" },
     generator_opts = {
         command = "commitlint",
-        args = { "--format", "commitlint-format-json" },
-        to_stdin = true,
-        format = "json",
+        -- args = { "--format", "commitlint-format-json" },
+        -- NOTE: Maybe add a function for resolving the configuration path?
+        args = { "-g", vim.fs.normalize("~/.commitlintrc.json"), "--edit", "$FILENAME" },
+        -- to_stdin = true,
+        to_temp_file = true,
+        format = "line",
         check_exit_code = function(code)
             return code <= 1
         end,
-        on_output = handle_commitlint_output,
+        on_output = on_output,
     },
     factory = h.generator_factory,
 })
--- a/lua/null-ls/builtins/diagnostics/commitlint.lua
+++ b/lua/null-ls/builtins/diagnostics/commitlint.lua
@@ -1,89 +1,100 @@
+---@diagnostic disable: undefined-global, unused-function, unused-local
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
-
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
--- NOTE: Unused
---[==[ local violation_to_diagnostic = function(violation)
-    local diagnostic = {
-        ruleId = violation.name,
-        message = violation.message,
-        level = violation.level,
-    }
-
-    if violation.name == "body-leading-blank" then
-        diagnostic.line = 2
-    elseif vim.startswith(violation.name, "body") then
-        diagnostic.line = 3
-    end
-
-    return diagnostic
-end
---]==]
-
--- NOTE: Unused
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -42,28 +45,72 @@ local handle_commitlint_output = function(params)
...
 end
+--]==]
+
+---@param line string
+local function get_severity(line)
+    local severities = {
+        ["✖"] = vim.diagnostic.severity.ERROR,
+        ["⚠"] = vim.diagnostic.severity.WARN,
+        -- NOTE: ↓ INFO is currently disable since it doesn't seem to produce anything but the same link
+        -- ['^ⓘ'] = vim.diagnostic.severity.INFO
```
jose-elias-alvarez, line 56: What do you mean by this? Are you saying this symbol can't be parsed?

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -42,28 +45,72 @@ local handle_commitlint_output = function(params)
...
     method = DIAGNOSTICS,
     filetypes = { "gitcommit" },
     generator_opts = {
         command = "commitlint",
-        args = { "--format", "commitlint-format-json" },
-        to_stdin = true,
-        format = "json",
+        -- args = { "--format", "commitlint-format-json" },
+        -- NOTE: Maybe add a function for resolving the configuration path?
+        args = { "-g", vim.fs.normalize("~/.commitlintrc.json"), "--edit", "$FILENAME" },
```
jose-elias-alvarez, line 106: Generally, we have avoided defining default config paths like this, since for users to define their own config paths, they need to override all the arguments. I would rather have users configure this using `extra_args` . We can mention the need to specify a config path in the documentation and provide an example. 

Edit: it also seems like we can leave this off and `commitlint` will automatically resolve the config file from its `cwd` - maybe that's because it's using a temp file now, in which case that change makes sense.

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -42,28 +45,72 @@ local handle_commitlint_output = function(params)
...
     name = "commitlint",
     meta = {
         url = "https://commitlint.js.org",
         description = "commitlint checks if your commit messages meet the conventional commit format.",
         notes = {
-            "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
-            "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
+            -- "Needs npm packages commitlint and a json formatter: `@commitlint/{config-conventional,cli}` and `commitlint-format-json`.",
+            -- "It works with the packages installed globally but watch out for [some common issues](https://github.com/conventional-changelog/commitlint/issues/613).",
+            [=[Needs a `commitlintrc.json` configuration file, [schema](https://json.schemastore.org/commitlintrc.json).]=],
```
jose-elias-alvarez, line 97: It seems like [the configuration can also be a JavaScript file](https://commitlint.js.org/#/reference-configuration).

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -42,28 +45,72 @@ local handle_commitlint_output = function(params)
...
     generator_opts = {
         command = "commitlint",
-        args = { "--format", "commitlint-format-json" },
-        to_stdin = true,
-        format = "json",
+        -- args = { "--format", "commitlint-format-json" },
+        -- NOTE: Maybe add a function for resolving the configuration path?
+        args = { "-g", vim.fs.normalize("~/.commitlintrc.json"), "--edit", "$FILENAME" },
+        -- to_stdin = true,
+        to_temp_file = true,
```
jose-elias-alvarez, line 108: Was this change necessary? It seems to have good `stdin` support.

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -42,28 +45,72 @@ local handle_commitlint_output = function(params)
...
+        local pos = { vim.regex("^" .. pattern):match_str(line) }
+
+        return pos ~= nil and severity
+    end
+end
+
+---@param line string: line from command output
+---@param _ {}: `params`
+local function on_output(line, _)
+    local valid_line = line:match("]$") and line
```
jose-elias-alvarez, line 69: I noticed while testing that the errors created when `commitlint` fails to run aren't handled, most likely because of this check. For example, this output: 

```
✖   Please add rules to your `commitlint.config.js`
    - Getting started guide: https://commitlint.js.org/#/?id=getting-started
    - Example config: https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js [empty-rules]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

Results in just a single diagnostic: 

```
{ {
    bufnr = 4,
    code = "[empty-rules]",
    col = 0,
    end_col = 0,
    end_lnum = 1,
    filename = "",
    lnum = 0,
    message = " - Example config: https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js ",
    row = 1,
    severity = 2,
    source = "commitlint"
  } }
```

I wonder if there's a better way to check this than to match against the closing bracket. For reference, the current JSON implementation does handle this properly, and I think it would be useful to show the information, since this linter has so many weird pitfalls that can cause it to not work.

jose-elias-alvarez (commented):
Looks good overall! I think it's good to remove a dependency, especially since this linter is already a pain to configure. I think we can go ahead and remove the commented-out sections, too.

lua/null-ls/builtins/diagnostics/commitlint.lua
kaiuri: It can be parsed, but its message is always the same

```
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

I thought it best to leave it commented out until we figure out where to put that message.

lua/null-ls/builtins/diagnostics/commitlint.lua
kaiuri: This linter does have _many_ weird pitfalls. Using it through its API, its raw output is actually a .json object.

Given this commit message:

```
feaT:   
```

When called from a script, commitlint's linter, `@commitlint/lint`, outputs:

```json
{
  "valid": false,
  "errors": [
    {
      "level": 2,
      "valid": false,
      "name": "subject-empty",
      "message": "subject may not be empty"
    },
    {
      "level": 2,
      "valid": false,
      "name": "type-empty",
      "message": "type may not be empty"
    }
  ],
  "warnings": [],
  "input": "feaT:"
}
```

To get reproduce it:

If we can get the raw output... Will do some research here.

kaiuri: I think there is. I'd suggest, instead of matching lines, getting the `raw` output and splitting its contents by ourselves.

lua/null-ls/builtins/diagnostics/commitlint.lua
kaiuri: It does, we should remove that argument.

lua/null-ls/builtins/diagnostics/commitlint.lua
kaiuri: I've removed the configuration note as the `-g` option is unneeded because of the temp file. Will leave a link at `doc/BUILTINS.md` as a reference for users.

lua/null-ls/builtins/diagnostics/commitlint.lua
jose-elias-alvarez: This would be ideal, but do you think it's possible? It may be worth opening an issue on their repository. The default CLI output is remarkably unfriendly to parsers, so any other format would be an improvement.

kaiuri: ### Parsing the raw string

Because of Lua's snappiness I suppose parsing the raw string would _not_ have too much of an overhead and parsing it raw would be a matter of:

1. Identifying `⧗` and disposing of its line.
2. Identifying `✖`/`⚠` and reading thru the string until the next `✖`/`⚠`, each reading than could be considered a diagnostics string and be parsed accordingly.
3. Appending the `ⓘ...` chunk to the end of each diagnostic parsed above on its message field.

I'll give this a try once I have some time to spare, probably today or tomorrow.

---

### Extra thoughts

#### Outputting parser-friendly diagnostics

On account on making the `@commitlint/cli` output raw diagnostics, I'm not sure it's possible. Skimming over its source code, I've found both `@commitlint/cli` and `@commitlint/format` - the latter which is responsible for the unfriendly output - to be deeply coupled together. Though putting together a CLI that just lints but does not format isn't hard, [here](https://gist.github.com/kaiuri/2461997d54b2768225177f71880e199a)'s a tiny gist how it'd might be done.

I fear this would all be out of `null-ls.nvim`'s scope, though its could work with through a neovim's node_provider. Though, sadly it doesn't work with `pnpm` and I'm kind of locked out on this one.

jose-elias-alvarez: Interesting idea! As you say I think it's out-of-scope for us here, but I think it's worth at least opening an issue upstream and seeing if there's interest, since I'm sure other integrations would appreciate it.

kaiuri:
### Update

Managed to parse each message from `format = "raw"`. Now I'm looking to parse each message's severity so as not to output a bunch of same severity messages.

kaiuri:
@jose-elias-alvarez, Sorry for the delay, it's been a busy few days. Before I commit anything I'd like your opinion on this chunk I wrote that parses multi line `@commitlint/cli` messages. Any suggestions, perceived bottlenecks you might find or anything really.

```lua
local COMMIT_MSG = [[
✖   Please add rules to your `commitlint.config.js`
    - Getting started guide: https://commitlint.js.org/#/?id=getting-started
    - Example config: https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js [empty-rules]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
]]

---Checks if a line is the start of a diagnostic's message or not.
---@param line string
---@return boolean
local function message_start(line)
    local symbols = {
        error = "✖",
        warn  = "⚠",
        hint  = "ⓘ",
        info  = "⧗",
    }
    if line == nil then return false end

    local line_start = line:sub(1, 3) -- each symbol has length 3
    for _, v in pairs(symbols) do
        if line_start == v then
            return true
        end
    end

    return false
end

---Parses @commitlint/cli's raw output into single messages that can later be categorized by severity.
---@param output string
---@return string[]
local function parse(output)

    ---@type string[]: Non-empty string array.
    local lines = vim.tbl_filter(function(s)
        return s ~= "" and true or false
    end, vim.split(output, "\n"))

    ---@type string[]: Array of diagnostic messages.
    local messages = {}
    for i = 1, #lines do
        local message = ""
        local curr_line = lines[i]
        if message_start(curr_line) then
            message = message .. curr_line

            --- Go to next line, but only do anything if we're on the same message, else, go back up
            i = i + 1
            while not message_start(lines[i])
                and lines[i] ~= nil do
                message = message
                    .. "\n" -- keep message formatting
                    .. lines[i]
                i = i + 1
            end
        end

        table.insert(messages, message)
    end

    return messages
end

local parsed = parse(COMMIT_MSG)
for _, v in pairs(parsed) do
    print(v)
end

--[[ OUTPUT:
✖   Please add rules to your `commitlint.config.js`
    - Getting started guide: https://commitlint.js.org/#/?id=getting-started
    - Example config: https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/config-conventional/index.js [empty-rules]
✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
]]
```

kaiuri:
@jose-elias-alvarez, it's done, looking forward to your review.

jose-elias-alvarez:
Not a problem, I'm also having a busy week on my end but will review when I can!

lua/null-ls/builtins/diagnostics/commitlint.lua
```diff
@@ -1,69 +1,118 @@
...
+    local diagnostics = {}
+
+    local messages = lex(params.output)
+    for i = 1, #messages do
+        local message = messages[i]
+        local diagnostic = {
+            row = 1,
+            col = 1,
+            source = "commitlint",
+            message = message:match("%s%s%s(.*)"),
```
jose-elias-alvarez, line 89:
```suggestion
            message = message:match("%s%s%s(.*)") or "",
```

This caused issues for me, since Neovim expects this to always be defined.

jose-elias-alvarez (changes requested):
This is an impressive piece of work, but honestly I'm not 100% sure this approach is worth the headache. Here is the output from a bad commit message with this PR:

Screen Shot 2022-09-30 at 8 11 12 PM

And here is the output with the current JSON implementation (note the error on the 2nd line):

Screen Shot 2022-09-30 at 8 09 38 PM

I suspect that your time and skills would be better put to use contributing a sane, parseable CLI output format to `commitlint` itself (which would benefit other consumers, too). I would love to remove the extra dependency here, but as it stands, I think this needs more work (and ideally a lot of test coverage) to match the current implementation.

lua/null-ls/builtins/diagnostics/commitlint.lua
kaiuri: Thanks for the compliment! I agree. I'm of the opinion short is always best, therefore maybe is best to keep using the previous `built-in` version.
I'll look into contributing a sensible, parseable CLI, output format to `@commitlint/cli`.

Thanks for the review @jose-elias-alvarez

jose-elias-alvarez:
Closing this and the parent issue for the reasons discussed above. Thanks again!

PR #1111, opened 2022-09-15
feat(diagnostics): add perlimports as a linter

WhoIsSethDaniel:


lua/null-ls/builtins/diagnostics/perlimports.lua
```diff
@@ -0,0 +1,40 @@
...
+        command = "perlimports",
+        to_stdin = true,
+        from_stderr = true,
+        format = "line",
+        args = { "--lint", "--read-stdin", "--filename", "$FILENAME" },
+        timeout = 5000, -- this can take a long time
+        check_exit_code = function(code, stderr)
+            local success = code <= 1
+            if not success then
+                print(stderr)
```
jose-elias-alvarez, line 24: Do you think this information is useful enough to surface it to the user? Since we're running the linter on change, I'd worry about spamming messages created by running when the file is in an incomplete state.

WhoIsSethDaniel: if you mean the 'print' I have no problem removing it. I just cargo-culted it from somewhere else. :-/

WhoIsSethDaniel: I changed `check_exit_code` to be a table.

jose-elias-alvarez:
Thanks!

PR #1112, opened 2022-09-15
Add --stdin-filename to mix formatter

blubber:
This allows the mix format command to correctly format non-elixir files, such as heex templates.

### DIFF
--- a/lua/null-ls/builtins/formatting/mix.lua
+++ b/lua/null-ls/builtins/formatting/mix.lua
@@ -13,7 +13,7 @@ return h.make_builtin({
     filetypes = { "elixir" },
     generator_opts = {
         command = "mix",
-        args = { "format", "-" },
+        args = { "format", "--stdin-filename", "$FILENAME", "-" },
         format = "raw",
         to_stdin = true,
     },


jose-elias-alvarez:
Thanks!

dkarter:
@blubber @jose-elias-alvarez this PR broke the ability to format Elixir < 1.14

This may affect quite a few users.

I added a fix using conditional params in my config but I'm not sure this is the right approach or if we should revert this change and just have developers using 1.14 add `extra_args = { '--stdin-filename', '$FILENAME' }` and document that in the "built-ins" docs for `mix`.

This is my change
https://github.com/dkarter/dotfiles/blob/402a4e976762d9055589a55babd5691e711beebd/config/nvim/lua/plugins/null-ls/init.lua#L54-L69

If you like my approach I'm happy to open a PR (and make the version check a bit more robust)

blubber:
Bummer, I'm quite new to Elixir and didn't realize that 1.14 was so new, in fact, it is the only version I've ever used. I think reverting it is the best option here, since we really can't depend on people using 1.14.

It's a little nasty that mix format fails silently. I ran into this exact same problem when I was setting up mix formatting for heex, it worked when I ran it manually, but not from null-ls. That's why I created the PR. The silent failure is a bit of a problem.

PR #1117, opened 2022-09-18
fix(loop): clean up temp files on exit

jose-elias-alvarez:
Fixes an issue raised in #1075. Temp file-based generators created using `generator_factory` will only call `cleanup` on `stdout` end, so if the user exits Neovim after a generator has started running but before it's done, `cleanup` will not be called. This PR hooks into `VimLeavePre` to ensure that `cleanup` is called in this edge case (shout out to the autocommand API, without which this would have been a pain). 

@soifou I tested this using artificial delays / conditions, but could you give this a try and let me know if it solves the issue for you?


soifou:
LGTM! No more temp files, thanks a lot!

PR #1126, opened 2022-09-20
fix(phpmd) hide stderr from notifications

przepompownia:


przepompownia:
This is a starting point to further discussion started on https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/1122#discussioncomment-3683045

lua/null-ls/builtins/diagnostics/phpmd.lua
```diff
@@ -12,12 +12,15 @@ return h.make_builtin({
...
-        format = "json",
+        format = "raw",
         to_temp_file = true,
         check_exit_code = function(code)
             return code <= 3
         end,
-        on_output = function(params)
+        on_output = function(params, done)
+            local output, err = params.output, params.err
+            local ok, parsed = pcall(vim.json.decode, output)
```
jose-elias-alvarez, line 22: One option here would be to create a diagnostic from the linter's error output, which would surface it to users at the top of the buffer:

```lua
if err then
    return done({ { message = "phpmd error: " .. params.err } })
end
```

We could also add a message encouraging users to read the log. Either way, this seems like a simple way to prevent the linter from erroring out while still making it clear to users that something went wrong.

przepompownia: I added some short message (at the moment without a prompt to see the log) instead of appending the whole stderr to the first line diagnostics table.

przepompownia: `log:error()` doest two thing: writes to the log file and runs `vim.notify`. It'd be good to have a separated methods.

jose-elias-alvarez: Agreed, makes sense as a future improvement. This looks good to me, and if it solves your issue for now then I'm happy to merge this.

przepompownia: @jose-elias-alvarez for me, it can be merged.

I found an (probably separate) issue after opening `non/existing/file.php` (`not/existing` really does not exist) but it seems to be not related to this change (i.e. it occurs also after commenting out `phpmd` from `sources`).

jose-elias-alvarez: Thanks, yes, it's unrelated and should be fixed by 9ed430b5430aded48a68dfb2fc42a36fd562a9e1.

przepompownia: @jose-elias-alvarez thanks for support!

PR #1128, opened 2022-09-21
Update markdownlint.lua

briandipalma:
I know I can configure this locally but this tool isn't designed for MDX and it's feedback is quite noisy/of negative value in most MDX files so I feel there is a case to be made that it should be default configuration.


jose-elias-alvarez:
Sure, I think this makes sense. Thanks!

Issue #1123, opened 2022-09-19
sqlfluff formatting inserts error text into file


Alvinferdi990:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.7.2 Build type: Release LuaJIT 2.1.0-beta3 Compiled by builduser

### Operating System

manjaro

### Minimal config

-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
    local null_ls = require("null-ls")
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = {
require("null-ls").builtins.formatting.sqlfluff.with{extra_args = { "--dialect", "postgres" }},
        },
        debug = true,
    })
end
 
local function load_plugins()
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = {
require("null-ls").builtins.formatting.sqlfluff.with{extra_args = { "--dialect", "postgres" }},
        },
        debug = true,
    })
end
 
local function load_plugins()
    -- only add other plugins if they are necessary to reproduce the issue
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {
                "jose-elias-alvarez/null-ls.nvim",
                requires = { "nvim-lua/plenary.nvim" },
                config = null_ls_config,
            },
        },
        config = {
            package_root = package_root,
            compile_path = compile_path,
        },
    })
end
 
if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #1137, opened 2022-09-26
fix: Update the sqlfluff diagnostics built-in

KingMichaelPark:
- The sqlfluff diagnostics linter was treating 1 exit codes as passing which was incorrect.
- I have adjusted the output method to conform to the way sqlfluff outputs diagnostics to stderr.
- Made default arguments more explicit instead of shortened flag

### DIFF
--- a/lua/null-ls/builtins/diagnostics/sqlfluff.lua
+++ b/lua/null-ls/builtins/diagnostics/sqlfluff.lua
@@ -22,18 +22,18 @@ local sources = {
         command = "sqlfluff",
         args = {
             "lint",
-            "-f",
-            "github-annotation",
-            "-n",
+            "--format=github-annotation",
+            "--nocolor",
             "--disable_progress_bar",
             "$FILENAME",
         },
-        from_stderr = false,
+        from_stderr = true,
         to_stdin = false,
-        to_temp_file = true,
+        to_temp_file = false,
         format = "json",
+        -- Return codes https://github.com/sqlfluff/sqlfluff/blob/main/src/sqlfluff/cli/__init__.py
         check_exit_code = function(c)
-            return c <= 1
+            return c < 1
         end,
         on_output = helpers.diagnostics.from_json({
             attributes = {
--- a/lua/null-ls/helpers/generator_factory.lua
+++ b/lua/null-ls/helpers/generator_factory.lua
@@ -51,8 +51,26 @@ local parse_args = function(args, params)
     return parsed
 end
 
+local function find_json(output)
+    local bracket = string.find(output, "%[")
+    local parenthesis = string.find(output, "{")
+    local char = ""
+    if (bracket or parenthesis) and bracket < parenthesis then
+        char = "%["
+    else
+        char = "{"
+    end
+    if output then
+        if output[0] ~= char then
+            output = string.sub(output, string.find(output, char), string.len(output))
+        end
+    end
+    return output
+end
+
 local json_output_wrapper = function(params, done, on_output, format)
     if params.output then
+        params.output = find_json(params.output)
         local ok, decoded = pcall(vim.json.decode, params.output)
         if decoded == vim.NIL or decoded == "" then
             decoded = nil


jose-elias-alvarez:
Thanks! This looks good to me. @Alvinferdi990 could you verify whether this fixes your issue?

syphar:
I was just testing it (and it is still broken), then I realized one thing: 

this PR is changing the _diagnostics_ source for sqlfluff, not the _formatting_ source. So there is no way this can fix the formatting issue

KingMichaelPark:
Okay, it must be that both were broken because the debugging behaviour was incorrect as well on postgres as an input source, hold off on merging this for now, I will check the formatter as well.

KingMichaelPark:
@syphar could you send me an example query? It seems that I can format things with postgres without any changes to my config, it does only format if it doesn't have any diagnostic issues which I suppose makes sense

syphar:
@KingMichaelPark I used the example query from #1123 , using the postgres dialect: https://github.com/syphar/dotfiles/blob/1a3cbb0e34e336bc3c109236056ce00065e4f475/.config/nvim/lua/dc/lsp/null_ls.lua#L124-L126

KingMichaelPark:
Okay I couldn't get it to insert debug issues (replicating your issue) when running but I have added extra precautions to the json parser for bad output when the `--github-anotations` flag seems to append additional - non-json - output in certain situations . It might be useful to have a more experienced sqlfluff dev check why this happens. @alanmcruickshank I don't think you need to contribute necessarily unless scanning this for a few minutes would lead to something obvious that I am missing.

syphar:
just a thought: wouldn't it be the best solution to just fix sqlfluff here?

KingMichaelPark:
The first commit is fine regardless but the second I agree should be raised there if you don't mind.

alanmcruickshank:
If you find examples of when sqlfluff outputs inappropriate stuff on either stdout or stderr, do post them a issues on the sqlfluff GitHub - they should be fairly easy to squash. The intent is that it should play nice with use cases like this.

KingMichaelPark:
Will raise a ticket there with this example then, thank you @alanmcruickshank  I appreciate it.

image

jose-elias-alvarez:
Thanks @KingMichaelPark and @alanmcruickshank, was the conclusion here to fix the behavior upstream, then?

KingMichaelPark:
It needs more investigating, my config was disallowing me from seeing the issue but after seeing it I can confirm that the sqlfluff fix is inserting that parsing error he has reported to the file and no combination of changing the formatting options remedies it so I will have to check the sqlfluff code to see what is happening.

Michael Park
Senior Data Engineer
[The Data Shed Logo]
________________________________
From: Jose Alvarez ***@***.***>
Sent: 29 September 2022 14:40
To: jose-elias-alvarez/null-ls.nvim ***@***.***>
Cc: Michael Park ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL]Re: [jose-elias-alvarez/null-ls.nvim] fix: Update the sqlfluff diagnostics built-in (PR #1137)


Thanks @KingMichaelPark<https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FKingMichaelPark&data=05%7C01%7Cmichaelpark%40thedatashed.co.uk%7C6600873a48fd41417baf08daa2201fd3%7Cf420563810bb4ebda523e746a01ddade%7C0%7C0%7C638000556098797552%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=MTnTooGff5kP8I3soBwRoqpBSR4EgZZBiufsw2yFYzA%3D&reserved=0> and @alanmcruickshank<https://eur03.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Falanmcruickshank&data=05%7C01%7Cmichaelpark%40thedatashed.co.uk%7C6600873a48fd41417baf08daa2201fd3%7Cf420563810bb4ebda523e746a01ddade%7C0%7C0%7C638000556098797552%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=Q9YDdmQ34H01cVWQglRUGrJ5qzVQz15xeFIisOpHbA4%3D&reserved=0>, was the conclusion here to fix the behavior upstream, then?

syphar:
I created an upstream issue with sqlfluff: 
https://github.com/sqlfluff/sqlfluff/issues/3892

cc @alanmcruickshank

syphar:
@jose-elias-alvarez fyi, 

setting `ignore_stderr = false` and the fixed `check_exit_code` on the source will lead to the call just failing without inserting the error message into the text. 

Though in this case it feels like the sqlfluff output is something we should handle, because it's some intermediate result.

PR #1142, opened 2022-09-28
fix(pylint): offset end_col as well

Dece:
Pylint diagnostics show with the correct start column, but the end column is off by one; this fixes it.


jose-elias-alvarez:
Thanks!

PR #1154, opened 2022-10-04
feat: add word to cspell.json

marianozunino:
Allows to add a word as an exception into the cspell.json file on the cwd.

For instance, I'm using the cspell `diagnostics/cod_actions` only if in the cwd there's a cspell.json.

```lua
	null_ls.builtins.diagnostics.cspell.with({
		condition = function(utils)
			return utils.root_has_file({ "cspell.json" })
		end,
	}),
		condition = function(utils)
			return utils.root_has_file({ "cspell.json" })
		end,
	}),
```


vuki656:
There is more than one way to name the JSON config file as stated [here](https://cspell.org/configuration/).

Would it be possible to support all the JSON variants like `.cspell.json` and `.cspell.config.json`? 

Thanks for the PR.

yoo:
Thanks for the PR. Instead of supporting all the variants, allow defining a callback to configure the config file path. This would allow global config/dictionary files. The default would be `vim.fn.findfile("cspell.json", vim.fn.getcwd() .. ";")`.

marianozunino:
I like @yoo idea of allowing a callback to be passed as part of the config, instead of searching for each of the possibles files that cspell allows. Not sure how to do that though.

Can you guys me point me to an example?

vuki656:
There are 3 JSON variations, I think it would be redundant for each user to have to specify their own. Plus, if you configure it one way, then use a project where there is a different format, it's not going to work. 

You have `.cspell.json` `cspell.json` and `cspell.config.json` if the file check would be case insensitive.

jose-elias-alvarez:
I'm not a `cspell` user so I'm not sure what the right direction is with regards to finding the path to the JSON file, but FYI, once #1160 is merged, it will be possible to do the following:

```lua
local cspell = null_ls.builtins.code_actions.with({
    config = {
        find_json = function(params)
            return "path-to-json-file"
        end,
    },
})
```

The built-in can then do the following:

```lua
local config = params:get_config()
local find_json = config.find_json or function()
    vim.fn.findfile("cspell.json", vim.fn.getcwd() .. ";")
end

local cspell_json = find_json(params)
```

marianozunino:
> I'm not a `cspell` user so I'm not sure what the right direction is with regards to finding the path to the JSON file, but FYI, once #1160 is merged, it will be possible to do the following:

@jose-elias-alvarez thanks, that was really helpful. I'm going to hold this and wait for #1160. Updated the code though.

PumpedSardines:
Are there any updates on this?

jose-elias-alvarez:
Apologies for the delay on my end. I just merged #1160 which should hopefully unblock this (though there's still a few outstanding comments).

marianozunino:
@jose-elias-alvarez updated the PR. Not sure how to add tests though.

jose-elias-alvarez:
Pushed a small change to make sure we use the resolved `cwd` when available, but other than that this looks good. I'd love to add coverage but it's too complicated to set it up without stubbing everything out, so I think this is okay. Thanks for the work on this!

peterfication:
I'm not sure if I'm making something wrong, but for me, it only works if my cspell config file is just one line. So a multi line cspell config results in an error that it can't be parsed: `Expected object key string but found T_END at character 2`.

I'm on nightly Neovim, so this could be the issue.

Posting it here anyways, if others experience the same issue.

peterfication:
Changing

```lua
local ok, cspell = pcall(vim.json.decode, vim.fn.readfile(cspell_json_file)[1])
```

to

```lua
local ok, cspell = pcall(vim.json.decode, table.concat(vim.fn.readfile(cspell_json_file), " "))
```

does the job for me.

The written file is one line then, but that's not too bad for me, as I can format it before doing a commit.

PumpedSardines:
Shouldn't this feature also create a `.cspell.json` if none is found. Currently it just throws an error
Capture d’écran 2022-12-22 à 11 38 43

marianozunino:
> Currently it just thr

I don't want `null-ls` to start creating cspell config files on every folder in that I open a file with vim. Thus why I only enable this `builtin` where there is an existing `cspell.json`

Like so:

```lua
		condition = function(utils)
			return utils.root_has_file({ "cspell.json" })
		end
	}),
```

marianozunino:
> Shouldn't this feature also create a `.cspell.json` if none is found. Currently it just throws an error

I don't want `null-ls` to start creating cspell config files on every folder in that I open a file with vim. Thus why I only enable this `builtin` where there is an existing `cspell.json`

Like so:

```lua
		condition = function(utils)
			return utils.root_has_file({ "cspell.json" })
		end
	}),
```

PumpedSardines:
I didn't mean when you open vim, I meant if you select the autocomplete add to cspell.json. Because I feel like if you choose to autocomplete by adding to cspell.json, you want to have that file.

And doesn't doing what you just sent turn off the regular cspell autocomplete? Wouldn't you still want to be able to autocomplete individual words even if there isn't a cspell.json?

But besides that, shouldn't the autocomplete option first check if there is a file instead of throwing that big error?

juanpprieto:
Screenshot 2022-12-22 at 5 41 39 AM

Hey guys, been trying to get this to work but running into the above error. Any clues?

here is my `cspell.json` 
```json
{
  "version": "0.2",
  "language": "en",
  "words": [],
  "flagWords": []
}
```

I've just configure it via the standard options
```
...code_actions.cspell,
...diagnostics.cspell,
```

jose-elias-alvarez:
> I didn't mean when you open vim, I meant if you select the autocomplete add to cspell.json. Because I feel like if you choose to autocomplete by adding to cspell.json, you want to have that file.
> 
> And doesn't doing what you just sent turn off the regular cspell autocomplete? Wouldn't you still want to be able to autocomplete individual words even if there isn't a cspell.json?
> 
> But besides that, shouldn't the autocomplete option first check if there is a file instead of throwing that big error?

I agree with this - since the source contributes multiple code actions, one of which does not depend on a config file, we shouldn't require users to check for this. 

For now, I think we should make sure we are handling errors if the user selects the code action but doesn't have a config file. In the future, I don't think it's a big deal for us to write a config file if it doesn't exist, since as you noted the user has specifically selected the code action.

jose-elias-alvarez:
@juanpprieto The `Cannot parse` error is solved by #1306. I can't replicate your issue with `utils.root_has_file`, though. If you're still having problems, can you open a separate issue?

juanpprieto:
Thanks @jose-elias-alvarez — The latest JSON/parse update from @marianozunino worked! 🙏🏼

Issue #1048, opened 2022-08-26
on-demand `params` generator argument creation and deepcopying


haolian9:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

currently, each `params` passed to generator contains a `content` key which might hold huge bytes of data; even worse, the params will be deepcopied for each generator and each completion.

in my use, i rarely use the whole lines of current buffer, the cursor/current line is what i want all the time; and i never modify the `params`, so deepcopying it is unnecessary.

(i have scratched a [commit](https://github.com/haolian9/null-ls.nvim/commit/8affe88ee2f716024eb6a85c4c8d80773a3e876f) which added an option `copy_params: bool` for generator and `with_content: bool` for util.make_params, maybe that can explain my need better.)

### Help

No

jose-elias-alvarez:
Your implementation seems pretty reasonable, and I think this is worth looking into if we can profile and determine that it's worth the added complexity. A couple of questions:

1. How would we choose which methods should have `with_content = true`? Completion and hover seem like good candidates, since those results are based on the cursor position. We'd want to make sure to document this, too.
2. I added deep copying because I found that it was too easy for sources to leak and modify the same object, which caused unpredictable behavior. What do you think of changing the logic so that we only copy if more than one generator is using the same object?

haolian9:
sorry for the delay!
(i was also waiting if there are otheres interest in this, seems no one. because no new comments were appended.)

for 1)
 i'm still not clear on this, `with_content` or whatever name in my use is a hard code,
it should be configurable, but i was too lazy to make it that way in that time.
i personally only use diagnositc, completion parts of null-ls.nvim, they works fine with `nvim_get_current_line`, but i had met an error when it's called in formatter part.

for 2)
it's a trade off IMO. as an user, i clearly know what i'll do to the params, so i dont need this costly safety guard; but for users does not care about this, this guard is worth.
and that' s why i made the options i had added work like as default (adding content, deep copying, unless i set them to false explicitly).

jose-elias-alvarez:
Well, without data showing that this is actually a problem, I'm not sure whether it's worth moving forward with this. Making this configurable would basically break every built-in source we have here, so even if there is a performance issue, I'd want to think carefully about how to proceed. Performance-wise, I think avoiding deep copying unless multiple sources are active will make a difference.

[...]

PR #1160, opened 2022-10-05
feat: add methods and lazy resolution to params table

jose-elias-alvarez:
One of the current weak spots of null-ls that it's not possible for users to configure sources in a way that generators can pick up on. This issue came up in #538, but here's a more concrete example, a simple `write-good` like source to create diagnostics for words that the user wants to avoid:

```lua
local bad_words = { "really", "very" }

local bad_words_source = require("null-ls.helpers").make_builtin({
    filetypes = {},
    generator = {
        fn = function(params)
            local diagnostics = {}
            for i, line in ipairs(params.content) do
                for _, word in ipairs(bad_words) do
                    local col, end_col = line:find(word)
                    if col and end_col then
                        table.insert(diagnostics, {
                            row = i,
                            col = col,
                            end_col = end_col,
                            source = "bad-words",
                            message = string.format("Don't use '%s'!", word),
                            severity = vim.diagnostic.severity.ERROR,
                        })
                    end
                end
            end
            return diagnostics
        end,
    },
})
```

The issue is that the source is totally inflexible. If the user wants to change the words that are highlighted, they have to either copy-paste the entire source definition or maintain their own fork, neither of which are ideal. We could also use global variables like ALE or a separate `config` key in the user's null-ls config, but neither solution felt right. Users should be able to configure the source using the existing `with` method.

I initially concluded that solving this problem would require a larger rewrite to give `generator.fn` access to `self` (a change I've been dragging my feet on, since it's gigantic and will break most existing configs) but I realized that we could instead use the `params` table itself. By assigning a `source_id` to the table, and with a little metatable magic, we can now do this:

```lua
local default_words = { "really", "very" }

local bad_words_source = require("null-ls.helpers").make_builtin({
    filetypes = {},
    generator = {
        fn = function(params)
            local diagnostics = {}
            for i, line in ipairs(params.content) do
                for _, word in ipairs(params:get_config().words or default_words) do
                    local col, end_col = line:find(word)
                    if col and end_col then
                        table.insert(diagnostics, {
                            row = i,
                            col = col,
                            end_col = end_col,
                            source = "bad-words",
                            message = string.format("Don't use '%s'!", word),
                            severity = vim.diagnostic.severity.ERROR,
                        })
                    end
                end
            end
            return diagnostics
        end,
    },
})
```

Users can then do this:

```lua
local sources = {
    bad_words_source.with({
        config = {
            words = { "other", "words" },
        },
    }),
}
```

The `params` table works identically, with two new methods:

1. `params:get_source()` uses the source's ID to resolve and return the matching source
2. `params:get_config()` uses the above method to return `source.config` (or an empty table)

I also changed most of the keys on the `params` table to resolve lazily. This is an issue that came up in #1048, and while I'm not 100% convinced that it meaningfully affects performance, I think it's a nice change.

Tests are passing without any changes, which is a good sign that this won't take too much more work to merge. I do think a total rewrite of null-ls sources is ultimately necessary, but I think this is a step in the right direction (and one we can reuse in the rewrite).

- [x] Move `make_params` tests to a separate file and add coverage
- [x] Add documentation
- [x] Add a new built-in documentation key to make it easier for authors to document available config options


jose-elias-alvarez:
I think this is good to go. I think the documentation could still be clearer, but for now I added a simple example to the gitsigns.nvim built-in and used that for documentation / example purposes.

PR #1162, opened 2022-10-06
support for ocamlformat

c3n21:
I've made a small PR to add ocamlformat as formatter for null-ls.

Right now it can already format every ocaml file with the default settings, just need some time to dig in the docs 
(https://github.com/ocaml-ppx/ocamlformat/blob/main/ocamlformat-help.txt) to support .ocamlformat.

### DIFF
--- a/lua/null-ls/builtins/formatting/ocamlformat.lua
+++ b/lua/null-ls/builtins/formatting/ocamlformat.lua
@@ -1,0 +1,20 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "ocamlformat",
+    meta = {
+        url = "https://github.com/ocaml-ppx/ocamlformat",
+        description = "Auto-formatter for OCaml code",
+    },
+    method = FORMATTING,
+    filetypes = { "ocaml" },
+    generator_opts = {
+        command = "ocamlformat",
+        args = { "--enable-outside-detected-project", "-" },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lua/null-ls/builtins/formatting/ocamlformat.lua
```diff
@@ -0,0 +1,19 @@
...
+    name = "ocamlformat",
+    meta = {
+        url = "https://github.com/ocaml-ppx/ocamlformat",
+        description = "",
+    },
+    method = FORMATTING,
+    filetypes = { "ocaml" },
+    generator_opts = {
+        command = "ocamlformat",
+        args = { "--enable-outside-detected-project", "$FILENAME" },
```
jose-elias-alvarez, line 16:
```suggestion
        args = { "--enable-outside-detected-project", "-" },
        to_stdin = true
```

According to [the documentation](https://github.com/ocaml-ppx/ocamlformat/blob/78f1e72de0d1c1bda318589382f6dee201de2c3c/ocamlformat-help.txt#L646), it looks like this is supported.

jose-elias-alvarez (commented):
Thanks for the PR! I believe the [`--name` flag](https://github.com/ocaml-ppx/ocamlformat/blob/78f1e72de0d1c1bda318589382f6dee201de2c3c/ocamlformat-help.txt#L569) may provide help in resolving config files, since we can pass the file's content via `stdin` and the true path to the file using this flag.

lua/null-ls/builtins/formatting/ocamlformat.lua
jose-elias-alvarez: I could be wrong, but based on the documentation, I think this will work, since the formatter will use the path to the file to resolve the config file:

```lua
args = { "--enable-outside-detected-project", "--name", "$FILENAME", "-" },
```

c3n21: > I could be wrong, but based on the documentation, I think this will work, since the formatter will use the path to the file to resolve the config file:
> 
> ```lua
> args = { "--enable-outside-detected-project", "--name", "$FILENAME", "-" },
> ```

with this
```lua
        args = { "--enable-outside-detected-project", "-" },
        to_stdin = true
```
it works already (it can find .**ocamlformat** file even in parent directories and format files correctly)

jose-elias-alvarez: I assume this mechanism is based on looking at the directory where `ocamlformat` is spawned, so I do think it would be a good idea to add the flag to better handle complex project structures. But if you really think it's unnecessary then I can go ahead and merge this.

c3n21: In the docs it says:
> --name=NAME
>
>Name of input file for use in error reporting and starting point
>when searching for '.ocamlformat' files. Defaults to the input
>file name. Some options can be specified in configuration files
>named '.ocamlformat' in the same or a parent directory of NAME,
>see documentation of other options for details.

So I guess it should be implicit.

I'm no ocaml expert, I'm just settings up a minimal coding environment for my ocaml class.

jose-elias-alvarez: Sure, if it works for you let's go with this. Thanks!

PR #1165, opened 2022-10-06
refactor!: remove support for neovim 0.7

jose-elias-alvarez:
Pretty straightforward. I'll wait for a couple of weeks before merging this to give users time to update, and I'll create a `0.7-compat` branch for users who can't or don't want to update.

PR #1166, opened 2022-10-06
refactor: use native neovim 0.8 fs utils

jose-elias-alvarez:
I'll wait until #1165 is merged to merge this. The only change here is to the pattern used to escape wildcard characters in each path, which now matches nvim-lspconfig's implementation.


jose-elias-alvarez:
I'm going to wait about a week for potential issues from #1160 to shake out then merge this.

PR #1176, opened 2022-10-09
fix(builtins): refine 'cwd' for mypy and isort

wy-z:
### mypy
Mypy needs `cwd` to discover the configuration file (https://mypy.readthedocs.io/en/stable/config_file.html)

### isort
Isort will detect files in the CWD as first-party (https://github.com/PyCQA/isort/blob/fa2de4761040a9fd82339fccc7cc42391eb082cf/CHANGELOG.md?plain=1#L434),
So although `setup.py` doesn't contain `isort` configuration, it helps `isort` sort correctly.

### DIFF
--- a/lua/null-ls/builtins/formatting/isort.lua
+++ b/lua/null-ls/builtins/formatting/isort.lua
@@ -23,6 +23,8 @@ return h.make_builtin({
         to_stdin = true,
         cwd = h.cache.by_bufnr(function(params)
             return u.root_pattern(
+                -- isort will detect files in the CWD as first-party
+                "setup.py",
                 -- https://pycqa.github.io/isort/docs/configuration/config_files.html
                 "setup.cfg",
                 ".isort.cfg",
--- a/lua/null-ls/builtins/formatting/isort.lua
+++ b/lua/null-ls/builtins/formatting/isort.lua
@@ -24,11 +24,11 @@ return h.make_builtin({
         cwd = h.cache.by_bufnr(function(params)
             return u.root_pattern(
                 -- isort will detect files in the CWD as first-party
-                "setup.py",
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/formatting/isort.lua
```diff
@@ -23,6 +23,8 @@ return h.make_builtin({
         to_stdin = true,
         cwd = h.cache.by_bufnr(function(params)
             return u.root_pattern(
+                -- isort will detect files in the CWD as first-party
```
jose-elias-alvarez, line 26: One question: these are resolved in order, meaning that if `setup.py` is found, that will be used as the `cwd`, even if one of the other files is found. Does it make sense to reorder this so that it prefers files specific to `isort`?

wy-z: Thanks for your review, how about the following order?
1. .isort.cfg: Isort official  configuration file
2. pyproject.toml: Python's official recommended project configuration file (https://packaging.python.org/en/latest/tutorials/packaging-projects/#packaging-python-projects)
3. setup.py: Legacy project configuration file (https://docs.python.org/3/distutils/index.html)
4. setup.cfg: Configuration file that process after the contents of the setup.py (https://docs.python.org/3/distutils/configfile.html)
5. tox.ini: Unofficial configuration file, but there are quite a few tools supported
6. .editorconfig: Unofficial configuration file

References:
- https://github.com/PyCQA/isort/blob/72687ecf43c8fd7f8125ac35defe1d029b5ec497/isort/settings.py#L66
- https://pycqa.github.io/isort/docs/configuration/config_files.html

jose-elias-alvarez: Yep, I think this order makes sense. Thanks!

PR #1184, opened 2022-10-11
Fix filename for `prettier.config.js` variants

mskelton:
The Prettier filename variants `prettier.config.js` and `prettier.config.cjs` were incorrectly prefixed with a dot which is not what Prettier expects.  This renames them to the proper non-dot prefix filenames.


jose-elias-alvarez:
I see from [the documentation](https://prettier.io/docs/en/configuration.html) that this is the right change. Thanks!

PR #1185, opened 2022-10-11
feat: reset_hunk and stage_hunk support range

c3n21:
Hope it looks good for you

### DIFF
--- a/lua/null-ls/builtins/code_actions/gitsigns.lua
+++ b/lua/null-ls/builtins/code_actions/gitsigns.lua
@@ -2,6 +2,7 @@ local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 
 local CODE_ACTION = methods.internal.CODE_ACTION
+local range_actions = { ["reset_hunk"] = true, ["stage_hunk"] = true }
 
 return h.make_builtin({
     name = "gitsigns",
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/code_actions/gitsigns.lua
```diff
@@ -27,7 +28,14 @@ return h.make_builtin({
                 table.insert(actions, {
                     title = name_to_title(name),
                     action = function()
-                        vim.api.nvim_buf_call(params.bufnr, action)
+                        local cb = action
+                        if range_actions[name] then
```
jose-elias-alvarez, line 32:
```suggestion
                        if params.range and range_actions[name] then
```

This will only be defined if the original request has a range, so we should add a check.

jose-elias-alvarez:
Hmm, this is a little trickier than I thought. At some point, Neovim [started setting the `range` property](https://github.com/neovim/neovim/blob/45cc5fd7656c48606bf337a4cc01380bedc8ba36/runtime/lua/vim/lsp/buf.lua#L774) for all code action requests, so using `params.range` to determine which action to apply won't work. I think we have to do what Neovim does and check whether the user is in visual mode. 

Also, this is minor and can be saved for a follow-up, but I think it would be nice if the title of the code action changed from `Stage hunk` to `Stage selection`.

c3n21:
Now it should check if the mode is "v" or "V", if it is then it will perform the second check to see if it's an action that supports ranged actions.

c3n21:
Now when staging or resetting a range, the title will be respectively "Stage selection" and "Reset selection"

PR #1188, opened 2022-10-15
feat: add NullLsToggle command

kylo252:
add `NullLsToggle` command with completion support

### DIFF
--- a/lua/null-ls/init.lua
+++ b/lua/null-ls/init.lua
@@ -32,9 +32,23 @@ M.setup = function(user_config)
     vim.api.nvim_create_user_command("NullLsInfo", function()
         require("null-ls.info").show_window()
     end, {})
+
     vim.api.nvim_create_user_command("NullLsLog", function()
         vim.cmd(string.format("edit %s", require("null-ls.logger"):get_path()))
     end, {})
+
+    vim.api.nvim_create_user_command("NullLsToggle", function(opts)
+        M.toggle(opts.args)
+    end, {
+        nargs = 1,
+        complete = function()
+            local list = {}
+            for _, source in ipairs(M.get_sources()) do
+                list[#list + 1] = source.name
+            end
+            return list
+        end,
+    })
 
     local augroup = vim.api.nvim_create_augroup("NullLs", {})
     vim.api.nvim_create_autocmd("FileType", {


lua/null-ls/init.lua
```diff
@@ -32,10 +32,24 @@ M.setup = function(user_config)
...
         vim.cmd(string.format("edit %s", require("null-ls.logger"):get_path()))
     end, {})
 
+    vim.api.nvim_create_user_command("NullLsToggle", function(opts)
+        M.toggle(opts.args)
+    end, {
+        nargs = 1,
+        complete = function()
+            local list = {}
+            for _, source in ipairs(M.get_sources()) do
```
jose-elias-alvarez, line 46: Should this complete with sources active for the current filetype?

lua/null-ls/init.lua
```diff
@@ -32,10 +32,24 @@ M.setup = function(user_config)
...
+                list[#list + 1] = source.name
```
jose-elias-alvarez, line 47: I think there's a possibility of duplicate sources here, e.g. if the user has `eslint` registered for both formatting and diagnostics.

lua/null-ls/init.lua
```diff
@@ -32,10 +32,24 @@ M.setup = function(user_config)
...
     end, {})
+
     vim.api.nvim_create_user_command("NullLsLog", function()
         vim.cmd(string.format("edit %s", require("null-ls.logger"):get_path()))
     end, {})
 
+    vim.api.nvim_create_user_command("NullLsToggle", function(opts)
+        M.toggle(opts.args)
+    end, {
+        nargs = 1,
```
jose-elias-alvarez, line 43: I wonder if this should also do something without any arguments. In my head, it makes sense for it to either toggle all null-ls sources or the sources active for the current filetype (if we make the change to completion I mentioned in another comment).

lua/null-ls/init.lua
kylo252: I made it for all _registered_ sources for the current filetype, since it needs to re-enable them again.

lua/null-ls/init.lua
kylo252: this will depend on how `source.toggle()` should behave, and since it's only `eslint`, I think it's probably fine to leave it like this? 
otherwise, we'd need something similar to [how `LspStop` handles nargs](https://github.com/neovim/nvim-lspconfig/blob/3592f769f2d6b07ce3083744cd0a13442f5d4f43/plugin/lspconfig.lua#L102-L123)

jose-elias-alvarez: If we just pass `eslint` to `source.toggle()` it'll toggle all sources [matching the given query](https://github.com/kylo252/null-ls.nvim/blob/756b42b86ee65fff2b3e57c8dfec69bd690b38c0/lua/null-ls/sources.lua#L18), which in this case means it'll toggle all sources named `eslint`. I think this would be the least surprising behavior for users, and there's at least a few other sources that share names, so I think we should deduplicate them.

lua/null-ls/info.lua
```diff
@@ -102,6 +102,21 @@ M.show_window = function()
         return info_lines
     end
 
+    local create_inactive_sources_info = function(ft)
+        local info_lines = {
+            "Inactive source(s)",
+        }
+
+        for _, source in ipairs(sources.get(ft)) do
```
jose-elias-alvarez, line 110: I haven't taken a close look, but is this logic okay? When I have multiple active sources and toggle them, I only see the first source under `Inactive source(s)`.

kylo252: @jose-elias-alvarez, sorry for the delay!  turned out to be two separate issues..    :smile:

It should be fixed now!

jose-elias-alvarez (commented):
Sorry for the review back-and-forth on this! I found one bug but other than that I think this looks good.

PR #1198, opened 2022-10-19
mlint work fine for octave(the open source of matlab) too

ayoubelmhamdi:
i'm not familiar with `lua` or 'null-ls`, i tried this, and it's worked for me,


jose-elias-alvarez:
Thanks!

PR #1203, opened 2022-10-21
feat: add checkstyle builtin

### LINKED ISSUE #1193: Add checkstyle as a built-in
### Issues

### Feature description

Add [checkstyle](https://checkstyle.org) as a built-in diagnostics tool.

I already have a rough version working in my own configs.

### Help

Yes

### Implementation help

Checkstyle requires a config file to be specified. It ships some configs in its JAR file, which can be loaded by specifying a classpath resolvable file in the JAR. Though likely users would want to specify different configs per project.

So what I would need help with is how I can read a vim variable (read from a local vimrc for example) and use that to compose the arguments for the executable on each run.

Additionally I'd need some help on how to have this run only on written files, as it does not accept STDIN.

Finally, some performance tweaks might also be nice, since I did some involuntary performance tests as commented on #596.

kmoschcau:
Closes #1193

### DIFF
--- a/lua/null-ls/builtins/diagnostics/checkstyle.lua
+++ b/lua/null-ls/builtins/diagnostics/checkstyle.lua
@@ -1,0 +1,91 @@
+local h = require("null-ls.helpers")
+
+local function parse_checkstyle_errors(params, output)
+    if params.err:match("Must specify a config XML file.") then
+        table.insert(output, {
+            message = "You need to specify a configuration for checkstyle."
+                .. " See https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#checkstyle",
+            severity = vim.diagnostic.severity.ERROR,
+            bufnr = params.bufnr,
+        })
+        return
+    end
+
+    if params.err:match("Checkstyle ends with %d+ errors.") then
+        return
+    end
+
+    table.insert(output, {
+        message = vim.trim(params.err),
+        severity = vim.diagnostic.severity.ERROR,
+        bufnr = params.bufnr,
+    })
+end
+
+local function handle_checkstyle_output(params)
+    local output = {}
+
+    if params.err then
+        parse_checkstyle_errors(params, output)
+    end
+
+    local results = params.output and params.output.runs and params.output.runs[1] and params.output.runs[1].results
+        or {}
+
+    for _, result in ipairs(results) do
+        for _, location in ipairs(result.locations) do
+            table.insert(output, {
+                row = location.physicalLocation.region.startLine,
+                col = location.physicalLocation.region.startColumn,
+                end_col = location.physicalLocation.region.startColumn + 1,
+                code = result.ruleId,
+                message = result.message.text,
+                severity = h.diagnostics.severities[result.level],
+                filename = location.physicalLocation.artifactLocation.uri,
+            })
+        end
+    end
+
+    return output
+end
+
+local function check_exit_code(code, stderr)
+    return code == 0 or stderr:match("Checkstyle ends with " .. code .. " errors.")
+end
+
+return h.make_builtin({
+    name = "checkstyle",
+    meta = {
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/checkstyle.lua
```diff
@@ -0,0 +1,78 @@
...
+            "This uses `/google_checks.xml` by default. Override it using `with()` and `extra_args`.",
+        },
+    },
+    method = require("null-ls.methods").internal.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "java" },
+    generator_opts = {
+        args = { "-f", "sarif", "$ROOT" },
+        check_exit_code = check_exit_code,
+        command = "checkstyle",
+        extra_args = { "-c", "/google_checks.xml" },
```
jose-elias-alvarez, line 71: Rather than setting this as a default, I think it would be a bit clearer to specify that this is necessary, then put an example under `usage`, similar to what we do for [sqlfluff](https://github.com/kylo252/null-ls.nvim/blob/new-command/lua/null-ls/builtins/formatting/sqlfluff.lua).

lua/null-ls/builtins/diagnostics/checkstyle.lua
```diff
@@ -0,0 +1,78 @@
...
+return h.make_builtin({
+    name = "checkstyle",
+    meta = {
+        url = "https://checkstyle.org",
+        description = "Checkstyle is a tool for checking Java source code for"
+            .. " adherence to a Code Standard or set of validation rules (best"
+            .. " practices).",
+        notes = {
+            "This requires `java` to be in our path.",
+            "Checkstyle only offers a jar file as download. It is recommended"
```
jose-elias-alvarez, line 55: Using `[[` and `]]` to surround the string will let you avoid breaking lines manually, which will make it easier to read this in the future.

jose-elias-alvarez (commented):
Looks good overall! I left a couple of comments about documentation.

lua/null-ls/builtins/diagnostics/checkstyle.lua
kmoschcau: Does it also remove indent whitespace or do I have to shift the text to the left to avoid that?

lua/null-ls/builtins/diagnostics/checkstyle.lua
kmoschcau: I had a first version where I didn't pass this by default and intercepted the specific error this tool throws to show one pointing at the notes. Would you prefer it that way?

jose-elias-alvarez: Sure, that sounds like a clear way to handle this.

kmoschcau: I already updated both PRs to do that and included tests for it.

jose-elias-alvarez:
LGTM, thanks!

lewisdiamond:
Tried that, checkstyle doesn't return JSON so there's an error `failed to decode json: Expected value but found invalid token at character 1`

kmoschcau:
@lewisdiamond what do you mean you tried it? Did you follow the usage instructions? Neither checkstyle nor PMD work out of the box. If you share your null-ls config, I could have a look to help you.

lewisdiamond:
Checkstyle was getting lost in symlinks, had to change the root (hacked it in the bash script for now).

rolandsaav:
@lewisdiamond I'm running into this right now. Could you clarify exactly what you did to get it to work?

rolandsaav:
> @lewisdiamond I'm running into this right now. Could you clarify exactly what you did to get it to work?

Nevermind. I was doing this for a school project and realized that the issue is we were given an older version of checkstyle that didn't support the "sarif" output format. Posting this in case anyone else runs into this.

PR #1206, opened 2022-10-23
fix(builtins): set correct golangci_lint cwd

marcelbeumer:
I discovered this issue working on https://github.com/marcelbeumer/go-playground/tree/main/gochat (with cwd being `gochat`).

golintci_lint was running from the `$ROOT`, which, if I understand correctly, is set the the dir containing `.git`, by default. I think it should run in a root appropriate for the file being saved (in this case the dir `gochat` containing `go.mod`).


lua/null-ls/builtins/diagnostics/golangci_lint.lua
```diff
@@ -34,7 +36,8 @@ return h.make_builtin({
             local issues = params.output["Issues"]
             if type(issues) == "table" then
                 for _, d in ipairs(issues) do
-                    if d.Pos.Filename == params.bufname then
+                    local fname = params.cwd .. "/" .. d.Pos.Filename
```
marcelbeumer, line 39: doing this here instead of using `--path-prefix` so it's easier for consumers to modify the args with `.with({ args `

lua/null-ls/builtins/diagnostics/golangci_lint.lua
```diff
@@ -16,14 +16,16 @@ return h.make_builtin({
         to_stdin = true,
         from_stderr = false,
         ignore_stderr = true,
+        cwd = function(params)
```
jose-elias-alvarez, line 19: It's not a huge deal, since this linter runs (relatively) infrequently, but if we don't expect this result to change, I think we want to cache this to avoid repeated lookups. You may also want to consider using the `root_pattern` utility. [This builtin](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/4c3ba1cabdb761ad98d149ed48ee695aaf52e253/lua/null-ls/builtins/formatting/prettier_standard.lua#L21) has an example of both.

lua/null-ls/builtins/diagnostics/golangci_lint.lua
jose-elias-alvarez: Was this not working, or did the other change make this one necessary? Either way I think this is fine, though we'll want to use the `join_path` utility ([example](https://github.com/kylo252/null-ls.nvim/blob/756b42b86ee65fff2b3e57c8dfec69bd690b38c0/lua/null-ls/builtins/diagnostics/tsc.lua#L47)) to guarantee consistency.

PR #1207, opened 2022-10-23
add diagnostic support for vhdl with vcom and ghdl

medwatt:


lua/null-ls/builtins/diagnostics/ghdl.lua
```diff
@@ -0,0 +1,32 @@
...
+        description = "Adds support for `ghdl` VHDL compiler/checker",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "vhdl" },
+    generator_opts = {
+        command = "ghdl",
+        to_stdin = false,
+        to_temp_file = true,
+        from_stderr = true,
+        args = { "-a", "--std=08", "--workdir=/tmp", "$FILENAME" },
```
jose-elias-alvarez, line 19: I know nothing about this language or CLI tool, but judging from [the documentation](https://ghdl.github.io/ghdl/using/ImplementationOfVHDL.html#vhdl-standards), it looks like there are multiple standards. Is this something we want to always set, or should we leave it up to users to configure?

jose-elias-alvarez, line 19: Also, just so I understand: is the [`--workdir` flag](https://ghdl.github.io/ghdl/using/InvokingGHDL.html#cmdoption-ghdl-workdir) necessary because the program will otherwise compile and leave files in the current directory? If this is necessary then we may want to mention that the source will not work as expected on platforms that don't use `/tmp`.

lua/null-ls/builtins/diagnostics/vcom.lua
```diff
@@ -0,0 +1,34 @@
...
+    meta = {
+        url = "https://eda.sw.siemens.com/en-US/ic/modelsim/",
+        description = "Adds support for Mentor Graphics Questa/ModelSim `vcom` VHDL compiler/checker",
+        notes = "If you want the work directory to be in the current working directory, remove '-work' and '/tmp/vcom_work' from the args list",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "vhdl" },
+    generator_opts = {
+        command = "vcom",
+        args = { "-2008", "-quiet", "-lint", "-work", "/tmp/vcom_work", "$FILENAME" },
```
jose-elias-alvarez, line 17: Same comments for above apply here.

jose-elias-alvarez (commented):
This looks good overall. I left a couple of comments, and I also wanted to mention that it looks like there is a [`ghdl language server`](https://github.com/ghdl/ghdl-language-server) and it's supported by [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ghdl_ls). I know nothing about this ecosystem, but if that works, it may be an easier solution, since actual LSP servers will provide a better experience than null-ls.

medwatt:
@jose-elias-alvarez I use mason.nvim, which is supposed to be the successor for nvim-lspconfig. I didn't see any reference to ghdl.

jose-elias-alvarez:
I don't use mason.nvim, but I don't think it's supposed to replace nvim-lspconfig. nvim-lspconfig's role is to configure language servers, while (as I understand it) Mason's role is to manage the installation of language servers and other tools.

Before moving forward with this PR, it would be good to confirm whether the language server provides the same functionality.

medwatt:
> I don't use mason.nvim, but I don't think it's supposed to replace nvim-lspconfig. nvim-lspconfig's role is to configure language servers, while (as I understand it) Mason's role is to manage the installation of language servers and other tools.
> 
> Before moving forward with this PR, it would be good to confirm whether the language server provides the same functionality.

Sorry for that. I think I confused nvim-lspconfig with nvim lsp installer. I don't use nvim-lspconfig directly, so I wouldn't know if the language server provides the same functionality. I user mason to do all the configuration / downloading of the lsp servers.

jose-elias-alvarez:
That's fine then, let me know what you think about the other comments I left. There's also a failing CI step from `stylua`.

PR #1217, opened 2022-10-30
refactor(docs): more consistent setup docs

montecamo:
I think it will be better for setup to be consistent with the rest of the docs, copy/pasting from [BUILTINS.md](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md) will also be easier


jose-elias-alvarez:
Agreed, thanks!

PR #1219, opened 2022-11-01
feat: Update deprecated command line flag

KingMichaelPark:
- The --disable_progress_bar flag was deprecated in place of --disable-progress-bar as per:

  https://docs.sqlfluff.com/en/stable/cli.html#cmdoption-sqlfluff-fix-disable-progress-bar


KingMichaelPark:
seems the commit that changed it was Oct. 3, the newest release of sqlfluff breaks linting because of it. However this update contains the fix I added for https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1123 (which seemingly is now deleted).

jose-elias-alvarez:
That's really strange that the issue link is broken, but if the newer version contains a fix we need, then it seems like a good idea. Could you add a mention of the need to upgrade in the PR description and commit message?

eshaffer321:
Just got hit by this today. Unfortunately you are not able to work around this by providing an extra_args of `--disable-progress-bar`. Only solution is to downgrade

KingMichaelPark:
I've updated the commit message

jose-elias-alvarez:
Thank you!

PR #1230, opened 2022-11-07
feat: Make window border and window highlight of NullLsInfo configurable

### LINKED ISSUE #1197: Make window border and window highlight of NullLsInfo configurable
### Issues

### Feature description

`:LspInfo` has configurable options for the window border (`require(lspconfig.ui.windows).default_options.border`) and also the border color (via `LspInfoBorder` highlight group).

It would be nice if `null-ls.nvim` also supports configuration for the info window.

### Help

Yes

### Implementation help

I don't think it's hard to implement but I don't have much free time to do it currently (I still have a pending PR adding support for `typos` that I haven't looked at yet for a month).

Hopefully someone notices this request and implements it.

guilhas07:
This commit closes #1197 .
The border option was added to the config, and its possible to set it with the same parameters as nvim_open_win() border.  
Added highlight groups:
- `NullLsInfoHeader` Window header
- `NullLsInfoTitle`      Titles
- `NullLsInfoBorder`  Window border
- `NullLsInfoSources` Sources names

Docs were changed to reflect the new config options.

Example of the `border = "rounded"` option:

image

### DIFF
--- a/doc/CONFIG.md
+++ b/doc/CONFIG.md
@@ -63,6 +63,7 @@ local defaults = {
     root_dir = require("null-ls.utils").root_pattern(".null-ls-root", "Makefile", ".git"),
     sources = nil,
     temp_dir = nil,
+    border = nil,
     update_in_insert = false,
 }
 ```
@@ -237,6 +238,13 @@ project for context and so will not work if this option changes.
 You can also configure `temp_dir` per built-in by using the `with` method,
 described in [BUILTIN_CONFIG](BUILTIN_CONFIG.md).
 
+### border ( table, string, optional)
+
+Defines the border to use for the `:NullLsInfo` UI window. Uses
+`NullLsInfoBorder` highlight group (see [Highlight Groups](#highlight-groups)).
+Accepts same border values as `nvim_open_win()`. See `:help nvim_open_win()` for
+more info.
+
 ### update_in_insert (boolean)
 
 Controls whether diagnostic sources run in insert mode. If set to `false`,
@@ -248,6 +256,16 @@ Note that by default, Neovim will not display updated diagnostics in insert
 mode. Together with the option above, you need to pass `update_in_insert = true`
 to `vim.diagnostic.config` for diagnostics to work as expected. See
 `:help vim.diagnostic.config` for more info.
+
+## Highlight Groups
+
+Below are listed the highlight groups that you can override for the
+`:NullLsInfo` window.
+
+- `NullLsInfoHeader` Window header
+- `NullLsInfoTitle` Titles
+- `NullLsInfoBorder` Window border
+- `NullLsInfoSources` Sources names
 
 ## Explicitly defining the project root
 
--- a/lua/null-ls/config.lua
+++ b/lua/null-ls/config.lua
@@ -30,6 +30,7 @@ local type_overrides = {
     should_attach = { "function", "nil" },
     sources = { "table", "nil" },
     temp_dir = { "string", "nil" },
+    border = { "table", "string", "nil" },
 }
 
 local wanted_type = function(k)
@@ -87,6 +88,15 @@ end
 M.setup = function(user_config)
     if config._setup then
         return
+    end
+
+    for group, hi in pairs({
+        NullLsInfoBorder = { link = "NormalFloat", default = true },
+        NullLsInfoHeader = { link = "Label", default = true },
+        NullLsInfoTitle = { link = "Type", default = true },
+        NullLsInfoSources = { link = "Title", default = true },
+    }) do
+        vim.api.nvim_set_hl(0, group, hi)
     end
 
     local validated = validate_config(user_config)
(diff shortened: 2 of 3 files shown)


lua/null-ls/init.lua
```diff
@@ -30,7 +30,7 @@ M.setup = function(user_config)
     c.setup(user_config)
 
     vim.api.nvim_create_user_command("NullLsInfo", function()
-        require("null-ls.info").show_window()
+        require("null-ls.info").show_window(c.get().border)
```
jose-elias-alvarez, line 33: I don't use UI-centric plugins, but is passing in the border as an argument a common pattern? I would expect this method to automatically pick up my configured border unless otherwise specified.

Edit: I suggested a few edits on the method itself that I think will maximize flexibility, but let me know what you think.

lua/null-ls/info.lua
```diff
@@ -160,21 +155,20 @@ M.show_window = function()
...
         local info_lines = { "* Note: current buffer has no sources attached" }
-        table.insert(highlights, { "Type", info_lines[1] })
+        table.insert(highlights, { "NullLsInfoTitle", info_lines[1] })
 
         lines = vim.list_extend(lines, indent_lines({ "" }))
         lines = vim.list_extend(lines, indent_lines(info_lines))
     end
 
-    local win_bufnr, win_id = make_window(0.8, 0.7)
+    local win_bufnr, win_id = make_window(0.8, 0.7, border)
```
jose-elias-alvarez, line 164:
```suggestion
    local win_bufnr, win_id = make_window(0.8, 0.7, opts.border or c.get().border)
```

lua/null-ls/info.lua
```diff
@@ -23,18 +23,13 @@ local make_window = function(height_percentage, width_percentage)
...
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-            { " ", "NormalFloat" },
-        },
+        border = "solid",
```
jose-elias-alvarez, line 26:
```suggestion
        border = border or "solid",
```

lua/null-ls/info.lua
```diff
@@ -23,18 +23,13 @@ local make_window = function(height_percentage, width_percentage)
...
     }
 
+    if border then
+        opts.border = border
+    end
+
```
jose-elias-alvarez, line 32:
```suggestion
```

jose-elias-alvarez (commented):
Looks great! Left a few comments so let me know what you think.

lua/null-ls/init.lua
guilhas07: Hey, you are totally right. I just pushed a commit with your suggestions applied. This was my first contribuition to open source so thanks a lot for the help :smiley: !

doc/CONFIG.md
```diff
@@ -63,6 +63,7 @@ local defaults = {
     root_dir = require("null-ls.utils").root_pattern(".null-ls-root", "Makefile", ".git"),
     sources = nil,
     temp_dir = nil,
+    border = "solid"
```
jose-elias-alvarez, line 66: I'm thinking `border` is a bit too broad as a variable name, since it's not immediately clear what it refers to. Maybe something like `info_window_border` or `info_border`?

jose-elias-alvarez: Also, if we're marking this as optional, I also think this should say `border = nil`. I think that makes the most sense, since defaulting to `border = "solid"` essentially means no border.

lua/null-ls/config.lua
```diff
@@ -30,8 +30,18 @@ local type_overrides = {
     should_attach = { "function", "nil" },
     sources = { "table", "nil" },
     temp_dir = { "string", "nil" },
+    border = { "table", "string", "nil" },
 }
 
+for group, hi in pairs({
```
jose-elias-alvarez, line 36: Can we move this into the `setup` method? Otherwise it's just a side effect from requiring the config file.

doc/CONFIG.md
guilhas07: I have the variable as `border` because of other plugins are using the same nomenclature eg: Mason, Lspconfig, so it should be easy for a new user to know what this variable is doing. Not to mention that exposes the name that  neovim uses when creating the window, which could be a plus too. But if you insist on it I can change it. I will default the border to nil then.

thatvegandev:
lol I spent maybe 30 minutes hacking together essentially this PR before looking to see that one was already open! Nice work @guilhas07, especially for a first contribution!

Side thought - I wish there was a way where plugin authors could hook into to ui preferences of the users. In other words, it there was some way @jose-elias-alvarez could say "I don't care how this is styled, just default to whatever the user prefers"

Issue #1225, opened 2022-11-03
Add ability to root file match to conditional utils


lyonelz96:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

At the moment, when registering a source you can do something like this

```lua
null_ls.builtins.formatting.stylua.with({
        condition = function(utils)
            return utils.root_has_file({ "stylua.toml", ".stylua.toml" })
        end,
    }),
```
From what I understand you can also use the other Conditional Utils in utils.lua (has_file, root_has_file and root_matches).
There is a case I encountered where maybe doing string matching for a file would be a great addition to have.

For example in the case of prettier or eslint.

Prettier can have multiple types of configs: .prettierrc.yaml, .prettierrc.yml, .prettierrc.json, .prettierrc.js, .prettierrc.cjs etc...

You could theoretically use the root_has_file function like this:

```lua
null_ls.builtins.formatting.pretter.with({
        condition = function(utils)
            return utils.root_has_file({ 
               ".prettierrc.yaml", 
               ".prettierrc.yml", 
               ".prettierrc.json" 
            }) -- and so on
        end,
    }),
```

Usually eslint is used in conjunction with prettier and it's the same case there, eslint can have multiple types of config files.

So it would be nice to have something like this

```lua
null_ls.builtins.formatting.pretter.with({
        condition = function(utils)
            return utils.root_has_file_matches(".prettierrc.*")
        end,
    }),
```
Did a little research and came up with something:

Using table

```lua
root_has_file_matches(pattern)
	local user_data = vim.loop.fs_scandir(root)
	local data = vim.loop.fs_scandir_next(user_data)
	local data_tbl = {}

	while data do
		table.insert(data_tbl, data)
		data = vim.loop.fs_scandir_next(user_data)
	end

	for _, name in ipairs(data_tbl) do
		if name:match(pattern) then
			return true
		end
	end

	return false
end
```

Using single string

```lua
root_has_file_matches(pattern)
	local user_data = vim.loop.fs_scandir(root)
	local data = vim.loop.fs_scandir_next(user_data)
	local data_string = ''

	while data do
		data_string = data_string .. ' ' .. data
		data = vim.loop.fs_scandir_next(user_data)
	end

	if data_string:match(pattern) then
		return true
	end

	return false
end
```

Let me know what you think and if you like it I can work on a PR this week.

### Help

[...]

PR #1231, opened 2022-11-08
feat: add ability to root file match to conditional utils

lyonelz96:
Implemented the functionality discussed in #1225

I added some unit tests following the format of the previous conditional utils but for some reason it fails. I printed out the table that I get from this line ```local utils = u.make_conditional_utils()``` in ```utils_spec.lua``` and it doesn't include the ```root_has_file_matches``` function for some reason. I'm not sure what that is about haha.

Edit: Of course it passes in here haha, weird.

### DIFF
--- a/lua/null-ls/utils.lua
+++ b/lua/null-ls/utils.lua
@@ -180,6 +180,7 @@ end
 ---@class ConditionalUtils
 ---@field has_file fun(patterns: ...): boolean checks if file exists
 ---@field root_has_file fun(patterns: ...): boolean checks if file exists at root level
+---@field root_has_file_matches fun(pattern: string): boolean checks if pattern matches a file at root level
 ---@field root_matches fun(pattern: string): boolean checks if root matches pattern
 
 --- creates a table of conditional utils based on the current root directory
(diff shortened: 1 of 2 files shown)


lua/null-ls/utils.lua
```diff
@@ -207,6 +208,24 @@ M.make_conditional_utils = function()
             end
             return false
         end,
+        root_has_file_matches = function(pattern)
+            local user_data = vim.loop.fs_scandir(root)
```
jose-elias-alvarez, line 212: I would simplify this logic a bit and also use clearer variable names:

```lua
local handle = vim.loop.fs_scandir(root)
local entry = vim.loop.fs_scandir_next(handle)

while entry do
    if entry:match(pattern) then
        return true
    end

    entry = vim.loop.fs_scandir_next(handle)
end

return false
```

jose-elias-alvarez (commented):
Not seeing any issues when running tests locally on my end, so this looks good overall.

lua/null-ls/utils.lua
lyonelz96: Oh yeah lol I don't know why I didn't think of doing it in order like that. Awesome! Thank you for the help on that. Fixed on https://github.com/jose-elias-alvarez/null-ls.nvim/commit/fede7b7ded01a447b4d15bb0648ed53441c2bc69

jose-elias-alvarez:
Looks good, thanks!

PR #1232, opened 2022-11-09
Resolve rome executable from node modules

otahontas:
This PR adds local-first resolving order for Rome formatter. I'd argue this should be the default since [Rome docs](https://docs.rome.tools/guides/getting-started/) guide users towards installing Rome to project dependencies.

### DIFF
--- a/lua/null-ls/builtins/formatting/rome.lua
+++ b/lua/null-ls/builtins/formatting/rome.lua
@@ -1,6 +1,7 @@
 local h = require("null-ls.helpers")
 local cmd_resolver = require("null-ls.helpers.command_resolver")
 local methods = require("null-ls.methods")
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks! Do you think it also makes sense to add a default set of root markers, similar to what we are doing for [ESLint](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/9c19fb55154e53dadb986d5aceed1c53e3131b35/lua/null-ls/builtins/formatting/eslint.lua#L48)?

otahontas:
Yep sure, I can add at least rome.json for root markers

otahontas:
@jose-elias-alvarez can we merge this?

jose-elias-alvarez:
Sure, apologies for the delay.

PR #1240, opened 2022-11-15
fix(builtins): keep ansiblelint in sync

sblask:
Ansiblelint does not accept input from stdin, so either the null-ls builtin needs to use `methods.internal.DIAGNOSTICS_ON_SAVE` to run ansiblelint on save or it needs to use `to_temp_file` so the buffer content is written to a temporary file on which ansiblelint can run. Otherwise ansiblelint always runs on the previous modification. Ansiblelint is fast enough so `to_temp_file` works well, but it changes the filename in the diagnostic. In combination with `multiple_files` that means that diagnostics do not show up. `multiple_files` seemed unnecessary anyway so removing it should not cause problems.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/ansiblelint.lua
+++ b/lua/null-ls/builtins/diagnostics/ansiblelint.lua
@@ -1,26 +1,23 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 
-local DIAGNOSTICS = methods.internal.DIAGNOSTICS
-
 return h.make_builtin({
     name = "ansiblelint",
     meta = {
         url = "https://github.com/ansible-community/ansible-lint",
         description = "Linter for Ansible playbooks, roles and collections.",
     },
-    method = DIAGNOSTICS,
+    method = methods.internal.DIAGNOSTICS,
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Given the arguments we're using, does the linter actually generate diagnostics for multiple files at once? If so, I think this could be considered a regression, and I'd prefer to run on save instead. If not, then I think these changes are okay.

sblask:
> Given the arguments we're using, does the linter actually generate diagnostics for multiple files at once? If so, I think this could be considered a regression, and I'd prefer to run on save instead. If not, then I think these changes are okay.

@jose-elias-alvarez I just tried running ansible-lint on https://github.com/sblask/dotfiles/blob/main/ansible/utils-python.yml after having created a problem in an included file: https://github.com/sblask/dotfiles/blob/main/ansible/utils-python.yml#L43 and  ansible-lint reports the problem. Just not sure how useful that is because I have to modify the included file to get an issue reported, so null-ls runs on that file anyway?

jose-elias-alvarez:
I see, I think this change is OK since it's clear from the argument structure that we are only targeting one file. Thanks!

PR #1246, opened 2022-11-18
fix: include code in djlint diagnostics

barrett-ruth:
CLI Output:
image

Before:
image

After:
image


jose-elias-alvarez:
Thank you!

PR #1251, opened 2022-11-21
* fix(builtins): add row_offset and col_offset to clang-format genera…

emailck:
…tor_opts
clang-format require 0-based offset but row and col was 1-based in lua


jose-elias-alvarez:
Thanks!

PR #1253, opened 2022-11-24
feat(dotenv-linter): add diagnostics

### LINKED ISSUE #1252: dotenv-linter
### Issues

### Feature description

Would it be reasonable to add something like [dotenv-linter](https://github.com/dotenv-linter/dotenv-linter) to the null-ls builtins (for formatting & linting `.env` files)?

### Help

Yes

### Implementation help

your [CONTRIBUTING](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONTRIBUTING.md) is awesome. I think I can see a decent roadmap, but I still feel pretty noobish with lua, so I might have dumb questions (e.g. I still don't really get what the deal is with tables. They're dictionaries, AND lists?)

barrett-ruth:
Resolves #1252 .

@jose-elias-alvarez commit should be completely done, but I don't know how to make the source just attach to ".env" files.

Nvim sets filetype "sh" for all files and obviously having the source attach to every shell script buffer (as it currently does) is not desirable. Any tips?

### DIFF
--- a/lua/null-ls/builtins/diagnostics/dotenv_linter.lua
+++ b/lua/null-ls/builtins/diagnostics/dotenv_linter.lua
@@ -1,0 +1,30 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "dotenv-linter",
+    meta = {
+        url = "https://github.com/dotenv-linter/dotenv-linter",
+        description = "Lightning-fast linter for .env files.",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "sh" },
+    generator_opts = {
+        command = "dotenv-linter",
+        args = { "$FILENAME" },
+        from_stderr = false,
+        ignore_stderr = false,
+        format = "line",
+        check_exit_code = function(code)
+            return code <= 1
+        end,
+        to_temp_file = true,
+        runtime_condition = h.cache.by_bufnr(function(params)
+            return params.bufname:find("%.env.*") ~= nil
+        end),
+        on_output = h.diagnostics.from_pattern([[%w+:(%d+) (%w+): (.*)]], { "row", "code", "message" }),
+    },
+    factory = h.generator_factory,
+})


lua/null-ls/builtins/diagnostics/dotenv_linter.lua
```diff
@@ -0,0 +1,31 @@
...
+    name = "dotenv-linter",
+    meta = {
+        url = "https://github.com/dotenv-linter/dotenv-linter",
+        description = "Lightning-fast linter for .env files.",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "sh" },
+    generator_opts = {
+        command = "dotenv-linter",
+        args = { "$FILENAME" },
```
jose-elias-alvarez, line 16: I suspect that with this argument structure, the linter will read the file from the disk, meaning that diagnostics won't update on change until the file is saved. Assuming that's the case, we could try using the [`to_temp_file`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#to_temp_file) option as a workaround.

jose-elias-alvarez (commented):
Looks good overall!

> Nvim sets filetype "sh" for all files and obviously having the source attach to every shell script buffer (as it currently does) is not desirable. Any tips?

We don't have a *great* solution for this right now, but our best option is to use the `runtime_condition` option ([example](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/9caa5d6d19e4d0bc8782c2a75988af0c20eed59e/lua/null-ls/builtins/diagnostics/actionlint.lua#L28)) to make sure diagnostics are only generated for matching buffers.

barrett-ruth:
The only "problem" as you likely know with the `runtime_condition` approach is that null-ls will show up on `:NullLsInfo`. Other than that, fixed.

jose-elias-alvarez:
> The only "problem" as you likely know with the `runtime_condition` approach is that null-ls will show up on `:NullLsInfo`. Other than that, I'll fix this.

Yep, this is an unfortunate consequence of the current approach. The goal for the future (though it'll require a major rewrite) is to have sources attach to buffers based on conditions, but this is the best we have for now. Thanks for the contribution!

PR #1263, opened 2022-12-01
feat(actionlint): Use actionlint config file

klemmster:
Actionlint is not automatically picking up config files when using the "pipe in" version to lint files.

This PR changes the behaviour to honor the config file.

It's my first bit of Lua and Vim development, so please don't hesitate to point our more idiomatic solutions

### DIFF
--- a/lua/null-ls/builtins/diagnostics/actionlint.lua
+++ b/lua/null-ls/builtins/diagnostics/actionlint.lua
@@ -1,4 +1,5 @@
 local h = require("null-ls.helpers")
+local utils = require("null-ls.utils")
 local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/diagnostics/actionlint.lua
```diff
@@ -13,7 +14,22 @@ return h.make_builtin({
     filetypes = { "yaml" },
     generator_opts = {
         command = "actionlint",
-        args = { "-no-color", "-format", "{{json .}}", "-" },
+        args = function()
```
jose-elias-alvarez, line 17: We should use the [by_bufnr](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/9caa5d6d19e4d0bc8782c2a75988af0c20eed59e/lua/null-ls/builtins/diagnostics/actionlint.lua#L28) cache helper to avoid having to run this lookup on every change.

lua/null-ls/builtins/diagnostics/actionlint.lua
```diff
@@ -13,7 +14,22 @@ return h.make_builtin({
...
+            local default_args = { "-no-color", "-format", "{{json .}}" }
+            local c = utils.make_conditional_utils()
```
jose-elias-alvarez, line 19: A potential issue with this approach is that the `.github` directory may not be contained at what null-ls interprets as the root directory. It may be more reliable to search upwards from the current file Here is an alternative approach using the `vim.fs` API (I haven't tested it extensively but it should help):

```lua
local config_file_names = vim.tbl_map(function(ext)
    return require("null-ls.utils").path.join(".github", "actionlint." .. ext)
end, { "yml", "yaml" })

local config_file_path = vim.fs.find(config_file_names, {
    path = params.bufname,
    upward = true,
    stop = vim.fs.dirname(params.root),
})[1]

if config_file_path then
    -- add to arguments
end
```

Alternatively, since we know from the `runtime_condition` that the built-in is only running on files in `.github/workflows`, we could use that to test for the existence of config files.

klemmster:
Thank you so much for the valuable feedback.
I've updated the PR accordingly, and again tested against 
 - no config file
 - either of the extensions

jose-elias-alvarez:
Thanks!

PR #1264, opened 2022-12-01
feat(builtins): add scheme formatter (emacs in batch mode)

delafthi:
Not the nicest way, but I find emacs still to be the best formatter (in terms of the style).

A couple things to point out which aren't that nice:

- I couldn't get emacs to read a `.dir-locals.el` file to set local variables, because it is likely to have 'not safe' variables which need user confirmation. When running in batch mode emacs automatically declines loading a `.dir-locals.el` file
- Thus, when a user wants to adjust some settings, for example use tabs instead of spaces, then he needs to adjust the `args` field and replace the whole elisp code.

Formatting of VHDL in emacs: #1265

### DIFF
--- a/lua/null-ls/builtins/formatting/emacs_scheme_mode.lua
+++ b/lua/null-ls/builtins/formatting/emacs_scheme_mode.lua
@@ -1,0 +1,32 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "emacs/scheme-mode",
+    meta = {
+        url = "https://www.gnu.org/savannah-checkouts/gnu/emacs/emacs.html",
+        description = [[An extensible, customizable, free/libre text editor — and more. Basically, using emacs in batch mode to format scheme files.]],
+        notes = {
+            [[Adjust the expression evaluated with the `--eval` flag to change settings within emacs.]],
+        },
+    },
+    method = FORMATTING,
+    filetypes = { "scheme", "scheme.guile" },
+    generator_opts = {
+        command = "emacs",
+        args = function(params)
+            return {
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/formatting/emacs_scheme_mode.lua
```diff
@@ -0,0 +1,30 @@
...
+        description = [[An extensible, customizable, free/libre text editor — and more. Basically, using emacs in batch mode to format scheme files.]],
+        notes = {
+            [[To change settings within emacs copy the elisp code evaluated with the `--eval` flag and adjust to your likings.]],
+        },
+    },
+    method = FORMATTING,
+    filetypes = { "scheme", "scheme.guile" },
+    generator_opts = {
+        command = "emacs",
+        args = {
```
jose-elias-alvarez, line 19: We can use a function to get the value of `shiftwidth` for the buffer in question:

```lua
args = function(params)
    return {
        "--batch",
        "--eval",
        string.format(
            '(let (scheme-file-content next-line) (while (setq next-line (ignore-errors (read-from-minibuffer ""))) (setq scheme-file-content (concat scheme-file-content next-line "\n"))) (with-temp-buffer (scheme-mode) (setq indent-tabs-mode nil) (setq standard-indent %d) (insert scheme-file-content) (indent-region (point-min) (point-max)) (princ (buffer-string))))',
            vim.bo[params.bufnr].shiftwidth
        ),
    }
end
```

jose-elias-alvarez (commented):
I'm astonished that this works! As long as the caveats are clearly mentioned in the documentation I think this is an okay approach.

delafthi:
Updated the note field and removed manual entry from BUILTINS.md. I didn't see, that it is automatically generated.

jose-elias-alvarez:
Thanks!

PR #1265, opened 2022-12-01
feat(builtins): add VHDL formatter (emacs in batch mode)

delafthi:
Not the nicest way, but I find emacs still to be the best formatter (in terms of the style).

A couple things to point out which aren't that nice:

- I couldn't get emacs to read a .dir-locals.el file to set local variables, because it is likely to have 'not safe' variables which need user confirmation. When running in batch mode emacs automatically declines loading a .dir-locals.el file
- Thus, when a user wants to adjust some settings, for example use tabs instead of spaces, then he needs to adjust the args field and replace the whole elisp code.

Formatting of scheme in emacs: #1264

### DIFF
--- a/lua/null-ls/builtins/formatting/emacs_vhdl_mode.lua
+++ b/lua/null-ls/builtins/formatting/emacs_vhdl_mode.lua
@@ -1,0 +1,32 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "emacs/vhdl-mode",
+    meta = {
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
I think we should use the same approach for `shiftwidth` that I suggested in #1264, but other than that, this also looks good.

delafthi:
> I think we should use the same approach for `shiftwidth` that I suggested in #1264, but other than that, this also looks good.

Same here, I applied your suggestions in #1264 with the newest force-push.

delafthi:
Updated the note field and removed manual entry from BUILTINS.md. I didn't see, that it is automatically generated.

jose-elias-alvarez:
Thanks for the useful (and interesting) contribution!

PR #1271, opened 2022-12-02
feat(health): document, add executable check functions

barrett-ruth:
Here I add executable check functions for girebase and gitsigns, along with some documentation.

Before:
image
After:
image

and when not installed:
image

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -53,7 +53,7 @@
 
 - Make sure your built-in source has a `name`.
 
-- If the source doesn't use `genertor_factory` to spawn an external command,
+- If the source doesn't use `generator_factory` to spawn an external command,
   define a `can_run` field to verify if the plugin is installed. Note that this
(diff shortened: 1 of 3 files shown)


doc/CONTRIBUTING.md
```diff
@@ -54,6 +54,17 @@
 - Make sure your built-in source has a `name`.
 
 - Define a `can_run` field to verify if the plugin is installed (if possible).
+  For example, the gitrebase source relies on git being installed:
```
jose-elias-alvarez, line 57: A couple of suggested changes:

1. We should mention that this is only for `:healthcheck` and does not affect the built-in at runtime.
2. We should also mention that if a built-in source spawns an external command using `generator_factory` (which is the case for most built-ins), it isn't necessary to specify this, since null-ls will already check the executable.

jose-elias-alvarez:
I went ahead and fixed a small typo. Thanks for the work on this!

PR #1278, opened 2022-12-06
fix for builtin golangci linter

eshepelyuk:
Hello

Hello, I am using golangci-lint version 1.50.1 and discovered a bug: `golangci-lint run` doesn't accept any positional arguments, only options.
So current builtins `args` are causing linter to fail because `$DIRECTORY` is used as a positional argument.
I am not sure if this error existed always or this is caused by golangci changed its behaviour at some version.

2nd inconsistency is when linter fails - there weren't any notifications or logs from this builtin, so I've added a warning.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/golangci_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/golangci_lint.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local log = require("null-ls.logger")
 
 local DIAGNOSTICS_ON_SAVE = methods.internal.DIAGNOSTICS_ON_SAVE
 
@@ -21,7 +22,6 @@ return h.make_builtin({
             "--fix=false",
             "--fast",
             "--out-format=json",
-            "$DIRNAME",
             "--path-prefix",
             "$ROOT",
         },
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Hmm, this definitely used to work, since I remember testing this myself. I see that the latest version was released on October 22, which is pretty recent. Could you try to verify if this affects behavior on older versions?

eshepelyuk:
> Hmm, this definitely used to work, since I remember testing this myself. I see that the latest version was released on October 22, which is pretty recent. Could you try to verify if this affects behavior on older versions?

Well, let's imagine I will try older versions and they would be compatible with current null-ls implementation. What will be the resolution ? The bug will still be present for 1.5.x.

jose-elias-alvarez:
I think the fix is fine, but ideally I'd like to make sure that the change doesn't affect older versions (meaning that leaving out the positional argument doesn't affect functionality).

jose-elias-alvarez:
Our policy so far has been to only support the latest version of a given CLI tool, so I'd still merge it. If it breaks older versions completely, I'd likely mention it in our pinned breaking changes issue.

eshepelyuk:
> Our policy so far has been to only support the latest version of a given CLI tool, so I'd still merge it. If it breaks older versions completely, I'd likely mention it in our pinned breaking changes issue.

Fair enough,thanks for explanation.
Will try to report the results during next upcoming days.

PR #1280, opened 2022-12-06
Add autoflake formatter

alexander-born:
add autoflake formatter

### DIFF
--- a/lua/null-ls/builtins/formatting/autoflake.lua
+++ b/lua/null-ls/builtins/formatting/autoflake.lua
@@ -1,0 +1,20 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "autoflake",
+    meta = {
+        url = "https://github.com/PyCQA/autoflake",
+        description = "Removes unused imports and unused variables as reported by pyflakes",
+    },
+    method = { FORMATTING },
+    filetypes = { "python" },
+    generator_opts = {
+        command = "autoflake",
+        args = { "--stdin-display-name", "$FILENAME", "-" },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lua/null-ls/builtins/formatting/autoflake.lua
```diff
@@ -0,0 +1,21 @@
...
+    name = "autoflake",
+    meta = {
+        url = "https://github.com/PyCQA/autoflake",
+        description = "Removes unused imports and unused variables as reported by pyflakes",
+    },
+    method = { FORMATTING, RANGE_FORMATTING },
+    filetypes = { "python" },
+    generator_opts = {
+        command = "autoflake",
+        args = { "--stdout", "$FILENAME" },
```
jose-elias-alvarez, line 17: Does the formatter read from `stdin` when a filename is specified? It looks like there's a `--stdin-display-name` flag that may be more appropriate.

alexander-born: Yeah it didn't properly test. The current pr does not work yet. I will fix it. Thanks for the hint.

jose-elias-alvarez: Hmm, that's usually what flags like `--stdin-display-name` are supposed to do: tell the CLI where the actual file is located so that it can determine config file locations. 

If you can replicate this behavior directly from the CLI (i.e. not via null-ls) I'd open an issue upstream. Otherwise we will need to do additional work to find the config file ourselves, but it theoretically shouldn't be necessary.

alexander-born: Yes seems to be a bug in upstream...
The config is only searched for actual files. It does not consider the `stdin_display_name`, see https://github.com/PyCQA/autoflake/blob/0373337022756fd930f01e6549a1b0a224e60376/autoflake.py#L1162

alexander-born: And passing the config file explicitly via `--config` expects a .cfg file not a project.toml, see https://github.com/PyCQA/autoflake/blob/0373337022756fd930f01e6549a1b0a224e60376/autoflake.py#L1196

So both problems are upstream bugs... I will report and fix them there.

alexander-born: I was wrong regarding the config being ignored, when using stdin_display_name, it does in fact work, I tested it wrongly.

lua/null-ls/builtins/formatting/autoflake.lua
```diff
@@ -0,0 +1,21 @@
...
+local FORMATTING = methods.internal.FORMATTING
+local RANGE_FORMATTING = methods.internal.RANGE_FORMATTING
+
+return h.make_builtin({
+    name = "autoflake",
+    meta = {
+        url = "https://github.com/PyCQA/autoflake",
+        description = "Removes unused imports and unused variables as reported by pyflakes",
+    },
+    method = { FORMATTING, RANGE_FORMATTING },
```
jose-elias-alvarez, line 13: One last thing: if we want this source to handle range formatting, then we need to handle this in the arguments (see [autopep8](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/autopep8.lua) for an example). If it does support range formatting, we can either add support now or remove it for now and save it for a follow-up.

alexander-born: I removed `RANGE_FORMATTING`, autoflake has no command line arguments to handle ranges like for example autopep8.

jose-elias-alvarez:
Thanks!

PR #1282, opened 2022-12-07
fix(builtins)!: modify editorconfig-checker to run using standard tool name

jay-babu:
Hello,

There have been a few discussions around the name of the command that runs when using this source.

- https://github.com/williamboman/mason.nvim/pull/722
- https://github.com/jay-babu/mason-null-ls.nvim/issues/37

This tool currently runs a command call `ec`, which is not the default name of the tool when installed.

https://github.com/editorconfig-checker/editorconfig-checker

after installing using one of the methods described in the repo, I get a tool named: `editorconfig-checker` unless I do some special tinkering to make it work with the `null-ls` wants it. This change now adheres to the standard name of the tool.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/editorconfig_checker.lua
+++ b/lua/null-ls/builtins/diagnostics/editorconfig_checker.lua
@@ -12,7 +12,7 @@ return h.make_builtin({
     method = DIAGNOSTICS,
     filetypes = {},
     generator_opts = {
-        command = "ec",
+        command = "editorconfig-checker",
         args = {
             "-no-color",
             "$FILENAME",


jose-elias-alvarez:
Hmm, it looks like there's an inconsistency on their end. Looking at [the installation options](https://github.com/editorconfig-checker/editorconfig-checker#installation), this command installs an `editorconfig-checker` binary into my Go binary directory:

```sh
go install github.com/editorconfig-checker/editorconfig-checker/cmd/editorconfig-checker@latest
```

The binaries from [the release page](https://github.com/editorconfig-checker/editorconfig-checker/releases), on the other hand, don't seem to have consistent naming at all. Each one's name includes the specific architecture it was built for.

Finally, I wasn't actually able to get the `go get` method to work, but the documentation says:

> This will place a binary called `ec` into the `bin` directory.

I imagine most users would use the `go install` method, but I think it would be a good idea to get some clarity on what the "standard" binary name is. [This issue](https://github.com/editorconfig-checker/editorconfig-checker/issues/208) seems to suggest that `ec` is actually correct.

jay-babu:
https://github.com/editorconfig-checker/editorconfig-checker/issues/232

the answer based on this discussion should resolve this discussion

jay-babu:
https://github.com/editorconfig-checker/editorconfig-checker/issues/232

This change can be merged as the owner agreed `editorconfig-checker` will be the official name.

jay-babu:
@jose-elias-alvarez, if you have some time to take a look

jose-elias-alvarez:
Thanks for verifying, I'll mention this in the breaking changes issue.

PR #1283, opened 2022-12-07
feat: add prettier-eslint foramter support

legends-killer:
add prettier-eslint formatter for users need to use eslint and prettier at the same time.

### DIFF
--- a/lua/null-ls/builtins/formatting/prettier_eslint.lua
+++ b/lua/null-ls/builtins/formatting/prettier_eslint.lua
@@ -1,0 +1,68 @@
+local h = require("null-ls.helpers")
+local cmd_resolver = require("null-ls.helpers.command_resolver")
+local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "prettier_eslint",
+    meta = {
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/formatting/prettier_eslint.lua
```diff
@@ -0,0 +1,68 @@
...
+        "jsonc",
+        "yaml",
+        "markdown",
+        "markdown.mdx",
+        "graphql",
+        "handlebars",
+    },
+    generator_opts = {
+        command = "prettier-eslint",
+        args = { "$FILENAME" },
```
jose-elias-alvarez, line 38: According to [the documentation](https://github.com/prettier/prettier-eslint-cli#cli-options), it looks like we should use the `--stdin` and `--stdin-filename` flags here.

legends-killer: Yeah, I removed these two flags because it seems that the --stdin and --std-filepath flags don't work properly in prettier-eslint-cli, it always gets nothing from stdin, I'll fix it until I figure out this.
But under the current circumstance, it works well.

legends-killer: OK, it looks like I‘ve been debugging it in a wrong way. In fact, it works well.

jose-elias-alvarez:
Thanks!

PR #1289, opened 2022-12-10
fix: set buflisted to false after apply_text_edits

LostNeophyte:
[apply_text_edits is setting buflisted to true](https://github.com/neovim/neovim/blob/8b84a10db76ef2bd15bbd3c06ae2d5dfaadc1482/runtime/lua/vim/lsp/util.lua#L404), which causes the temporary buffer to be visible in bufferline.nvim for a split second


jose-elias-alvarez:
LGTM, thanks!

PR #1293, opened 2022-12-13
feat: add `swiftlint` builtin diagnostics and formatting

Ty3uK:
This PR adds support for swiftlint as diagnostics and formatting source.

Sorry for poor description, not sure how to properly describe this tiny feature 🙈

https://user-images.githubusercontent.com/2384454/207402202-e0f19862-bf65-4aab-97bb-67c857332549.mp4

In this video I use both `swiftlint` and `swiftformat` sources

### DIFF
--- a/lua/null-ls/builtins/diagnostics/swiftlint.lua
+++ b/lua/null-ls/builtins/diagnostics/swiftlint.lua
@@ -1,0 +1,44 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local handle_swiftlint_output = function(params)
+    local parser = h.diagnostics.from_json({
+        attributes = {
+            severity = "severity",
+            col = "character",
+            code = "rule_id",
+            message = "reason",
+        },
+        severities = {
+            ["warning"] = "Warning",
+            ["error"] = "Error",
+        },
+    })
+
+    return parser({ output = params.output })
+end
+
+return h.make_builtin({
+    name = "swiftlint",
+    meta = {
+        url = "https://github.com/realm/SwiftLint",
+        description = "A tool to enforce Swift style and conventions.",
+    },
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/swiftlint.lua
```diff
@@ -0,0 +1,44 @@
...
+        url = "https://github.com/realm/SwiftLint",
+        description = "A tool to enforce Swift style and conventions.",
+    },
+    method = DIAGNOSTICS,
+    filetypes = { "swift" },
+    generator_opts = {
+        command = "swiftlint",
+        args = { "--reporter", "json", "--use-stdin", "--quiet" },
+        to_stdin = true,
+        format = "json_raw",
```
jose-elias-alvarez, line 36: We generally only want to use `json_raw` if you are specifically handling errors, which doesn't seem to be the case. Was there a specific reason you chose this over `json`?

lua/null-ls/builtins/diagnostics/swiftlint.lua
```diff
@@ -0,0 +1,44 @@
...
+        use_cache = true,
```
jose-elias-alvarez, line 37: This option only has an effect if you have multiple sources using the same output, so I don't think it's necessary here.

lua/null-ls/builtins/diagnostics/swiftlint.lua
Ty3uK: No reason at all, sorry.
It's just copy-pasted from `eslint` source 🙈

lua/null-ls/builtins/diagnostics/swiftlint.lua
Ty3uK: By "this" you mean `use_cache`?

jose-elias-alvarez: Yes, that's what I meant (GitHub unfortunately doesn't make this totally clear).

jose-elias-alvarez:
Thanks!

PR #1294, opened 2022-12-14
feat(builtins): add gomodifytags support in code actions

gh-liu:


jose-elias-alvarez:
Thanks! Let me know when you think this is fully ready to review and if there's a way for me to test this locally, since it's more complicated than the typical source and I'd like to know what expected behavior looks like in case I have to maintain it.

gh-liu:
> Thanks! Let me know when you think this is fully ready to review and if there's a way for me to test this locally, since it's more complicated than the typical source and I'd like to know what expected behavior looks like in case I have to maintain it.

1. Install `gomodifytags`
```bash
go install github.com/fatih/gomodifytags@latest
```
2. New a go file which contains a struct
```go
package main

type Server struct {
	Name        string
	Port        int
	EnableLogs  bool
	BaseDomain  string
	Credentials struct {
		Username string
		Password string
	}
}
```
3. Move the cursor to the struct name or the field name and run code actions, input [the tags or options](https://github.com/fatih/gomodifytags#adding-tags--options).

gomodifytags

jose-elias-alvarez:
Thanks! I'll give this a shot this weekend and let you know if I have any feedback.

lua/null-ls/builtins/code_actions/gomodifytags.lua
```diff
@@ -0,0 +1,158 @@
...
+    local output = vim.fn.system(cmd)
+    if vim.api.nvim_get_vvar("shell_error") == 0 then
+        vim.fn.execute(":e " .. bufname) -- reload the file
+    else
+        log:error(output) -- print error message
+    end
+end
+
+local get_input_tag_and_exec = function(bufname, struct_name, field_name, op)
+    vim.ui.input({ prompt = "Enter tags: " }, function(tag)
```
jose-elias-alvarez, line 32: I'm not familiar with Go, so I don't really know what this tool actually does, but: does it support adding multiple tags at once?

lua/null-ls/builtins/code_actions/gomodifytags.lua
```diff
@@ -0,0 +1,158 @@
...
+            exec(bufname, struct_name, field_name, "-clear-options", nil)
+        end,
+    }
+end
+
+return h.make_builtin({
+    name = "gomodifytags",
+    meta = {
+        url = "https://github.com/fatih/gomodifytags",
+        description = "Go tool to modify struct field tags",
```
jose-elias-alvarez, line 98: I think it'd be good to mention in the `notes` field (see the example [here](https://github.com/gh-liu/null-ls.nvim/blob/d2e7a24698fcc160d2d7c0ca4b3f0a4e37679e93/doc/CONTRIBUTING.md#L81)) that this requires installing the Go tree-sitter parser.

jose-elias-alvarez (commented):
I can't say too much since I'm not familiar with the tool / Go in general, but it seems to work well! 

One small edge case I ran into: if the user edits the buffer but doesn't save it, then applies a code action, the buffer and the file are in an inconsistent state. It seems like a good idea to handle this, either by warning the user if the buffer is modified or just saving any changes.

lua/null-ls/builtins/code_actions/gomodifytags.lua
gh-liu: Yes, input `xml,json` will add xml and json tag at once.

lua/null-ls/builtins/code_actions/gomodifytags.lua
gh-liu: Sorry, I fotgot this, I will add it.

gh-liu:
> It seems like a good idea to handle this, either by warning the user if the buffer is modified or just saving any changes.

I prefer the second way, just saving any changes.

jose-elias-alvarez:
Thanks for the great contribution! I think there's a lot of options for tree-sitter-based code action sources, so I hope this can serve as an example to other contributors, too.

gh-liu:
> Thanks for the great contribution! I think there's a lot of options for tree-sitter-based code action sources, so I hope this can serve as an example to other contributors, too.
Also thank you for writing null-ls.nvim to make a lot of options to achieve easyily.

gh-liu:
> Thanks for the great contribution! I think there's a lot of options for tree-sitter-based code action sources, so I hope this can serve as an example to other contributors, too.

Also thank you for writing null-ls.nvim to make a lot of options to achieve easyily.

Issue #1255, opened 2022-11-24
Add a terraform validate builtin


sblask:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

You can run `terraform validate` to find problems with terraform code, it's complementing tflint. I am currently testing it from my config:

```lua
local terraform_validate = {
    name = "terraform_validate",
    method = null_ls.methods.DIAGNOSTICS_ON_SAVE,
    filetypes = { "terraform" },
    generator = null_ls.generator({
        command = "terraform",
        args = {
            "validate",
            "-json",
        },
        cwd = function(params)
            return vim.fs.dirname(params.bufname)
        end,
        to_stdin = false,
        from_stderr = true,
        multiple_files = true,
        format = "json",
        check_exit_code = function(_code, _stderr)
            -- check for warnings even if there are no errors
            return false
        end,
        on_output = function(params)
            local working_directory = vim.fs.dirname(params.bufname)
            local rewritten_diagnostics = {}
            for _, diagnostic in ipairs(params.output.diagnostics) do
                local rewritten_diagnostic = {
                    message = diagnostic.detail,
                    row = 0,
                    col = 0,
                    source = "terraform validate",
                    severity = helpers.diagnostics.severities[diagnostic.severity],
                    filename = params.bufname,
                }
                if diagnostic.range ~= nil then
                    rewritten_diagnostic.col = diagnostic.range.start.column
                    rewritten_diagnostic.end_col = diagnostic.range["end"].column
                    rewritten_diagnostic.row = diagnostic.range.start.line
                    rewritten_diagnostic.end_row = diagnostic.range["end"].line
                    rewritten_diagnostic.filename = working_directory .. "/" .. diagnostic.range.filename
                end
                table.insert(rewritten_diagnostics, rewritten_diagnostic)
            end
            return rewritten_diagnostics
        end,
    }),
}
null_ls.register(terraform_validate)
```

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I had a look at https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/test/spec/builtins/diagnostics_spec.lua but it's a bit unclear to me how it works

jose-elias-alvarez:
[...]

PR #1302, opened 2022-12-17
feat(builtins): add terraform validate builtin (#1255)

sblask:


lua/null-ls/builtins/diagnostics/terraform_validate.lua
```diff
@@ -0,0 +1,65 @@
...
+    },
+    method = DIAGNOSTICS_ON_SAVE,
+    filetypes = { "terraform" },
+    generator_opts = {
+        command = "terraform",
+        args = {
+            "validate",
+            "-json",
+        },
+        cwd = function(params)
```
jose-elias-alvarez, line 20: Since this source only runs on save, I don't think it's a big deal performance-wise, but we could consider caching this to avoid repeated lookups if we don't expect changes ([example](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/5d8e925d31d8ef8462832308c016ac4ace17597a/lua/null-ls/builtins/formatting/eslint.lua#L47)).

lua/null-ls/builtins/diagnostics/terraform_validate.lua
```diff
@@ -0,0 +1,65 @@
...
+            return vim.fs.dirname(params.bufname)
+        end,
+        from_stderr = true,
+        to_stdin = false,
+        multiple_files = true,
+        format = "json",
+        check_exit_code = function(_code, _stderr)
+            -- check for warnings even if there are no errors
+            return false
```
jose-elias-alvarez, line 29: The basic effect of combining this with `from_stderr` is to treat any output as error output, then treat that output as standard output. Are you sure there's no other way to handle this? What happens if we pass in unparsable files, for example?

lua/null-ls/builtins/diagnostics/terraform_validate.lua
```diff
@@ -0,0 +1,65 @@
...
+        end,
+        on_output = function(params)
+            local working_directory = vim.fs.dirname(params.bufname)
```
jose-elias-alvarez, line 32: Small optimization: since we already resolved this when spawning the command, we should be able to use `params.cwd` instead.

lua/null-ls/builtins/diagnostics/terraform_validate.lua
```diff
@@ -0,0 +1,65 @@
...
+                    filename = params.bufname,
+                }
+                if diagnostic.range ~= nil then
+                    rewritten_diagnostic.col = diagnostic.range.start.column
+                    rewritten_diagnostic.end_col = diagnostic.range["end"].column
+                    rewritten_diagnostic.row = diagnostic.range.start.line
+                    rewritten_diagnostic.end_row = diagnostic.range["end"].line
+                    -- vim.fs.dirname not working in tests
+                    if working_directory ~= nil then
+                        rewritten_diagnostic.filename = working_directory .. "/" .. diagnostic.range.filename
```
jose-elias-alvarez, line 54: Nit: for consistency let's use `utils.path.join` for this.

jose-elias-alvarez (commented):
Thanks! I appreciate the test coverage. I left a few comments as well.

lua/null-ls/builtins/diagnostics/terraform_validate.lua
sblask: Done.

lua/null-ls/builtins/diagnostics/terraform_validate.lua
sblask: Nice!

lua/null-ls/builtins/diagnostics/terraform_validate.lua
sblask: That allowed me to remove the if/else!

lua/null-ls/builtins/diagnostics/terraform_validate.lua
sblask: There is no command line flag to treat warnings as errors if that's what you mean. But at least I consistently get json back. It would be nice to not have to parse it on every save, but I am not sure how.

sblask:
There is one thing that is not quite working as expected: assuming I have files open from two different directories using `-O` say `foo/variables.tf` , `foo/main.tf`  and `bar/main.tf`. Let's say I have problems in all files, only problems from one directory are visible at a time. If I switch to `foo/main.tf` I see the problems from all files under foo and I no longer see the problems for the file under bar and vice versa. What I would want is that I always see all the problems and adding/fixing a problem only affects findings for files in the same directory and should leave the findings in other directories untouched. I am not sure this comes from what I have done or whether the `multiple_files` option needs an additional scope for this to work?

jose-elias-alvarez:
I think this is expected, right? If you set `cwd` to `vim.fs.dirname(params.bufname)` then you will spawn the linter in that directory, so you'll only see diagnostics for files in that directory. Is there some root marker you can use to spawn the linter at the top level so that it captures diagnostics for all files?

sblask:
> I think this is expected, right? If you set `cwd` to `vim.fs.dirname(params.bufname)` then you will spawn the linter in that directory, so you'll only see diagnostics for files in that directory.

Maybe I misunderstood how this works, I thought it's one null-ls "instance" per buffer that runs in this case `terraform validate`. By default this updates the diagnostics for just that buffer/file. With `multiple_files` it should also update the diagnostics for all the buffers for files that match the `filename`s in the generated diagnostics (https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1302/files#diff-ddb6f45671840e942bb4e7c18000d501a5db5d291982e7a5d0245a652d7fa7ffR52). Buffers that don't match `filename` should not get their diagnostics updated? Unless a diagnostic got fixed.

> Is there some root marker you can use to spawn the linter at the top level so that it captures diagnostics for all files?

No, terraform is a bit of a mess ;-)

jose-elias-alvarez:
Diagnostics are namespaced to specific sources, so when a multi-file source returns diagnostics, it'll first wipe out any existing diagnostics it previously created. This is because multi-file sources are intended to run across entire projects, and fixing an issue in file `a` may create (or fix) an issue in file `b`.

I'm not sure we have a good solution for this situation, actually. If we run `terraform validate` at a higher level in the tree (e.g. by using a `.git` folder as a root marker) what happens?

lua/null-ls/builtins/diagnostics/terraform_validate.lua
jose-elias-alvarez: As long as the linter always outputs valid JSON then I think this is fine.

sblask:
> Diagnostics are namespaced to specific sources, so when a multi-file source returns diagnostics, it'll first wipe out any existing diagnostics it previously created. This is because multi-file sources are intended to run across entire projects, and fixing an issue in file `a` may create (or fix) an issue in file `b`.

I guess in terraform's case only diagnostics in files in the same directory should be wiped. But I am not sure this can be generalised in any way. So maybe this should just be merged and if a similar use case pops up with another tool a possible solution might become clearer as well.

> I'm not sure we have a good solution for this situation, actually. If we run `terraform validate` at a higher level in the tree (e.g. by using a `.git` folder as a root marker) what happens?

That will give an error, you have to be in a directory with terraform files.

I guess in terraform's case only diagnostics in files in the same directory should be wiped

sblask:
> One idea: maybe we can add an option for multi-file sources to disable the default behavior of wiping out all diagnostics and instead let the source control it. That way, we could get all diagnostics in the source's namespace but only clear the ones matching the current dirname. Does that sound like it would solve the issue? It shouldn't be too hard to set up if so.

You mean the built-in would need to handle the wiping? Not sure how that would work, but that should solve the problem.

jose-elias-alvarez:
Yeah, it's possible but the ergonomics aren't great. Now that I think about it, you should be able to work around this by simply keeping old diagnostics from files outside of the current working directory (untested but hopefully illustrative):

```lua
on_output = function(params)
    local source_id = params.source_id
    local namespace = require("null-ls.diagnostics").get_namespace(source_id)
    local old_diagnostics = vim.diagnostic.get(nil, { namespace = namespace })

    local diagnostics = {}
    for _, old_diagnostic in ipairs(old_diagnostics) do
        if not vim.startswith(old_diagnostic.filename, params.cwd) then
            -- keep diagnostics from other directories
            table.insert(diagnostics, old_diagnostic)
        end
    end
    -- insert new diagnostics into the same diagnostics table, then return
end
```

Let me know if that makes sense. Otherwise we can merge this as-is and wait until we see another source with the same issue so we can come up with a more general solution.

sblask:
@jose-elias-alvarez I implemented your suggestion, tested it for a while and it seems to work fine! Thanks!

jose-elias-alvarez:
Cool, thanks for the work on this!

PR #1306, opened 2022-12-22
fix(builtins): cspell parse multiline jsons

marianozunino:
My `lua` skills are as good as my tailoring...

Thanks to @peterfication for this comment: 
https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1154#issuecomment-1362590225

### DIFF
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -43,7 +43,7 @@ local create_cspell_json = function(cwd)
         flagWords = {},
     }
     local cspell_json_str = vim.json.encode(cspell_json)
-    local cspell_json_file_path = (cwd or vim.loop.cwd()) .. "/cspell.json"
+    local cspell_json_file_path = require("null-ls.utils").path.join(cwd or vim.loop.cwd(), "cspell.json")
     vim.fn.writefile({ cspell_json_str }, cspell_json_file_path)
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks! While we're fixing this, could we also fix [this check](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/158d5817da817c81123d24c2c392e0f517ed532b/lua/null-ls/builtins/code_actions/cspell.lua#L92)? We should check if `cspell_json_file == nil` and not an empty string. I also think it's more appropriate as a warning.

marianozunino:
> Thanks! While we're fixing this, could we also fix [this check](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/158d5817da817c81123d24c2c392e0f517ed532b/lua/null-ls/builtins/code_actions/cspell.lua#L92)? We should check if `cspell_json_file == nil` and not an empty string. I also think it's more appropriate as a warning.

You bet.

Also I added the `create_cspell_json` function which, well, creates the file if is not present. Also, this behaviour may be prevented using the config table.

```lua
		config = {
			create_config_file = false, -- defaults to true
		},
	}),
```

jose-elias-alvarez:
Pushed a couple of small changes but this looks good and seems to be working well on my end. Thanks for working on this!

peterfication:
Thanks a lot for incorporating the quick fix :)

PumpedSardines:
Thank you all so much! Nice to see this fix being merged!!!

PR #1310, opened 2022-12-25
Option to change the default cspell json file name

PumpedSardines:
Currently when choosing the option "Add to cspell json file" and no json file exists, a file of the name `cspell.json` will be created. Since I always use `.cspell.json` I would want a way to always use that name instead.

### DIFF
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -83,8 +83,8 @@ return h.make_builtin({
             if not is_cspell_config_file_name(create_cspell_json) then
                 vim.notify(
                     "Invalid default file name for cspell json file: "
-                    .. create_config_file_name
-                    .. '. The name "cpsell.json" will be used instead',
+                        .. create_config_file_name
+                        .. '. The name "cpsell.json" will be used instead',
                     vim.log.levels.WARN
                 )
                 create_config_file_name = "cspell.json"
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -34,8 +34,17 @@ local find_cspell_config = function(cwd)
     return cspell_json_file
 end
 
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -34,16 +34,25 @@ local find_cspell_config = function(cwd)
     return cspell_json_file
 end
 
+local is_cspell_config_file_name = function(file_name)
```
jose-elias-alvarez, line 37: We can use `vim.tbl_contains()` here instead.

lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -64,13 +73,23 @@ return h.make_builtin({
...
 
             -- create_config_file if nil defaults to true
             local create_config_file = config.create_config_file ~= false
 
+            local create_config_file_name = config.create_config_file_name or "cspell.json"
+            if not is_cspell_config_file_name(create_cspell_json) then
+                vim.notify(
+                    "Invalid default file name for cspell json file: "
+                        .. create_config_file_name
+                        .. '. The name "cpsell.json" will be used instead',
```
jose-elias-alvarez, line 87:
```suggestion
                        .. '. The name "cspell.json" will be used instead',
```

jose-elias-alvarez, line 87:
```suggestion
                        .. '. The name "cspell.json" will be used instead',
```

lua/null-ls/builtins/code_actions/cspell.lua
PumpedSardines: Oh nice!

jose-elias-alvarez:
Looks good, thanks! I noticed that we don't yet have documentation for these options but we can save that for a follow-up.

PumpedSardines:
I would be happy to add that! I'll make a PR for it 🤩

PR #1313, opened 2022-12-27
feat(deno_fmt): respect textwidth and shiftwidth

jceb:


jose-elias-alvarez:
Seems like a good default, thanks!

anasrar (commented):
this make deno ignore format setting on  [configuration file](https://deno.land/manual@v1.29.1/getting_started/configuration_file#configuration-file)

jose-elias-alvarez:
> this make deno ignore format setting on [configuration file](https://deno.land/manual@v1.29.1/getting_started/configuration_file#configuration-file)

So the CLI settings will always override the config file? If so I think we should revert this.

jceb:
> this make deno ignore format setting on [configuration file](https://deno.land/manual@v1.29.1/getting_started/configuration_file#configuration-file)

I'd say if the configuration file has different settings than the editor deno and the will always clash. The way to fix it is to provide the exact same settings to deno and to the editor, e.g. automatically via [editorconfig](https://editorconfig.org/).

anasrar:
why not just check the root project if has `deno.json` or `deno.jsonc`, if not present just use current editor setting otherwise use `deno.json` (the default behavior `deno fmt`)

jceb:
Considering that there are other text editors out there that can't rely on a custom null-ls behavior to detect `deno.json` they'll just rely on editor settings to  format code. This leads me back to ensuring that the editor is configured properly.

PR #1318, opened 2022-12-28
fix(opacheck): improve error notifications

eshepelyuk:


eshepelyuk:
Hello @jose-elias-alvarez 

Can you plz help in fixing the test ?
I want to use null-ls logger in diagnostics parser, but as soon as I did it - the tests started to fail.

Although logger is used in some other diagnostic builtins, it looks like there is no tests for them, so it never happens before.

lua/null-ls/builtins/diagnostics/opacheck.lua
```diff
@@ -17,6 +18,8 @@ local handle_opacheck_output = function(params)
                 filename = d.location.file,
                 code = d.code,
             })
+        elseif idx < 5 then
```
jose-elias-alvarez, line 21: I think it would be good to add a comment here about why this is necessary, for future reference.

jose-elias-alvarez:
> Hello @jose-elias-alvarez
> 
> Can you plz help in fixing the test ? I want to use null-ls logger in diagnostics parser, but as soon as I did it - the tests started to fail.
> 
> Although logger is used in some other diagnostic builtins, it looks like there is no tests for them, so it never happens before.

Not sure why this would cause failures, but does adding this line to the top of the file help?

```lua
require("luassert.mock")(require("null-ls.logger"), true)
```

lua/null-ls/builtins/diagnostics/opacheck.lua
eshepelyuk: Added

eshepelyuk:
> > Hello @jose-elias-alvarez
> > Can you plz help in fixing the test ? I want to use null-ls logger in diagnostics parser, but as soon as I did it - the tests started to fail.
> > Although logger is used in some other diagnostic builtins, it looks like there is no tests for them, so it never happens before.
> 
> Does adding this line to the top of the file help?
> 
> ```lua
> require("luassert.mock")(require("null-ls.logger"), true)
> ```

Thanks @jose-elias-alvarez - it is fixed.

jose-elias-alvarez:
Thanks!

PR #1329, opened 2023-01-06
Option to run cspell diagnostics from a specific cspell config

juanpprieto:


lua/null-ls/builtins/diagnostics/cspell.lua
```diff
@@ -32,8 +32,18 @@ return h.make_builtin({
...
+            local find_json = config.find_json or h.cspell.find_cspell_config
+            local cspell_json_file = find_json(params.cwd)
+
+            if cspell_json_file == nil or cspell_json_file == "" then
+                cspell_json_file = "cspell.json" -- default for the -c option
+            end
+
             local cspell_args = {
                 "lint",
+                "-c",
```
juanpprieto, line 45: See `-c` flag here 
https://cspell.org/docs/getting-started/#options

lua/null-ls/helpers/cspell.lua
```diff
@@ -0,0 +1,25 @@
+local CSPELL_CONFIG_FILES = {
```
juanpprieto, line 1: This feels like it might be the wrong place for this logic as it's not a global null-ls helper.

I just couldn't find a more suitable folder. Maybe add a `common` or `shared` folder within builtins?

Up to you

lua/null-ls/builtins/diagnostics/cspell.lua
jose-elias-alvarez: As I understand it, `cspell` can run without a config file. What will happen when this flag is used but the config file doesn't exist?

lua/null-ls/helpers/cspell.lua
jose-elias-alvarez: Yes, I agree that if we're going to go with this approach we should put this elsewhere.

jose-elias-alvarez (commented):
I'm not sure about this approach - I agree that there's an inconsistency between the diagnostics source and the code action source, but I think this is a lot of complexity to add for something that's already solved by `extra_args`. There's also potential performance issues - running `findfile()` each time the user requests code actions is okay, but running it on every change (which is what happens for diagnostics sources) could cause serious slowdown.

Since these sources are so closely linked, perhaps there's a better solution where one source can derive config values from the other.

jose-elias-alvarez:
FYI: I wonder if the approach used in #1339 could also help solve some of the inconsistencies between the code action source and the diagnostics source.

PR #1331, opened 2023-01-08
Change the default argument of the `yamlfmt` formatter

paveloom:
Today I encountered `yamlfmt` replacing the contents of my files with a single hyphen, and then appending new hyphens on new lines with each save. Perhaps, this has something to do with the [new version](https://github.com/google/yamlfmt/releases/tag/v0.7.0). Anyhow, here's a small fix that works for me.

### DIFF
--- a/lua/null-ls/builtins/formatting/yamlfmt.lua
+++ b/lua/null-ls/builtins/formatting/yamlfmt.lua
@@ -14,7 +14,7 @@ return h.make_builtin({
     generator_opts = {
         command = "yamlfmt",
         to_stdin = true,
-        args = { "-" },
+        args = { "-in" },
     },
     factory = h.formatter_factory,
 })


jose-elias-alvarez:
Hmm, I'm not seeing anything in the changelog or documentation related to this. It looks like the `-in` and `-` are supposed to have the same behavior.

Could you try an older version of the formatter and verify? If this change is necessary, then I'd say that this either needs to be documented upstream or could be considered a regression.

paveloom:
Yep, works fine on `v0.6.0`.

It might be intentional. Take a look at this pesky hyphen there:

image

paveloom:
Aight, the issue is solved upstream.

jose-elias-alvarez:
Thanks for figuring this out!

PR #1343, opened 2023-01-14
feat: include snip.id in LuaSnip

oncomouse:
LuaSnip expansion of completed snippets is difficult with this module because the completed item does not contain snip.id. With this PR, snip.id is included. Now snippets can be expanded using the same code as [cmp_luasnip](https://github.com/saadparwaiz1/cmp_luasnip/), which works much better.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -20,7 +20,7 @@ jobs:
 
   selene:
     name: Lint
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Apologies - I did something dumb with Git and closed this PR, but I merged the commit in #1350. Thanks!

PR #1344, opened 2023-01-14
Remove --fast default from golangci-lint

BlakeWilliams:
First, thanks for null-ls! I've found it to be super helpful.

I spent a good amount of time trying if/why debug if null-ls wasn't working and why I would see no errors via golangci-lint. I turned on debug mode for null-ls and tracked down the problem, the `--fast` flag disables a significant amount of the linters it provides, so the feedback in vim is different from other editors (like vscode) and the golangci-lint CLI output (omitting `--fast`).

This change removes `--fast` from the default arguments for `golangci-lint`. To get the old behavior back, you can use the `extra_args` argument when configuring `golangci-lint`. e.g.

```lua
null_ls.builtins.diagnostics.golangci_lint.with({
  extra_args = { "--fast" },
}),
```

Since it's easier to add extra flags, and this isn't a default flag for other editors, I thought it'd make sense to omit the `--fast` flag and prefer running all linters as the default behavior.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/golangci_lint.lua
+++ b/lua/null-ls/builtins/diagnostics/golangci_lint.lua
@@ -20,7 +20,6 @@ return h.make_builtin({
         args = {
             "run",
             "--fix=false",
-            "--fast",
             "--out-format=json",
             "--path-prefix",
             "$ROOT",


jose-elias-alvarez:
I don't know enough about this linter to say for sure, but judging from [the VS Code integration instructions](https://golangci-lint.run/usage/integrations/) it seems like the recommended setup for VS Code does indeed use the `--fast` flag. Are there any other examples?

BlakeWilliams:
> I don't know enough about this linter to say for sure, but judging from [the VS Code integration instructions](https://golangci-lint.run/usage/integrations/) it seems like the recommended setup for VS Code does indeed use the `--fast` flag. Are there any other examples?

That is true, but it makes me wonder why it's not the default configuration in VSCode. Maybe it's because it's easier to add extra flags than to remove them?

I looked through a few other similar plugins for different editors (on the integration page you linked), as far as I can tell most aren't setting `--fast` as a default, but at least one of them does.

BlakeWilliams:
For comparisons sake, here's the difference (in line count) I get when using `--fast`.

```
wordup ❯ golangci-lint run | wc -l
     102
wordup ❯ golangci-lint run --fast | wc -l
       3
```

jose-elias-alvarez:
Hmm, what is the speed difference? We shouldn't have the same problem where the extension blocks the editor, and we only run the linter on save, but it might be good to check if it's going to affect performance.

academo:
I was just setting up golangci-lint via nulls because I was using the LSP version via lsp-config that gave me problems and the lspconfig verison [doesn't include --fast](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#golangci_lint_ls) by default either so I was getting different results.

Regarding speed. Here are some tests:

Setup: 
* file of 96KB [>2K lines of code](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/state/manager_test.
* Use hyperfine for benchmarking

## Tests without cache (notice the prepare)

### without `--fast`
```
❯ hyperfine --runs 10 'GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run ./pkg/services/ngalert/state/manager_test.go' --prepare="rm -rf /tmp/testcache;mkdir /tmp/testcache"
Benchmark 1: GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run ./pkg/services/ngalert/state/manager_test.go
  Time (mean ± σ):      6.377 s ±  0.040 s    [User: 76.838 s, System: 2.875 s]
  Range (min … max):    6.331 s …  6.477 s    10 runs
```

### with `--fast`
```
❯ hyperfine --runs 10 'GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run --fast ./pkg/services/ngalert/state/manager_test.go' --prepare="rm -rf /tmp/testcache;mkdir /tmp/testcache"
Benchmark 1: GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run --fast ./pkg/services/ngalert/state/manager_test.go
  Time (mean ± σ):      8.042 s ±  2.054 s    [User: 100.977 s, System: 4.610 s]
  Range (min … max):    6.320 s … 10.790 s    10 runs
```

## Tests with cache (notice the warm up runs):

### Without `--fast`
```
❯ hyperfine --runs 10 'GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run ./pkg/services/ngalert/state/manager_test.go' --warmup=2
Benchmark 1: GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run ./pkg/services/ngalert/state/manager_test.go
  Time (mean ± σ):     668.1 ms ± 153.9 ms    [User: 1824.5 ms, System: 471.0 ms]
  Range (min … max):   559.4 ms … 949.8 ms    10 runs
```


### With `--fast`
```❯ hyperfine --runs 10 'GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run --fast ./pkg/services/ngalert/state/manager_test.go' --warmup=2
Benchmark 1: GOLANGCI_LINT_CACHE=/tmp/testcache golangci-lint run --fast ./pkg/services/ngalert/state/manager_test.go
  Time (mean ± σ):     585.2 ms ±  30.5 ms    [User: 1475.0 ms, System: 451.1 ms]
  Range (min … max):   524.1 ms … 633.2 ms    10 runs
```

## Conclusion

It looks like `--fast` is a bit faster in a cached scenario and worst in a non cached scenario (which correlate to the docs where they mention the first run won't be fast).

Considering this is quite a big file I'd say the difference between `--fast` and non `--fast` is negligible considering the more feedback you get from the non-fast

jose-elias-alvarez:
Thanks for the detailed breakdown - I agree that the difference isn't significant enough to worry about. Let's merge this and see how it goes.

PR #1346, opened 2023-01-15
feat(builtins): add clazy diagnostics

knarfS:


lua/null-ls/builtins/diagnostics/clazy.lua
```diff
@@ -0,0 +1,59 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS_ON_SAVE = methods.internal.DIAGNOSTICS_ON_SAVE
+
+local get_severity = function(severity_str)
+    local severity = h.diagnostics.severities.warning
+    if severity_str == "error" then
```
jose-elias-alvarez, line 8: Maybe easier to use early returns here?

lua/null-ls/builtins/diagnostics/clazy.lua
```diff
@@ -0,0 +1,59 @@
...
+        },
+    },
+    method = DIAGNOSTICS_ON_SAVE,
+    filetypes = { "cpp" },
+    generator_opts = {
+        command = "clazy",
+        args = {
+            "--standalone",
+            "-p",
+            "build/compile_commands.json",
```
jose-elias-alvarez, line 34: Do we always expect this to be in the `build` directory?

lua/null-ls/builtins/diagnostics/clazy.lua
```diff
@@ -0,0 +1,59 @@
...
+            "$FILENAME",
+        },
+        format = "line",
+        to_stdin = false,
+        from_stderr = true,
+        on_output = function(line, params)
```
jose-elias-alvarez, line 40: You may be able to use our [`from_pattern`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/cff4c345d69f74c70192e8b59ae065ba0808c622/lua/null-ls/helpers/diagnostics.lua#L92) helper here. If you're worried about non-matching lines, it'll automatically discard incomplete diagnostics.

lua/null-ls/builtins/diagnostics/clazy.lua
knarfS: Fixed

knarfS: I was using the same options as the `clang_check` diagnostics parser, but you are correct. It would probably work for most of the users, but not for all.
I added/updates the notes to reflect that. Let me know if the notes are too verbose now or if you don't like the wording.

Also let me know, if I should open a PR to update the `clang_check` options. But that would be a breaking change?

lua/null-ls/builtins/diagnostics/clazy.lua
knarfS: Unfortunately, `clazy` returns not only findings for the actual file, but also for some of the included files. I had to filter out those diagnostics. Therefor I couldn't use `from_pattern`.

I also added the options `--ignore-included-files` and `--header-filter` to reduce the noise in the output but couldn't get rid of it completely. I addition, according to `clazy`s help output, these to options are useful for performance reasons.

knarfS: One of the added tests checks, if those files are filtered out correctly.

jose-elias-alvarez:
LGTM, thanks!

Issue #1321, opened 2023-01-04
Range formatting for prettierd


sarmong:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

prettierd [supports range formatting](https://github.com/fsouza/prettierd/issues/432#issuecomment-1366976236), but it is not included in null-ls

### Help

Yes, but I don't know how to start. I would need guidance

jose-elias-alvarez:
Since it looks like prettierd is also using `--range-start` and `--range-end`, it may work if we just copy-paste the `args` key from [the Prettier built-in](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/prettier.lua).

sarmong:
But it requires `=` instead of space, would it work by default?

jose-elias-alvarez:
You're right, the helper won't handle this case. I haven't tested this much but something like this should at least help you get started (the offsets might not be exactly right):

```diff
diff --git a/lua/null-ls/builtins/formatting/prettierd.lua b/lua/null-ls/builtins/formatting/prettierd.lua
index 4c82eba..c20dff2 100644
--- a/lua/null-ls/builtins/formatting/prettierd.lua
+++ b/lua/null-ls/builtins/formatting/prettierd.lua
@@ -4,6 +4,7 @@ local methods = require("null-ls.methods")
 local u = require("null-ls.utils")
 
 local FORMATTING = methods.internal.FORMATTING
+local RANGE_FORMATTING = methods.internal.RANGE_FORMATTING
 
 return h.make_builtin({
     name = "prettierd",
@@ -11,7 +12,7 @@ return h.make_builtin({
         url = "https://github.com/fsouza/prettierd",
         description = "prettier, as a daemon, for ludicrous formatting speed.",
     },
-    method = FORMATTING,
+    method = { FORMATTING, RANGE_FORMATTING },
     filetypes = {
         "javascript",
         "javascriptreact",
@@ -32,7 +33,18 @@ return h.make_builtin({
     },
     generator_opts = {
         command = "prettierd",
-        args = { "$FILENAME" },
+        args = function(params)
+            if params.method == FORMATTING then
+                return { "$FILENAME" }
+            end
+
+            local row, end_row = params.range.row - 1, params.range.end_row - 1
[...]

PR #1354, opened 2023-01-18
 feat(prettierd): add support for range formatting 

sarmong:
#1321


jose-elias-alvarez:
Glad this worked, thanks!

PR #1356, opened 2023-01-20
add search by project of phpstan and psalm binaries

andresmeireles:
Add searching of phpstan and psalm binaries on  project directory where neovim was opened before searching on the global path


andresmeireles:
not follow contributing rules

PR #1363, opened 2023-01-21
fix: builtin.with should use deeply merging instead of shallow merging

adoyle-h:
For example,

```
builtins.diagnostics.editorconfig_checker.with {
  generator_opts = {
    command = 'editorconfig-checker',
  },
}
```

null-ls will throw an error: `on_output: expected function, got nil`. Because the default values of generator_opts will be overwritten, and `generator_opts.on_output` will be `nil`.

### DIFF
--- a/lua/null-ls/helpers/make_builtin.lua
+++ b/lua/null-ls/helpers/make_builtin.lua
@@ -101,7 +101,7 @@ local function make_builtin(opts)
     generator_opts._last_cwd = nil
 
     builtin.with = function(user_opts)
-        return make_builtin(vim.tbl_extend("force", opts, user_opts))
+        return make_builtin(vim.tbl_deep_extend("force", opts, user_opts))
     end
 
     return builtin


jose-elias-alvarez:
While I think this change is theoretically fine, you can currently accomplish this by doing the following:

```lua
builtins.diagnostics.editorconfig_checker.with {
  command = 'editorconfig-checker',
}
```

The [logic here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/ef3d4a438f96865e3ae018e33ed30156a955ed00/lua/null-ls/helpers/make_builtin.lua#L27) is that `generator_opts` is, from a user perspective, an implementation detail that doesn't necessarily need to be exposed. I'll freely admit that the interface is a bit of a mess, though. If there's something we can do to make this clearer, let me know.

adoyle-h:
I get it now. Thank you. @jose-elias-alvarez

PR #1367, opened 2023-01-22
fix: filter out permanent cpplint header order warnings

### LINKED ISSUE #1351: Cpplint diagnostic - temp file causes unwanted warnings
### FAQ

### Issues

### Neovim Version

0.8.1

### Dev Version?

- [ ] I am using a stable Neovim release version, or if I am using a dev version of Neovim I have confirmed that my issue is reproducible on a stable version.

### Operating System

macOS Ventura 13.0.1

### Minimal Config

```Lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = { null_ls.builtins.diagnostics.cpplint },
        debug = true,
    })
end

local function load_plugins()
    -- only add other plugins if they are necessary to reproduce the issue
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {
                "jose-elias-alvarez/null-ls.nvim",
                requires = { "nvim-lua/plenary.nvim" },
                config = null_ls_config,
            },
        },
        config = {
            package_root = package_root,
            compile_path = compile_path,
        },
    })
end

if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system({ "git", "clone", "https://github.com/wbthomason/packer.nvim", install_path })
    load_plugins()
    require("packer").sync()
else
    load_plugins()
    require("packer").sync()
end
```

### Steps to Reproduce

1. Create a simple C header file, e.g:
```
class Test {
 public:
  void func();
};
```
2. Write an implementation so that a system header is after the `test.h` header:
```
#include "./test.h"

#include <string>

void Test::func() { std::string foo = "bar"; }
```

3. The warning `Found C++ system header after other header. Should be: .null-ls_528489_test.h, c system, c++ system, other.` appears.

### Reproducibility Check

- [X] I confirm that my minimal config is based on the `minimal_init.lua` template and that my issue is reproducible by running `nvim --clean -u minimal_init.lua` and following the steps above.

### Expected Behavior

The warning should not appear since the headers are in proper order (e.g: file header, c/cpp system and other headers). Cpplint sees the first include as a `other` header, which is incorrect.

### Actual Behavior

Since `null-ls` serves the generated temp file to `cpplint`, the linter returns the warning message, seeing `.null-ls_#######_test.h` as the correct file header.

### Debug Log

```
[TRACE Mon Jan 16 20:40:31 2023] ...te/pack/packer/start/null-ls.nvim/lua/null-ls/client.lua:97: starting null-ls client
[TRACE Mon Jan 16 20:40:31 2023] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:102: received LSP request for method initialize
[TRACE Mon Jan 16 20:40:31 2023] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:127: received LSP notification for method initialized
[TRACE Mon Jan 16 20:40:31 2023] .../site/pack/packer/start/null-ls.nvim/lua/null-ls/rpc.lua:127: received LSP notification for method textDocument/didOpen
[TRACE Mon Jan 16 20:40:31 2023] ...ack/packer/start/null-ls.nvim/lua/null-ls/generators.lua:21: running generators for method NULL_LS_DIAGNOSTICS_ON_OPEN
[DEBUG Mon Jan 16 20:40:31 2023] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:320: spawning command "cpplint" at /Users/mdvictor/Workspace/repro-null-ls-err with args { "/Users/mdvictor/Workspace/repro-null-ls-err/.null-ls_814785_test.cpp" }
[TRACE Mon Jan 16 20:40:31 2023] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:204: error output: /Users/mdvictor/Workspace/repro-null-ls-err/.null-ls_814785_test.cpp:0:  No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"  [legal/copyright] [5]
/Users/mdvictor/Workspace/repro-null-ls-err/.null-ls_814785_test.cpp:3:  Found C++ system header after other header. Should be: .null-ls_814785_test.h, c system, c++ system, other.  [build/include_order] [4]

[TRACE Mon Jan 16 20:40:31 2023] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:205: output: Done processing /Users/mdvictor/Workspace/repro-null-ls-err/.null-ls_814785_test.cpp
Total errors found: 2

[TRACE Mon Jan 16 20:40:31 2023] ...ck/packer/start/null-ls.nvim/lua/null-ls/diagnostics.lua:181: received diagnostics from source 1
[TRACE Mon Jan 16 20:40:31 2023] ...ck/packer/start/null-ls.nvim/lua/null-ls/diagnostics.lua:182: { {
    col = 0,
    end_col = 0,
    end_lnum = 0,
    lnum = 0,
    message = 'No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"',
    row = "0",
    severity = 3,
    source = "cpplint"
  }, {
    col = 0,
    end_col = 0,
    end_lnum = 3,
    lnum = 2,
    message = "Found C++ system header after other header. Should be: .null-ls_814785_test.h, c system, c++ system, other.",
    row = "3",
    severity = 2,
    source = "cpplint"
  } }
```

### Help

### Implementation Help

I've been looking into `lua/null-ls/builtins/diagnostics/cpplint.lua` and removing the `to_temp_file` option seems to fix it, in the sense that the warning does not appear anymore, but since the original file is being served to `cpplint` now, warnings, in general, appear or not only after the file is saved (and thus the new version served to `cpplint`). 

Or that is my understanding, at least. I don't have much experience with neither `nvim` nor `null-ls`. To fix the stale warnings,  I added:
```
       null_ls.builtins.diagnostics.cpplint.with({
               method = null_ls.methods.DIAGNOSTICS_ON_SAVE,
       })
```

So I get the latest `cpplint` messages on save, but I'm not sure that this is the best way.

I could try to fix this with some guidance, if possible. I'd say the first thing would be to remove the temp file option, after that I'm not sure.

### Requirements

mdvictor:
Fixes the situation where header order warnings returned by `cpplint` cannot be fixed because `null-ls` calls the linter on a generated temp file. More details in this [discussion](https://github.com/jose-elias-alvarez/null-ls.nvim/discussions/1300).

It seems there are multiple messages that `cpplint` can return based on improper order of headers in a file, but they are all labelled as `build/include_order`, code [here](https://github.com/cpplint/cpplint/blob/fa12a0bbdafa15291276ddd2a2dcd2ac7a2ce4cb/cpplint.py#L5147).

This PR checks if the returned message from `cpplint` has severity `build` and label `include_order` and filters the message out  if it does.

Fixes #1351

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cpplint.lua
+++ b/lua/null-ls/builtins/diagnostics/cpplint.lua
@@ -37,7 +37,7 @@ return h.make_builtin({
         on_output = function(line, params)
             local rez = diagnostics(line, params)
 
-            if rez.severity == 2 and rez.label == "include_order" then
+            if rez.severity == 2 and (rez.label == "include_order" or rez.label == "header_guard") then
                 return nil
             end
 
--- a/lua/null-ls/builtins/diagnostics/cpplint.lua
+++ b/lua/null-ls/builtins/diagnostics/cpplint.lua
@@ -3,6 +3,20 @@ local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
+local diagnostics = h.diagnostics.from_pattern(
+    "[^:]+:(%d+):  (.+)  %[(.+)%/(.+)%] %[%d+%]",
+    { "row", "message", "severity", "label" },
+    {
+        severities = {
+            build = h.diagnostics.severities["warning"],
+            whitespace = h.diagnostics.severities["hint"],
+            runtime = h.diagnostics.severities["warning"],
+            legal = h.diagnostics.severities["information"],
+            readability = h.diagnostics.severities["information"],
+        },
+    }
+)
+
 return h.make_builtin({
     name = "cpplint",
     meta = {
@@ -20,19 +34,15 @@ return h.make_builtin({
         to_stdin = false,
         from_stderr = true,
         to_temp_file = true,
-        on_output = h.diagnostics.from_pattern(
-            "[^:]+:(%d+):  (.+)  %[(.+)%/.+%] %[%d+%]",
-            { "row", "message", "severity" },
-            {
-                severities = {
-                    build = h.diagnostics.severities["warning"],
-                    whitespace = h.diagnostics.severities["hint"],
-                    runtime = h.diagnostics.severities["warning"],
-                    legal = h.diagnostics.severities["information"],
-                    readability = h.diagnostics.severities["information"],
-                },
-            }
-        ),
+        on_output = function(line, params)
+            local rez = diagnostics(line, params)
+
+            if rez.severity == 2 and rez.label == "include_order" then
+                return nil
+            end
+
+            return rez
+        end,
         check_exit_code = function(code)
             return code >= 1
         end,


jose-elias-alvarez:
This looks good - one change I would consider making is to actually include the filtering code in `on_output`. With this implementation, users who want to set their own `filter` function would override the default one included here. Something like this may work (untested):

```lua
on_output = function(line, params)
    local diagnostic = h.diagnostics.from_pattern(
        "[^:]+:(%d+):  (.+)  %[(.+)%/(.+)%] %[%d+%]",
        { "row", "message", "severity", "label" },
        {
            severities = {
                build = h.diagnostics.severities["warning"],
                whitespace = h.diagnostics.severities["hint"],
                runtime = h.diagnostics.severities["warning"],
                legal = h.diagnostics.severities["information"],
                readability = h.diagnostics.severities["information"],
            },
        }
    )
    if diagnostic.severity == 2 and diagnostic.label == "include_order" then
        return
    end

    return diagnostic
end
```

mdvictor:
I changed the code to do the filtering inside `on_output`

mdvictor:
Need to also filter out label: `header_guard`
Example file: https://github.com/cpplint/cpplint/blob/fa12a0bbdafa15291276ddd2a2dcd2ac7a2ce4cb/samples/protobuf-sample/simple.def#L1631

This one also uses the temp file name. Will push later today.

jose-elias-alvarez:
Great, thanks!

PR #1368, opened 2023-01-22
fix: prettier bug adding unecessary output to file

milkias17:
Without the --no-plugin-search option, null_ls includes extraneous output from auto loaded prettier plugins into the file. So for example, in the following output, the addition information was being inserted into the neovim buffer.

image

### DIFF
--- a/lua/null-ls/builtins/formatting/prettier.lua
+++ b/lua/null-ls/builtins/formatting/prettier.lua
@@ -38,6 +38,7 @@ return h.make_builtin({
     generator_opts = {
         command = "prettier",
         args = h.range_formatting_args_factory({
+            "--no-plugin-search",
             "--stdin-filepath",
             "$FILENAME",
         }, "--range-start", "--range-end", { row_offset = -1, col_offset = -1 }),


jose-elias-alvarez:
I understand the issue, but this looks like it disables plugins entirely. I tried this with [prettier-plugin-toml](https://github.com/bd82/toml-tools/tree/master/packages/prettier-plugin-toml) and Prettier no longer recognizes the parser.

The TOML plugin is the only one that I've tried, but I also wonder why it doesn't seem to have the same issue. What plugin is causing this issue for you? Is it possible that the plugin itself is inserting the extra output?

milkias17:
The issue only occurs when plugins include extra output to prettier like in the screenshot. So in my case, it's a ui library called daisyUI and it just automatically inserts the above text to the prettier output. And so, null_ls treats it as if it was the formatted file output so when I ask to format the buffer, the extra output is also inserted. And I checked if this is something that a prettier plugin shouldn't do and if this is an error that would be caused anywhere by testing it out on vscode and the prettier plugin handles it correctly.

In any case, with the plugins that I tested, I assumed it didn't disable plugins but if it does I guess this solution isn't really viable.

milkias17:
would maybe changing the command to do in place formatting be viable?

jose-elias-alvarez:
I would argue that the plugin itself is wrong in inserting extra outputs, since it'll break integrations that rely on capturing `stdout`. How does the Prettier VS Code extension handle this?

milkias17:
In the vscode extension's case, although I haven't checked, I would assume they use the [prettier api](https://prettier.io/docs/en/api.html) which doesn't rely on stdout. In null-ls's case, although that isn't an option, I would have assumed using in place formatting(-w flag) would have worked. But no for some reason the logged output from the plugin still appears(this time replacing the whole buffer)

jose-elias-alvarez:
Right, that makes sense. I would then argue that the extension should remove the extra output, since like I said it will cause issues for any similar tools that rely on `stdout` (which is specifically encouraged by Prettier itself).

milkias17:
Yea I guess that makes sense, I didn't like that it did it either, I just thought if there was an easy way to handle this type of issue, its worthwhile to do so.

PR #1381, opened 2023-02-04
fix(ansiblelint): Use 'description' field

andrewferrier:
For ansiblelint, typically the `check_name` field does not contain particularly helpful content, such as `yaml[indentation]`. The `description` field is human-readable, containing the equivalent `Wrong indentation: expected 8 but found 7`. This makes for much more helpful reading when looking at the diagnostics output from `ansiblelint`.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/ansiblelint.lua
+++ b/lua/null-ls/builtins/diagnostics/ansiblelint.lua
@@ -37,7 +37,7 @@ return h.make_builtin({
                 table.insert(params.messages, {
                     row = row,
                     col = col,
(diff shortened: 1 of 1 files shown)


andrewferrier:
Actually, an even better change here would be to put `message.check_name` in the `code` field, as it is effectively a short code for the error.

jose-elias-alvarez:
I'll defer to you - just let me know when this is ready to merge (I think there's a test for this built-in that needs to be fixed up, too).

jose-elias-alvarez:
Superseded by #1434.

PR #1383, opened 2023-02-05
feat(builtins): add erb-formatter support

vascomfnunes:
Adds a built-in formatter to support `erb-formatter` to handle embedded Ruby files (https://github.com/nebulab/erb-formatter)

`erb-formatter` seems to produce much better results than `erb_lint`.


jose-elias-alvarez:
LGTM, thanks!

PR #1384, opened 2023-02-06
fix: utils.root_pattern not checking the start_path for patterns

laytan:
In https://github.com/jose-elias-alvarez/null-ls.nvim/commit/f1182c2d3748da7c94f5288f8054741f26f0cf3b the function was changed from using `M.search_ancestors` which first checked the `start_path` and then the ancestors, to a `vim.fn.parents` iteration which does not include the `start_path` check anymore.

This PR adds a check for the `start_path` first (like it was before that commit) and then does the `vim.fn.parents`.


jose-elias-alvarez:
Thanks for catching this and for the PR!

PR #1386, opened 2023-02-07
feat(buitlins): add verilator-linter support

junjie1475:


lua/null-ls/builtins/diagnostics/verilator.lua
```diff
@@ -0,0 +1,38 @@
...
+
+local FORMATTING = methods.internal.DIAGNOSTICS
+
+return h.make_builtin({
+    name = "verilator",
+    meta = {
+        url = "https://www.veripool.org/verilator/",
+        description = "Verilog and SystemVerilog linter power by Verilator",
+    },
+    method = FORMATTING,
```
jose-elias-alvarez, line 12: Unless I'm missing something, I don't think this is the correct method, right?

junjie1475: Fixed

jose-elias-alvarez:
Thanks!

PR #1388, opened 2023-02-08
fix(builtins): handle extra files in ansible-lint results

halftan:
ansible-lint will include lint results from included/imported task files.

```yaml
# roles/foobar/tasks/main.yml
- name: Ping
  ping:

- name: Include other task
  include_tasks: baz.yml
```

```yaml
# roles/foobar/tasks/baz.yml
- name: Some lint results here
  set_fact:
    foo: yes

```

ansible-lint result:
```bash
ansible-lint -P=basic -f codeclimate -q roles/foobar/tasks/main.yml | jq
[
  {
    "type": "issue",
    "check_name": "yaml[truthy]",
    "categories": [
      "formatting",
      "yaml"
    ],
    "severity": "info",
    "description": "truthy value should be one of \\[false, true]",
    "fingerprint": "8564f80bca9d93054e646e7ec5cfe5ee2c279b821ca183183c79d055bc062b8d",
    "location": {
      "path": "roles/foobar/tasks/baz.yml",
      "lines": {
        "begin": 3
      }
    }
  }
]
```

ansible-lint was run against **main.yml** but the result was from **baz.yml**, which will result in a confusing floating text in nvim:
image

This PR tries to fix this issue.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/ansiblelint.lua
+++ b/lua/null-ls/builtins/diagnostics/ansiblelint.lua
@@ -28,19 +28,21 @@ return h.make_builtin({
             }
             params.messages = {}
             for _, message in ipairs(params.output) do
-                local col = nil
-                local row = message.location.lines.begin
-                if type(row) == "table" then
-                    row = row.line
-                    col = row.column
+                if params.temp_path:match(vim.pesc(message.location.path)) then
+                    local col = nil
+                    local row = message.location.lines.begin
+                    if type(row) == "table" then
+                        row = row.line
+                        col = row.column
+                    end
+                    table.insert(params.messages, {
+                        row = row,
+                        col = col,
+                        message = message.check_name,
+                        severity = severities[message.severity],
+                        filename = params.bufname,
+                    })
                 end
-                table.insert(params.messages, {
-                    row = row,
-                    col = col,
-                    message = message.check_name,
-                    severity = severities[message.severity],
-                    filename = message.location.path,
-                })
             end
             return params.messages
         end,
--- a/test/spec/builtins/diagnostics_spec.lua
+++ b/test/spec/builtins/diagnostics_spec.lua
@@ -1035,35 +1035,58 @@ describe("diagnostics", function()
         it("should create a diagnostic", function()
             local output = [[
                 [
-                  {
-                    "type": "issue",
-                    "check_name": "[risky-file-permissions] File permissions unset or incorrect",
-                    "categories": [
-                      "unpredictability",
-                      "experimental"
-                    ],
-                    "severity": "blocker",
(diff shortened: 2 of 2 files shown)


lua/null-ls/builtins/diagnostics/ansiblelint.lua
```diff
@@ -18,6 +18,7 @@ return h.make_builtin({
         check_exit_code = function(code)
             return code <= 2
         end,
+        multiple_files = true,
```
jose-elias-alvarez, line 21: Is this necessary? If the goal is to filter non-matching results then I'm not sure we want to actually handle these results.

lua/null-ls/builtins/diagnostics/ansiblelint.lua
```diff
@@ -30,6 +31,10 @@ return h.make_builtin({
             for _, message in ipairs(params.output) do
                 local col = nil
                 local row = message.location.lines.begin
+                local filename = message.location.path
+                if message.location.path:find(".null-ls_", 1, true) then
```
jose-elias-alvarez, line 35: Again, not sure whether the goal here is to filter diagnostics or handle them. If you want to replace the temp file path with the actual path, it should be available under `params.bufname`.

lua/null-ls/builtins/diagnostics/ansiblelint.lua
halftan: Makes sense. It was just me trying to be lazy by letting null-ls handle results for other files.

halftan:
Update:
- Removed `multiple_files`
- Use `params.bufname`, `params.root`, and `params.temp_path` to filter ansible-lint results

lua/null-ls/builtins/diagnostics/ansiblelint.lua
```diff
@@ -28,19 +28,21 @@ return h.make_builtin({
             }
             params.messages = {}
             for _, message in ipairs(params.output) do
-                local col = nil
-                local row = message.location.lines.begin
-                if type(row) == "table" then
-                    row = row.line
-                    col = row.column
+                if params.root .. "/" .. message.location.path == params.temp_path then
```
jose-elias-alvarez, line 31: Is it necessary to check that these two paths are equal? It seems simpler to just check that `params.temp_path` contains `message.location.path`:

```lua
if message.location.path:match(vim.pesc(params.temp_path)) then
```

If that doesn't work, then we should use our `path.join` utility to avoid potential cross-platform issues with path separators.

halftan: 👍 Good idea. I'm not a vim veteran. Just learned nvim has `vim.pesc` to handle special characters in lua regexp. This really helped a lot.

halftan:
Update:
- Use `vim.pesc` to escape strings so that we can use `string.match` to match sub-paths

jose-elias-alvarez:
Thanks!

Issue #1378, opened 2023-02-01
For textlint, add dynamic code = `from_node_modules()`


sys9kdr:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

textlint is a node module. So I would like to add `dynamic_command = cmd_resolver.from_node_modules()`.

* https://github.com/jose-elias-alvarez/null-ls.nvim/blob/944c72a0a85b4a6affddda60d55eacf83075b718/lua/null-ls/builtins/formatting/textlint.lua
* https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b287e9d124037a7ea5de8e3368a3b75274cec8b4/lua/null-ls/builtins/diagnostics/textlint.lua

### Help

Yes

jose-elias-alvarez:
Sounds fine to me, should be a straightforward PR.

jose-elias-alvarez:
Closed by #1396

PR #1396, opened 2023-02-11
Textlint dynamic command

sys9kdr:
* Add dynamic commands (see #1378).
* Fix a blank setting for filetypes.
* Add cwd setting.

Sorry. More fixes than I explained in the Issue.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/textlint.lua
+++ b/lua/null-ls/builtins/diagnostics/textlint.lua
@@ -1,5 +1,7 @@
 local h = require("null-ls.helpers")
+local cmd_resolver = require("null-ls.helpers.command_resolver")
 local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
 
@@ -37,6 +39,7 @@ return h.make_builtin({
         format = "json_raw",
         check_exit_code = { 0, 1 },
         on_output = handle_output,
+        dynamic_command = cmd_resolver.from_node_modules(),
         cwd = h.cache.by_bufnr(function(params)
             return u.root_pattern(
                 -- https://textlint.github.io/docs/configuring.html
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/textlint.lua
```diff
@@ -29,14 +31,26 @@ return h.make_builtin({
         description = "The pluggable linting tool for text and Markdown.",
     },
     method = DIAGNOSTICS,
-    filetypes = {},
+    filetypes = {"txt", "markdown"},
```
jose-elias-alvarez, line 34: I don't really know much about this tool, but is this change correct? The documentation says the following:

> textlint supports Markdown and plain text by default.

sys9kdr: Yes. This change is correct. This tool supports markdown and plain text (filetype=txt in neovim).

sys9kdr: Sorry. I forgot Checkstyle. I'll fix style.

jose-elias-alvarez: My bad, I was looking at the diff wrong (for some reason I thought your PR was removing filetypes and not adding them).

jose-elias-alvarez:
Thanks for putting this in!

lua/null-ls/builtins/diagnostics/textlint.lua
yuys13: textlint supports more than just markdown and txt. Therefore, I don't think this change is correct.
https://github.com/textlint/textlint#supported-file-formats
However, users can be added with extra_filetypes, so this may not be a problem.

jose-elias-alvarez: For similar sources like Prettier that support additional filetypes with plugins, our recommended approach is for users to specify the additional filetypes using `extra_filetypes`, so I think this change is OK.

yuys13: Well, that certainly makes sense.

PR #1400, opened 2023-02-14
feat(builtins): add vacuum

cbochs:
This PR adds support for diagnostics using Vacuum, a fast OpenAPI linter written in Go.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/vacuum.lua
+++ b/lua/null-ls/builtins/diagnostics/vacuum.lua
@@ -1,0 +1,46 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+local severities = { 1, 2, 3, 4 }
+
+return h.make_builtin({
+    name = "vacuum",
+    meta = {
+        url = "https://quobix.com/vacuum",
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/diagnostics/vacuum.lua
```diff
@@ -0,0 +1,46 @@
...
+    generator_opts = {
+        command = "vacuum",
+        args = {
+            "spectral-report",
+            "--stdin",
+            "--stdout",
+        },
+        format = "json",
+        to_stdin = true,
+        ignore_stderr = true,
```
jose-elias-alvarez, line 24: Quick question - is this necessary? What happens to diagnostics if the linter is unable to parse the source file, for example?

cbochs: Good question. I'm not entirely certain. In fact, this was mostly copied from [lua/null-ls/builtins/diagnostics/spectral.lua](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/spectral.lua), which is a very similar tool. I can look into this if you would like

jose-elias-alvarez: It's probably not worth stressing about, if it causes issues we can look into it later.

jose-elias-alvarez:
Thanks!

PR #1412, opened 2023-02-21
fix(cspell): diagnostic builtin uses local config

benelan:
The cspell diagnostic builtin is unable to find local configuration files because `stdin` is being provided to the CLI, which isn't aware of the file's path. Cspell's global configuration file (in the home directory) is used if it exists.

The code_action builtin works correctly. When you choose the "add to cspell json file" action it uses the closest ancestor, or when no config exists it creates on in the current directory. However since the diagnostic builtin is using the global config, the error does not go away.

You can repro when you either
1) Don't have a global config
3) Have a global config and local config

It works correclty when you have a global config and no local config. The code_action builtin adds the word to the global config, which the diagnostic builtin is able to see.

This change resolves the issue. I didn't see a contributing doc, but let me know if I need to change anything or provide additional info, such as repro sample/steps.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/cspell.lua
+++ b/lua/null-ls/builtins/diagnostics/cspell.lua
@@ -36,7 +36,7 @@ return h.make_builtin({
                 "lint",
                 "--language-id",
                 params.ft,
-                "stdin",
+                params.bufname,
             }
 
             -- only enable suggestions when using the code actions built-in, since they slow down the command


jose-elias-alvarez:
This is likely to break updating diagnostics when the buffer changes, since cspell will be reading the file as it is on the disk and not its buffer state. Could you verify?

benelan:
Ah yup nice catch, I guess I didn't get that far in my testing. Another potential option is using this from the code_action builtin:

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/689cdd78f70af20a37b5309ebc287ac645ae4f76/lua/null-ls/builtins/code_actions/cspell.lua#L16-L35

and doing:

```lua
local cspell_args = {
    "lint",
    "--language-id",
    params.ft,
    "--config",
    find_cspell_config(params.root),
    "stdin"
}
```

But that would probably impact performance, although I'm not sure how significantly. Do you have any better ideas?

benelan:
The way I (and probably most people) use cspell is I have a global config with most of my usual words, and then a few project configs with additional spellings. Maybe we could create config option for the cspell diagnostic builtin where users can provide a table of local cspell configs to look for? That could cut down on the iterations. Something like

```lua
diagnostics.cspell.with {
  config = {
    cspell_project_configs = {
      vim.fn.expand "~/dev/project1/cspell.json",
      vim.fn.expand "~/dev/work/project2/cspell.json",
      vim.fn.expand "~/dev/work/project3/cspell.json",
    },
  },
},
```
And then if none of those are an ancestor of the current buffer it defaults to the global config like it currently does.

jose-elias-alvarez:
We could theoretically use the same logic from the code action built-in and [cache the results](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/456cd2754c56c991c5e4df60a807d054c1bc7148/doc/HELPERS.md?plain=1#L353) to avoid running the same lookup each time. The downside is that if the user creates a new config file manually or via the code action source, then they'll have to restart Neovim for it to be picked up, which is not ideal.

jose-elias-alvarez:
The `cspell` built-in sources are some of the most complex we have, and there's still more features that are missing  - see #1415 for an example. Honestly I think the best approach would be to create a companion plugin that uses the null-ls API, similar to [prettier.nvim](https://github.com/MunifTanjim/prettier.nvim).

benelan:
Gotcha, that makes sense. Well I'll close this PR since it doesn't work and go the plugin route if I have the time. Thanks for the context

PR #1414, opened 2023-02-23
fix(mypy): disable multiple files

sileht:
When a file is passed to mypy, it only checks that file, so the plugin should not handle filename.

Also, filename returned by mypy is relative to cwd of where mypy is run. This cwd may be different from the current cwd of neovim, this result of diagnostic.filename to be unpredictable as neovim toolings will concat cwd to diagnostic.filename to guess the right buffer.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/mypy.lua
+++ b/lua/null-ls/builtins/diagnostics/mypy.lua
@@ -43,7 +43,7 @@ benefits of dynamic (or "duck") typing and static typing.]],
         check_exit_code = function(code)
             return code <= 2
         end,
-        multiple_files = true,
+        multiple_files = false,
         on_output = h.diagnostics.from_patterns({
             -- see spec for pattern examples
             {


jose-elias-alvarez:
> When a file is passed to mypy, it only checks that file, so the plugin should not handle filename.

I'm not sure that this is right - based on the information in #577, which added this in the first place, it does seem that `mypy` will return diagnostics for multiple files at once.

If the issue is that the returned file path is not correct, then we should fix that instead, I think.

sileht:
> > When a file is passed to mypy, it only checks that file, so the plugin should not handle filename.
> 
> I'm not sure that this is right - based on the information in #577, which added this in the first place, it does seem that `mypy` will return diagnostics for multiple files at once.

Well, it depends, mypy does support multiple files if you pass a directory as parameter, not a file. And here we only pass a file, so I don't see how it will returned multifiles.

Maybe the reporter of #577 passes more/other parameters to mypy.

> If the issue is that the returned file path is not correct, then we should fix that instead, I think.

mypy returned paths are correct but relative, not absolute, and putting a relative filename in diagnostics is broken by design, as when cwd changes, the path becomes wrong.

Here, I just propose a local change I have since a while. Fixing the path to make it absolute may also work, but I have no idea how to do that.

jose-elias-alvarez:
It’s been a while since I tested it, but if I recall correctly, the [sample repo](https://github.com/IamTheFij/mypy_multifile_test) from the issue should illustrates how this works (it was surprising to me too).

Can you give me an example of an issue caused by using partial paths in diagnostics? Either way, if that’s the issue, fixing it should be pretty simple. We’d just need to wrap `on_output` to process the diagnostic from `h.diagnostics.from_patterns` using something like `vim.loop.fs_realpath()` to make the path absolute.

rudenkornk:
Looks like I encountered same problems and created a similar PR: https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1439
Also, `mypy` does return diagnostics for multiple files even if you passed a specific file. Consider this example:
```python
# Contents of value.py

def get_some_value():
    return "value"
```
```python
# Contents of typed_context.py

from .value import get_some_value

def print_some_value() -> None:
    print(get_some_value())
```
```bash
> mypy typed_context.py
value.py:1: error: Function is missing a return type annotation  [no-untyped-def]
typed_context.py:4: error: Call to untyped function "get_some_value" in typed context  [no-untyped-call]
Found 2 errors in 2 files (checked 1 source file)
```

jose-elias-alvarez:
I merged #1439 which takes care of the absolute path issue. As discussed here, I don't think the `multiple_files` change is correct, so I'll close this.

PR #1416, opened 2023-02-24
feat(builtins): allow using dprint bin from node_modules

Gabrielbdd:
The current dprint config only works when dprint is installed globally (and present on PATH). So, it does not work when dprint is locally installed with some Node.js's package manager (ex: `npm i -D dprint`).

This PR sets the `dynamic_command` option on the dprint formatter to allow that (I just copied/pasted from the prettier config :yum: ).


jose-elias-alvarez:
Thanks!

PR #1422, opened 2023-02-27
fix(rubocop): change auto-correct to autocorrect

tetzng:
I encountered an error message stating
`'--auto-correct' is deprecated and should be replaced with '--autocorrect'`

It's due to Rubocop v1.30.0 where 'auto-correct' became a duplicate and was replaced with 'autocorrect'. 
https://github.com/rubocop/rubocop/pull/10547

Therefore, I have made the necessary correction to the code to use '--autocorrect' instead."

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -3979,7 +3979,7 @@ local sources = { null_ls.builtins.formatting.rubocop }
 - Filetypes: `{ "ruby" }`
 - Method: `formatting`
 - Command: `rubocop`
-- Args: `{ "--auto-correct", "-f", "quiet", "--stderr", "--stdin", "$FILENAME" }`
+- Args: `{ "-a", "-f", "quiet", "--stderr", "--stdin", "$FILENAME" }`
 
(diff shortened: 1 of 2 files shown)


jose-elias-alvarez:
Thanks, this is definitely something we want to fix. One question: how do older versions of Rubocop behave when `--autocorrect` is used? If they break completely then I want to mention this in our breaking changes issue.

tetzng:
Thank you for your response. Regarding your question, in older versions of Rubocop, using `--autocorrect` will result in an error message like the following:

> invalid option: --autocorrect
> Did you mean?  auto-correct

Alternatively, you can use the shorthand `-a`. In older versions, this means `--auto-correct`, while in version 1.30.0 and later, it means `--autocorrect`.

If a fix is necessary, please let us know and we will make the necessary changes.

jose-elias-alvarez:
Normally I prefer using the long form, but in this case I think `-a` is the right call, since it avoids the issue entirely. Could you update the PR to use `-a` and leave a comment explaining why we're using the shorthand?

PR #1425, opened 2023-03-01
Fix crystal formatting

vypxl:
Crystal formatting is broken as `crystal tool format` by default formats files in-place.

To format stdout (as null-ls expects), one needs to add an argument `-`. That way `crystal tool format` reads from stdin and writes to stdout.


jose-elias-alvarez:
Thanks!

PR #1427, opened 2023-03-02
feat: Add lisp-format formatter for Common Lisp

even4void:
lisp-format is a fixer for Common Lisp, Emacs and Scheme code. (Neo)vim already supports formatting Lisp code using the internal formatting functions for Lisp, among others. This formatter avoids the need to define autocommand to format on save using `equalprg` facilities.

### DIFF
--- a/lua/null-ls/builtins/formatting/lispformat.lua
+++ b/lua/null-ls/builtins/formatting/lispformat.lua
@@ -1,0 +1,23 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "lispformat",
+    meta = {
+        url = "https://github.com/eschulte/lisp-format",
+        description = "A tool to format lisp code. Designed to mimic clang-format.",
+        notes = {
+            "This requires a working installation of vanilla Emacs. To install Emacs, refer to the [installation instructions](https://www.gnu.org/software/emacs/manual/html_node/efaq/Installing-Emacs.html) or use a package manager.",
+        },
+    },
+    method = FORMATTING,
+    filetypes = { "lisp" },
+    generator_opts = {
+        command = "lisp-format",
+        args = { "$FILENAME" },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lua/null-ls/builtins/formatting/lispformat.lua
```diff
@@ -0,0 +1,23 @@
...
+        description = "A tool to format lisp code. Designed to mimic clang-format.",
+        notes = {
+            "This requires a working installation of vanilla Emacs. To install Emacs, refer to the [installation instructions](https://www.gnu.org/software/emacs/manual/html_node/efaq/Installing-Emacs.html) or use a package manager.",
+        },
+    },
+    method = FORMATTING,
+    filetypes = { "lisp" },
+    generator_opts = {
+        command = "lisp-format",
+        args = { "$FILENAME" },
```
jose-elias-alvarez, line 19: This argument structure seems a little strange. I also found [this issue](https://github.com/eschulte/lisp-format/issues/4) which seems to hint that it may not work as expected. Can you verify that this works with `stdin` / `stdout` vs. just formatting the file in place?

even4void: I noticed other cases of builtin formatters which make use of `$FILENAME`. My understanding is that lisp-format expect a list of options followed by the file itself, as discussed in the issue you referenced. 

The above settings worked for me when I defined the source as a builtin as is. If I replace `$FILENAME` with `-`, it doesn't work. Note that I've tested the above code both as a builtin in null-ls or as a user-defined source. I can't tell more about possible issue.

jose-elias-alvarez: Right, so if it formats the file in place, the buffer will go out of sync with the file on disk. If a user has [formatting on save](https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Formatting-on-save) set up, for example, this sequence of events may cause issues:

1. Make changes to the buffer
2. Write the changes, triggering the autocommand
3. null-ls spawns the formatting source, which writes to the file on disk (ignoring changes to the buffer)
4. Neovim warns the user that the file has been changed

If the formatter doesn’t support `stdio`, then we can either put in a change upstream or [use the `to_temp_file` option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#to_temp_file). We also have [other Emacs-based sources](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/456cd2754c56c991c5e4df60a807d054c1bc7148/lua/null-ls/builtins/formatting/emacs_scheme_mode.lua#L7) that do support `stdin`, but I’m not sure if there’s any way to do the same with a wrapper like this.

even4void: Indeed, it looks like this has unwanted side-effects like ignoring changes to the buffer, which just happened to me.  I will look into possible workarounds as you suggest (temporary file or support for stdin). Thanks for the  follow-up, I appreciate your responsiveness.

PR #1434, opened 2023-03-06
fix(ansiblelint): Use 'description' and 'check_name' field

casenull:
This is in reference to #1381, this adds the same changes (using the 'description' field of ansiblelint as message and the 'check_name' field as code).

I'm not sure if I'm using the `code` property correctly.

The tests have also been modified accordingly.


jose-elias-alvarez:
Based on the comments in #1381 I think this is a good use of `code`. Thanks for putting this in!

PR #1436, opened 2023-03-06
Fix ocamlformat not overwriting

itsjoeoui:
Before, it was not overwriting the original file for some reason.


itsjoeoui:
Hey @jose-elias-alvarez, I hope you are doing well :)
I was trying to configure ocamlformat today, and it was not working (not overriding the original file)
After reading the discussion at #1162 , and making this change, it started working just fine.

itsjoeoui:
Demo: https://cscloud.itsjoeoui.com/vd6cW0Fz

jose-elias-alvarez:
Thanks! Not sure what happened in #1162 but based on the documentation I think this is the right fix.

PR #1439, opened 2023-03-07
fix(mypy): Fix wrong filenames in mypy diagnostics

rudenkornk:
mypy has a wrong configuration, providing erroneous combination of "on_output" and "cwd" args for generator_factory.

In this configuration cwd often returns the root of the project, which contains some config such as "pyproject.toml" or "mypy.ini". The problem is that in the same time "on_output" does not take into account such non-standard cwd and returns mypy's output as is, which has relative (!) paths. As a result, null-ls registers buffers, relative to the root of the project (rather than the actual cwd).

Neovim fails to show such diagnostics (because it has wrong paths) and some other tools, for example "folke/trouble.nvim" also fail, because they try to set non-trivial cursor position to a non-existent buffer.

This commit asks mypy to show absolute paths, which solves the problem.


rudenkornk:
I am not sure if this is a correct way to fix this. Maybe there are some helpers or just a better way to rewrite `on_output` function here.
Also, I do not know who is responsible for generating correct paths:
1. Is it linter config, which must output correct absolute paths (which I did in this PR)?
2. Or is it something like `generator_factory`, who should convert filenames to absolute (or correct relative) versions using linter configuration `cwd` parameter.

lua/null-ls/builtins/diagnostics/mypy.lua
```diff
@@ -44,34 +54,36 @@ benefits of dynamic (or "duck") typing and static typing.]],
...
+                    overrides = overrides,
+                },
+                -- no column or error code
+                {
+                    pattern = "([^:]+):(%d+): (%a+): (.*)",
+                    groups = { "filename", "row", "severity", "message" },
+                    overrides = overrides,
+                },
+            })
+            local diagnostics = parser(line, params)
```
jose-elias-alvarez, line 78:
```suggestion
            local diagnostic = parser(line, params)
```

lua/null-ls/builtins/diagnostics/mypy.lua
```diff
@@ -44,34 +54,36 @@ benefits of dynamic (or "duck") typing and static typing.]],
...
+            if diagnostics and diagnostics.filename then
+                local dir = cwd(params.bufname)
```
jose-elias-alvarez, line 80: This is likely to cause performance issues, since it'll get resolved once for each diagnostic (and diagnostics update very frequently). If we're going to go with this approach,  I would recommend reusing the `cache` helper here.

jose-elias-alvarez (commented):
Have you looked into the [`--show-absolute-path`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-show-absolute-path) flag?

rudenkornk:
> Have you looked into the [`--show-absolute-path`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-show-absolute-path) flag?

🤦‍♀️ Thanks, fixed.

jose-elias-alvarez:
I'm glad the fix was easier than expected. Thanks!

PR #1443, opened 2023-03-09
fix(gomodifytags): remove deprecated vim.treesitter.get_node_at_pos

Bacbia3696:
vim.treesitter.get_node_at_pos() is deprecated, use vim.treesitter.get_node() instead.

https://neovim.io/doc/user/deprecated.html


jose-elias-alvarez:
Thank you!

PR #1445, opened 2023-03-10
fix: bslint extra whitespace in message removed

casonadams:
Sorry I missed this when I first added the code.


jose-elias-alvarez:
Thanks!

Issue #1448, opened 2023-03-11
Inconsistent working directory across null-ls sources (pylint and flake8)


dolfandringa:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

0.8.3

### Dev Version?

- [X] I am using a stable Neovim release version, or if I am using a dev version of Neovim I have confirmed that my issue is reproducible on a stable version.

### Operating System

Fedora linux

### Minimal Config

```Lua
local null_ls_config = function()
    local null_ls = require("null-ls")
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = {
		null_ls.builtins.diagnostics.pylint,
		null_ls.builtins.diagnostics.flake8,
	},
        debug = true,
    })
end
```

### Steps to Reproduce

Create a git repository in `/home/me/project` (or in my case `/home/me/Documents/git/march`) and a subdirectory `/home/me/project/subproject` (with me `/home/me/Documents/git/march/backend`). In `subproject` start nvim on a python file. Check the `:NullLsLog` output. Flake8 and pylint are launched from different directories, causing them to look in different folders for config files. 

### Reproducibility Check

- [X] I confirm that my minimal config is based on the `minimal_init.lua` template and that my issue is reproducible by running `nvim --clean -u minimal_init.lua` and following the steps above.

### Expected Behavior

I would expect both flake8 and pylint to be launched from the current working directory that I launch nvim from, and thus, find the config in that folder. 

### Actual Behavior

In the following output you can see that flake8 is launched from `/home/me/Documents/git/march/backend` while pylint is launched from `/home/me/Documents/git/march` (which is the git root, but not the current nvim working directory).

```
[DEBUG Sat 11 Mar 2023 06:36:28 PM PST] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:320: spawning command "pylint" at /home/me/Documents/git/march with args { "--from-stdin", "/home/me/Documents/git/march/backend/march_backend/auth.py", "-f", "json" }
[...]

PR #1449, opened 2023-03-11
fix(builtins/diagnostics/pylint): set the cwd according to the config files

dolfandringa:
Similar to flake8, determine the `cwd` based on the pylint configuration files that it can find using `null-ls.utils.root_pattern`.

I tested this manually by moving the pyproject.toml and .flake8 config files around and comparing both. It seems to all work as expected. I also added some tests for pylint.
Any comments on the PR and tests would be appreciated.

### DIFF
--- a/lua/null-ls/builtins/diagnostics/pylint.lua
+++ b/lua/null-ls/builtins/diagnostics/pylint.lua
@@ -1,5 +1,6 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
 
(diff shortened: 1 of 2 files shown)


dolfandringa:
This adresses #1448

jose-elias-alvarez:
Thanks for investigating and for putting the PR together!

dolfandringa:
> Thanks for investigating and for putting the PR together!

Sure, my please. Was fun to add some Lua to my track record :D. And I love contributing to cool projects. :D Thanks for writing null-ls. Its awesome.

PR #1452, opened 2023-03-15
fix(builtins): Separate diagnostic messages for 'checkmake' linter

DanWiseProgramming:
It appears that due to the format specified for `checkmake` (missing newline character at the end of each line), all diagnostic messages are provided to `null-ls` on a single line, instead of several lines.

Before:
Screenshot 2023-03-15 at 10 45 04

WIth fix:
Screenshot 2023-03-15 at 10 58 57


jose-elias-alvarez:
Makes sense, thanks for the fix!

Issue #1454, opened 2023-03-16
gomodifytags is broken in neovim v0.8.3


ArtAndreev:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.8.3 Build type: Release LuaJIT 2.1.0-beta3 Compiled by brew@Ventura-arm64.local  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.8.3/share/nvim"  Run :checkhealth for more info

### Dev Version?

- [X] I am using a stable Neovim release version, or if I am using a dev version of Neovim I have confirmed that my issue is reproducible on a stable version.

### Operating System

macOS Ventura 13.2.1

### Minimal Config

```Lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
    local null_ls = require("null-ls")
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = {
            null_ls.builtins.code_actions.gomodifytags,
        },
        debug = true,
    })
end

local lspconfig_config = function()
    require('lspconfig').gopls.setup({
        on_attach = function(_, bufnr)
            local bufopts = {noremap = true, silent = true, buffer = bufnr}
            vim.keymap.set('n', '<leader>ca', vim.lsp.buf.code_action, bufopts)
        end,
    })
end

[...]

PR #1455, opened 2023-03-16
fix gomodifytags treesitter get node call

ArtAndreev:
Closes https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1454.

### DIFF
--- a/lua/null-ls/builtins/code_actions/gomodifytags.lua
+++ b/lua/null-ls/builtins/code_actions/gomodifytags.lua
@@ -1,6 +1,7 @@
 local h = require("null-ls.helpers")
 local log = require("null-ls.logger")
 local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
 
 local CODE_ACTION = methods.internal.CODE_ACTION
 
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/code_actions/gomodifytags.lua
```diff
@@ -109,7 +109,12 @@ return h.make_builtin({
                 -- End of Execution helpers
 
                 -- Main
-                local tsnode = vim.treesitter.get_node({ bufnr = bufnr, row = row - 1, col = col - 1 })
+                local tsnode
+                if vim.fn.has('nvim-0.9') == 1 then
```
jose-elias-alvarez, line 113: Nit: could you use [this utility](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/2fcc10637b325e656d45a760901e1e05170e36c9/lua/null-ls/utils/init.lua#L42) instead?

jose-elias-alvarez (commented):
Thanks for putting this in! Could you also run the file through `stylua` to pass CI?

jose-elias-alvarez:
Thanks!

PR #1457, opened 2023-03-17
feat: add filetype filter to get_source

abalabahaha:
As discussed in #1409, but re-using the existing `matches_filetype()` for less code duplication.

Side note, is there a reason the method filter doesn't use `matches_method()`?

### DIFF
--- a/lua/null-ls/sources.lua
+++ b/lua/null-ls/sources.lua
@@ -34,12 +34,28 @@ local matches_filetype = function(source, filetype)
     return source.filetypes[filetype] or source.filetypes["_all"] and source.filetypes[filetype] == nil
 end
 
+local matches_method = function(source, method)
+    if source.methods[method] then
+        return true
+    end
+
+    if methods.overrides[method] then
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Seems good, thanks! Not sure why `matches_method` isn’t getting used here - if you’d like to include that change, too, feel free. 

Could we also mention the new option in the [query documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/SOURCES.md#get_sourcequery)?

abalabahaha:
> Seems good, thanks! Not sure why `matches_method` isn’t getting used here - if you’d like to include that change, too, feel free.

Done in a 2nd commit, hopefully it doesn't get squashed :)

> Could we also mention the new option in the [query documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/SOURCES.md#get_sourcequery)?

I think the original commit already added `filetype` under the linked `get_source(query)` section, let me know if I missed something. Made a slight change so it's between `name` and `method` like `get_sources()`.

PR #1458, opened 2023-03-17
fix(builtins/diagnostics/vacuum): use report instead of spectral-report

cbochs:
I noticed a few errors with the original addition of this builtin. Notably,

* A generator function was missing
* Using `spectral-report` instead of vacuum's own `report` implementation
* Severities were not quite correct


jose-elias-alvarez:
I don't have the bandwidth to test this on my end but the changes look good. Thanks!

PR #1462, opened 2023-03-22
Refactoring constraint visual mode

mactep:
Hi.

 I'm creating this PR to discuss these two changes I made in my local config. The first one regards the `diagnostic_config` setting that is ignored if one try to set the same config for all null-ls sources, but works if configured per source using `with`.

The other change is just a small filter around the `refactoring.nvim` source. Since the documentation states that a visual selection is needed, I added this filter to hide the options when there is nothing selected.

I understand that this PR is way off the standards and may go against some design decisions, so I'm opening this PR as a draft. Let me know what are your thoughts on it.


jose-elias-alvarez:
The change to the refactoring.nvim source seems fine. I think I’m missing something but can you explain the rationale for the `diagnostic_config` change?

mactep:
Sorry for the delay in my response.

After reading the [documentation](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONFIG.md#diagnostic_config-table-optional) I tried to hide the virtual text for all my diagnostics sources using the `diagnostic_config` property, but it didn't work.

To pinpoint what was going on, first I tried setting that setting per server inside `with` and it worked. So I started looking in the code and found out that anything that I passed to null-ls setup `diagnostic_config` was being ignored after passing through `validate_config` function. Thus `c.get().diagnostic_config` in `diagnostics.lua` is nil.

Here is my minimal init.lua:

```lua
null_ls.setup({
  sources = {
    null_ls.builtins.diagnostics.revive.with({
      args = {
        "-config", vim.fn.stdpath("config") .. "/revive.toml",
        "-formatter", "json",
        "./..."
      },
    }),
  },
  diagnostics_format = "#{m} (#{s})",
  diagnostic_config = {
    virtual_text = false,
  },
})
```

Please let me know if I got something wrong.

mactep:
Done

jose-elias-alvarez:
Great, thanks!

PR #1464, opened 2023-03-23
fix(goimports_reviser): remove the deprecated -file-path arg

kcaashish:
Removed the deprecated -file-path arg and moved $FILENAME to the end as suggested. https://github.com/incu6us/goimports-reviser/commit/979c6e70f1b9dd25a30af74d8ae3efd4205a1bd2

EDIT: based on the discussion #1472, using temp file for now, till the issue with `stdin` gets resolved upstream.


jose-elias-alvarez:
Thanks! Good idea to put the note there.

PR #1465, opened 2023-03-25
feat(builtins/diagnostics/credo): toggle per file/full workspace diagnostics

madlep:
Add `full_workspace` boolean config param to credo.

When `false` (default), credo will run for current file in buffer - this preserves existing behaviour.
When `true`, credo will run globally for all files in the project workspace

Attached is a demo showing with config param disabled (default), then enabled. 
- The first  time nvim is opened, only the credo diagnostics for the selected buffer are generated. When a different buffer is opened, diagnostics for that buffer are additionally generated.
- The config is modified to add `full_workspace = true` for credo. nvim is opened again. When credo runs, project wide diagnostics are generated and populated (instead of just the current buffer).

https://user-images.githubusercontent.com/27223/227716696-686c8bfc-7c4a-4f90-9017-aa6f5b99969d.mov

### DIFF
--- a/lua/null-ls/builtins/diagnostics/credo.lua
+++ b/lua/null-ls/builtins/diagnostics/credo.lua
@@ -19,6 +19,15 @@ return h.make_builtin({
         notes = {
             "Searches upwards from the buffer to the project root and tries to find the first `.credo.exs` file in case the project has nested `credo` configs.",
         },
+        config = {
+            {
+                key = "full_workspace",
+                type = "boolean",
+                description = [[- `false` (default) - run credo for a single file
+- `true` - run credo on the entire workspace. If this is slow on large projects, you may wish to set `method = null_ls.methods.DIAGNOSTICS_ON_SAVE` in `with()` call.]],
+                usage = [[true]],
+            },
+        },
     },
     method = DIAGNOSTICS,
     filetypes = { "elixir" },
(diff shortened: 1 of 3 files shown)


lua/null-ls/builtins/diagnostics/credo.lua
```diff
@@ -34,12 +43,29 @@ return h.make_builtin({
...
 
+            -- `multiple_files = true` must be set ONLY if running
+            -- `full_workspace=true`.
+            -- If it is set when credo is only generating diagnostics per file,
+            -- then existing diagnostics in open buffers will be cleared on
+            -- each subsequent execution in a different buffer
+            if params:get_config().full_workspace then
+                -- this is hacky, but there isn't any way to set `multiple_files`
+                -- dynamically based on user config properly
+                params:get_source().generator.multiple_files = true
```
madlep, line 66: I tried a few different ways to dynamically set `multiple_files` based on user config, but without much luck.

Happy to modify this to a better approach if there's something I'm missing. Mutating the generator during `on_output` doesn't seem great, but couldn't find a better way that didn't involve messing with a bunch of core logic in null-ls.

jose-elias-alvarez:
This approach seems alright, but wouldn't a user be able to achieve the same results by overriding `args` and `multiple_files`? I don't think it's currently possible to do so, but it would be an easy change.

Alternatively: we could just add a separate source (e.g. `credo_project`) that runs on multiple files.

PR #1467, opened 2023-03-26
fix(gomodifytags): detect struct type declaration

ArtAndreev:
Before all type identifiers were detected as struct declarations. This led to the next problems:
* declarations of other types were always struct declarations;
* using types for fields were detected the same way.

### DIFF
--- a/lua/null-ls/builtins/code_actions/gomodifytags.lua
+++ b/lua/null-ls/builtins/code_actions/gomodifytags.lua
@@ -121,6 +121,15 @@ return h.make_builtin({
 
                 -- Ops on struct
                 if (tsnode:type()) == "type_identifier" then
+                    local tspnode = tsnode:parent()
+                    if tspnode == nil or tspnode:type() ~= "type_spec" then
+                        return
+                    end
+                    local typename_node = tspnode:field("type")[1]
+                    if typename_node == nil or typename_node:type() ~= "struct_type" then
+                        return
+                    end
+
                     struct_name = vim.treesitter.query.get_node_text(tsnode, 0)
                     if struct_name == nil then
                         return


jose-elias-alvarez:
Thanks! Could you give me an example of the issue this change solves? I'm a little bit paranoid about having to maintain this source, since I'm not familiar with Go.

ArtAndreev:
Sure! Here is an example of types that causes two different issues:

image

`gomodifytags` works only with struct types, i.e., that are declared with
```go
type name struct{/* fields */}
```

To call code action `gomodifytags` a cursor should be on name of struct, on the screenshot, it's `Strukt`, cursor is on it. In TS terms it's `type_identifier`. The current implementation of the code action checks for it on line 123.

As you see there are also another `type_identifier`s, that appear in snippet:
1. type of field is struct, that is part of another package (line 6 of the snippet). Tag generation failes, because there is no struct `Message` in this file.
2. type of field is struct, that is part of this package (line 7). Tag generation works, but there is no easy way to determine, that this struct lies in this package and use case, when cursor is on type of field to generate tags, is quite nonsense, I think.
3. type of field is not struct (line 14). Tag generation simply fails.

These cases are covered by checking of parent TS node type in fix, it should be `type_spec`.

4. case for `struct_type`, that is not type declaration (`type_spec`).

Covered by the current implementation with check of `type_identifier`.

6. type is derived from not struct, e.g., from int.

This last case is covered by checking of type in fix, it should be sibling of current TS node with `struct_type` type.

I missed one case with generic type:
image

Will fix.

jose-elias-alvarez:
Thanks for the thorough breakdown! Should we wait for #1469, or is this OK to merge?

ArtAndreev:
It's ok to merge, I'll adapt #1469 to changes

ArtAndreev:
Should I rebase the branch?

jose-elias-alvarez:
Excellent, thanks!

PR #1469, opened 2023-03-26
feat(builtins/code_actions): add impl

ArtAndreev:
Add `impl` tool, that generates interface implementation stubs in go language. To use it: 

1. Place your cursor on a name in type declaration:
  ```go
  type |S struct{}
  ```
2. Run code action: `impl: implement interface`;
3. Input interface name in format `package.Interface`, for example, `io.Reader`;
4. Stub will be generated after the type declaration or after the last method of the type.

The code action tool chooses receiver name automatically based on another receiver names of type methods or generates it from the first letter of receiver type lowercased (idiomatic go way). Receiver type is chosen automatically too: pointer or not.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -136,6 +136,25 @@ local sources = { null_ls.builtins.code_actions.gomodifytags }
 
 - Requires installing the Go tree-sitter parser.
 
+### [impl](https://github.com/josharian/impl)
+
+impl generates method stubs for implementing an interface.
+
+#### Usage
+
+```lua
+local sources = { null_ls.builtins.code_actions.impl }
+```
+
+#### Defaults
+
+- Filetypes: `{ "go" }`
+- Method: `code_action`
+
+#### Notes
+
+- Also requires installing the Go tree-sitter parser.
+
 ### [ltrs](https://github.com/jeertmans/languagetool-rust)
 
 LanguageTool-Rust (LTRS) is both an executable and a Rust library that aims to provide correct and safe bindings for the LanguageTool API.
--- a/lua/null-ls/builtins/code_actions/gomodifytags.lua
+++ b/lua/null-ls/builtins/code_actions/gomodifytags.lua
@@ -5,6 +5,11 @@ local u = require("null-ls.utils")
 
 local CODE_ACTION = methods.internal.CODE_ACTION
 
+local treesitter_get_node_text = vim.treesitter.get_node_text
+if not u.has_version("0.9.0") then
+    treesitter_get_node_text = vim.treesitter.query.get_node_text
+end
+
 return h.make_builtin({
     name = "gomodifytags",
     meta = {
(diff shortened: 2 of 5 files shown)


ArtAndreev:
Tested on Neovim v0.8.3 and nightly (v0.9.0-dev-1288+g4863ca6b8)

doc/BUILTINS.md
```diff
@@ -136,6 +136,25 @@ local sources = { null_ls.builtins.code_actions.gomodifytags }
...
+```
+
+#### Defaults
+
+- Filetypes: `{ "go" }`
+- Method: `code_action`
+
+#### Notes
+
+- Requires installing the Go tree-sitter parser.
```
jose-elias-alvarez, line 156:
```suggestion
- Also requires installing the Go tree-sitter parser.
```

Small nit, but this tripped me up.

lua/null-ls/builtins/code_actions/gomodifytags.lua
```diff
@@ -110,18 +111,17 @@ return h.make_builtin({
                 -- End of Execution helpers
 
                 -- Main
-                local tsnode
-                if u.has_version("0.9.0") then
-                    tsnode = vim.treesitter.get_node({ bufnr = bufnr, row = row - 1, col = col - 1 })
-                else
-                    tsnode = vim.treesitter.get_node_at_pos(bufnr, row - 1, col - 1, {})
-                end
+                local tsnode = treesitter.get_node_at_pos(bufnr, row - 1, col - 1)
```
jose-elias-alvarez, line 114: I'm a little confused - wasn't this the issue #1455 was meant to solve? It would also be helpful if we could split up these fixes into another PR. 

Edit: never mind, I understand now, but see my comments on `treesitter.lua`.

lua/null-ls/utils/treesitter.lua
```diff
@@ -0,0 +1,56 @@
+local u = require("null-ls.utils")
+local treesitter = vim.treesitter
+
+local M = {}
+
+-- Returns the smallest named node at the given position
+--
+-- In Neovim v0.9.0 vim.treesitter.get_node_at_pos() is removed.
+-- Use vim.treesitter.get_node() instead.
+M.get_node_at_pos = function(bufnr, row, col)
```
jose-elias-alvarez, line 10: I understand the goal here, but I don't love the idea of adding a compatibility layer. It's not immediately clear to new authors that they should use this vs. just using the utilities directly, and it increases our maintenance burden (I haven't kept up with Neovim development and am not sure why there have been so many changes to the API).

Since it looks like we'll be able to drop this in 0.9, and as far as I know these are the only 2 sources using the API at the moment, I'd prefer to inline these into the sources themselves.

ArtAndreev: Got it, inlining it

jose-elias-alvarez:
I think this needs a rebase after #1467 but otherwise looks good.

jose-elias-alvarez:
Thank you!

ArtAndreev:
You too!

PR #1470, opened 2023-03-27
fix: Use vim.treesitter.get_node_text() to avoid deprecation warning

yutkat:
fix: https://github.com/neovim/neovim/pull/22761


ArtAndreev:
Hello!
Just in case, I fixed it in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1469 with other deprecations

yutkat:
Oh, it was fixed in that Issue. I didn't see it.

PR #1477, opened 2023-03-29
Add sqlfmt formatter to built-ins

ronmorgen1:
This adds the [sqlfmt](https://sqlfmt.com/) formatter as a built-in to null-ls.

### DIFF
--- a/lua/null-ls/builtins/formatting/sqlfmt.lua
+++ b/lua/null-ls/builtins/formatting/sqlfmt.lua
@@ -24,7 +24,7 @@ return h.make_builtin({
         to_stdin = true,
         check_exit_code = { 0, 1 },
         cwd = h.cache.by_bufnr(function(params)
-            return u.root_pattern("dbt_project.yml")(params.bufnr) or vim.fn.getcwd()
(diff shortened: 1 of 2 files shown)


lua/null-ls/builtins/formatting/sqlfmt.lua
```diff
@@ -0,0 +1,31 @@
...
+        "sql",
+        "jinja",
+    },
+    generator_opts = {
+        command = "sqlfmt",
+        args = {},
+        to_stdin = true,
+        check_exit_code = { 0, 1 },
+        cwd = h.cache.by_bufnr(function(params)
+            return u.root_pattern("dbt_project.yml")(params.bufnr) or vim.fn.getcwd()
```
jose-elias-alvarez, line 27:
```suggestion
            return u.root_pattern("dbt_project.yml")(params.bufnr)
```

Can we remove the fallback here? If `root_pattern` returns `nil` then it'll fall back to the configured `cwd`, which may not match.

ronmorgen1: That's a good point :) Adjusted

jose-elias-alvarez:
Thanks!

PR #1491, opened 2023-04-03
Fix spectral cannot be loaded

tiev:
Validation error for spectral diagnostic
image


jose-elias-alvarez:
Thanks! This is something we should have a CI check for.

PR #1495, opened 2023-04-05
feat(rome): support for json

igorlfs:


igorlfs:
> Thanks - could you also update the documentation under meta.notes

Woops, didn't even read the notes! Should be OK now!

PR #1497, opened 2023-04-06
feat: add support for flawfinder

ghost:
I don't know if I've done this right and I don't know lua, but i'll try to do a pull request for flawfinder support anyway.


lua/null-ls/builtins/diagnostics/flawfinder.lua
```diff
@@ -0,0 +1,40 @@
...
+
+return h.make_builtin({
+    name = "flawfinder",
+    meta = {
+        url = "https://dwheeler.com/flawfinder/",
+        description = "Flawfinder, a simple program that examines C/C++ source code and reports possible security weaknesses (“flaws”) sorted by risk level.",
+    },
+    method = methods.internal.DIAGNOSTICS_ON_SAVE,
+    filetypes = { "c", "cpp" },
+    async = true,
```
jose-elias-alvarez, line 12:
```suggestion
```

This is not necessary if you are using `generator_factory`.

lua/null-ls/builtins/diagnostics/flawfinder.lua
```diff
@@ -0,0 +1,40 @@
...
+    to_temp_file = true,
```
jose-elias-alvarez, line 13: This is a generator option, so it needs to go into the table passed into `generator_factory`.

lua/null-ls/builtins/diagnostics/flawfinder.lua
```diff
@@ -0,0 +1,40 @@
...
+        command = "flawfinder",
+        args = {
+            "-S",
+            "-Q",
+            "-D",
+            "-C",
+            "$FILENAME",
+        },
+        to_stdin = false,
+        from_stderr = false,
```
jose-elias-alvarez, line 24:
```suggestion
```

No need to define these explictly, as these are the defaults.

lua/null-ls/builtins/diagnostics/flawfinder.lua
```diff
@@ -0,0 +1,40 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+return h.make_builtin({
+    name = "flawfinder",
+    meta = {
+        url = "https://dwheeler.com/flawfinder/",
+        description = "Flawfinder, a simple program that examines C/C++ source code and reports possible security weaknesses (“flaws”) sorted by risk level.",
+    },
+    method = methods.internal.DIAGNOSTICS_ON_SAVE,
```
jose-elias-alvarez, line 10: Is it necessary to only run this on save? We may be able to use `DIAGNOSTICS` (which updates on change) if we are using a temp file, but we'd need to test it.

jose-elias-alvarez (commented):
Thanks for working on this - just a few minor changes and I think it's all set.

lua/null-ls/builtins/diagnostics/flawfinder.lua
jose-elias-alvarez: Have you tried moving `to_temp_file` into the table passed into `generator_factory`? I believe that should make it work.

Issue #1489, opened 2023-04-02
Adding F# formatter fsprojects/fantomas to Built-in sources


kumkee:
### Issues

- [x] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

I'd love to add [fsprojects/fantomas](https://github.com/fsprojects/fantomas) as an F# formatter to the [Builtins](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md) sources.
Currently, there is nothing in the BUILTINS for the language yet. I am using [ionide/Ionide-vim](https://github.com/ionide/Ionide-vim) as my LSP but would like to have a formatter as well.

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I'm happy to implement this and submit a PR but need someone to point me in the right direction.

The formatter is a command-line tool whose usage can be found [here](https://fsprojects.github.io/fantomas/docs/end-users/GettingStarted.html). For example,
```
dotnet fantomas ./input/array.fs --out ./output/array.fs 
```
I do know I need to follow examples from [this folder](https://github.com/jose-elias-alvarez/null-ls.nvim/tree/main/lua/null-ls/builtins/formatting). 
However, since I am using [NvChad](https://github.com/NvChad/NvChad) with my [config](https://github.com/kumkee/nvchad_config), if I make a fork how do I test if my added source is working?

Thank you

kumkee:
[fantomas](https://github.com/fsprojects/fantomas) has removed both `--stdin` and `--stdout` cli options. 
I've [replied to an old discussion](https://github.com/fsprojects/fantomas/issues/2665#issuecomment-1493410321) there and tried convincing them to bring the options back.

Without those cli options, I don't know how I could write my config file. Is there a work around from here?

jose-elias-alvarez:
There is a [`from_temp_file` option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md#from_temp_file) for formatters that don't support `stdio` that may work here.

kumkee:
The following is my failing attempt https://github.com/jose-elias-alvarez/null-ls.nvim/compare/main...kumkee:null-ls.nvim:main

What else do I need to do?

jose-elias-alvarez:
Closed by #1499.

PR #1499, opened 2023-04-06
feat(builtins/formatting/fantomas): A formatter for F#

kumkee:
### Intention
Mentioned in https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1489 this is to add F# formatter [Fantomas](https://github.com/fsprojects/fantomas).

### Question
Both [filetype_map.lua](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/_meta/filetype_map.lua) and [_meta/formatting.lua](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/_meta/formatting.lua) are marked as _DO NOT EDIT MANUALLY_. I don't see these in other accepted PRs either.
 Then how do I generate the lines in those two files? Without these, the formatter doesn't work.

### DIFF
--- a/lua/null-ls/builtins/_meta/filetype_map.lua
+++ b/lua/null-ls/builtins/_meta/filetype_map.lua
@@ -1,4 +1,3 @@
-local formatting = require "formatting"
 -- THIS FILE IS GENERATED. DO NOT EDIT MANUALLY.
 -- stylua: ignore
 return {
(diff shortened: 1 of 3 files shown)


jose-elias-alvarez:
> Then how do I generate the lines in those two files? Without these, the formatter doesn't work.

The files are updated automatically and committed when a PR is merged. The code looks good - thanks for the contribution!

PR #1500, opened 2023-04-07
fix(builtins/diagnostics/markdownlint): change severity to warning

ArtAndreev:
I think, the error severity is too strict for markdown issues. This PR relaxes it to warning.


jose-elias-alvarez:
Makes sense - the VS Code extension also has these as warnings, so I think we can use that for guidance. Thanks!

Issue #1488, opened 2023-04-01
Custom severity levels for ruff diagnostics


thatlittleboy:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Hi there, I'm using the Ruff diagnostic built-in source.

### Problem

The current implementation for the **severity** is not ideal / too inflexible -- it extracts the "severity" from the first uppercase letter of the code,

https://github.com/thatlittleboy/null-ls.nvim/blob/defa0518762abb32279d601443344fb9773f9330/lua/null-ls/builtins/diagnostics/ruff.lua#L70-L72
```lua
        on_output = h.diagnostics.from_pattern(
            [[(%d+):(%d+): ((%u)%w+) (.*)]],
            { "row", "col", "code", "severity", "message" },
```

which means that Ruff rule codes that share the same initial letter will get lumped together with the same severity (e.g. flake8-comprehensions `C4` v.s. flake8-commas `COM`, flake8-implicit-str-concat `ISC` v.s. flake8-import-conventions `ICN` v.s. isort `I`, and the various pylint categores `PLC`, `PLR`, `PLE`, `PLW`, just to name a few).

Taking the `pylint` example, since they all start with "P", given the current implementation, we can't assign different severities to them, which is a problem.

Furthermore, there's currently no way to customize the severity of the various codes.

### Help

Yes

### Implementation help

I propose to make changes in 2 steps:

1. Make the implementation for mapping code -> severity more flexible, instead of only relying on the first letter. Unfortunately, I don't think it is possible to do so completely within Regex (happy to be proven wrong / hear ideas on this), so I was thinking of maintaining a lua table + writing a for-loop to do the matching, implemented in an adapter. Something like

    ```lua
      local starts_with = function(str, start)
        return str:sub(1, #start) == start
      end

      local default_ruff_rule_severities = {
        COM = h.diagnostics.severities["information"], -- flake8-commas
        PIE = h.diagnostics.severities["warning"], -- flake8-pie
        PLC = h.diagnostics.severities["information"], -- pylint conventions
        PLE = h.diagnostics.severities["error"], -- pylint errors
        PLR = h.diagnostics.severities["hint"], -- pylint refactors
        PLW = h.diagnostics.severities["warning"], -- pylint warnings
        C4 = h.diagnostics.severities["warning"], -- flake8-comprehensions
      }

[...]

PR #1503, opened 2023-04-08
feat(ruff): improve severity mapping

thatlittleboy:
This PR addresses **part 1** of #1488 , which is to use a more flexible approach in assigning severities to code. And addresses the main problem brought up in that issue, which is that purely using a regex approach is insufficient to target specific rules.

Part 2 (exposing an option to configure the severity mapping) to come later.

Please read the review comments for explanation on certain design choices. I'm open to any constructive criticism!
Thanks.

Related to #1488

### DIFF
--- a/lua/null-ls/builtins/diagnostics/ruff.lua
+++ b/lua/null-ls/builtins/diagnostics/ruff.lua
@@ -1,6 +1,75 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
+
+local _starts_with = function(str, start)
+    return str:sub(1, #start) == start
+end
+
+local ruff_rule_severities = {
+    { "ANN", 0, h.diagnostics.severities["information"] }, -- flake8-annotations
+    { "COM", 0, h.diagnostics.severities["information"] }, -- flake8-commas
+    { "ISC", 0, h.diagnostics.severities["error"] }, -- flake8-implicit-str-concat
+    { "PIE", 0, h.diagnostics.severities["warning"] }, -- flake8-pie
+    { "PGH", 0, h.diagnostics.severities["information"] }, -- pygrep-hooks
+    { "PLC", 0, h.diagnostics.severities["information"] }, -- pylint conventions
+    { "PLE", 0, h.diagnostics.severities["error"] }, -- pylint errors
+    { "PLR", 0, h.diagnostics.severities["hint"] }, -- pylint refactors
+    { "PLW", 0, h.diagnostics.severities["warning"] }, -- pylint warnings
+    { "PTH", 0, h.diagnostics.severities["hint"] }, -- flake8-use-pathlib
+    { "RET", 0, h.diagnostics.severities["warning"] }, -- flake8-return
+    { "RUF", 0, h.diagnostics.severities["warning"] }, -- ruff-specific
+    { "SIM", 0, h.diagnostics.severities["information"] }, -- flake8-simplify
+    { "TRY", 0, h.diagnostics.severities["information"] }, -- tryceratops
+    { "T20", 0, h.diagnostics.severities["information"] }, -- flake8-print
+    { "C4", 0, h.diagnostics.severities["warning"] }, -- flake8-comprehensions
+    { "EM", 0, h.diagnostics.severities["information"] }, -- flake8-errmsg
+    { "UP", 0, h.diagnostics.severities["information"] }, -- pyupgrade
+    { "A", 0, h.diagnostics.severities["warning"] }, -- flake8-builtins
+    { "B", 0, h.diagnostics.severities["warning"] }, -- flake8-bugbear
+    { "D", 0, h.diagnostics.severities["information"] }, -- pydocstyle
+    { "E", 0, h.diagnostics.severities["error"] }, -- pycodestyle errors
+    { "F", 0, h.diagnostics.severities["warning"] }, -- pyflakes
+    { "I", 0, h.diagnostics.severities["warning"] }, -- isort
+    { "S", 0, h.diagnostics.severities["warning"] }, -- flake8-bandit
+    { "W", 0, h.diagnostics.severities["warning"] }, -- pycodestyle warnings
+}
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/diagnostics/ruff.lua
```diff
@@ -1,6 +1,75 @@
...
+    { "I", 0, h.diagnostics.severities["warning"] }, -- isort
+    { "S", 0, h.diagnostics.severities["warning"] }, -- flake8-bandit
+    { "W", 0, h.diagnostics.severities["warning"] }, -- pycodestyle warnings
+}
+-- TODO: account for user-provided severities (custom mappings)
+-- local user_severities = { I = "error", B904 = "error" }
+-- for k, sv in pairs(user_severities) do
+--     sv = h.diagnostics.severities[sv]
+--     table.insert(ruff_rule_severities, {k, 1, sv})
+-- end
```
thatlittleboy, line 42: This is my plan for when we extend this functionality to custom user (severity) mappings. We'll need to merge the two mappings together, but taking care to sort the final mapping AFTER, because:

Suppose we have a default mapping for `{"C4": error, "C": warning}`, and the user provides `{"C": hint}` say, we still want to assign error to C4 rules, and hint for all other C rules (whatever they may be).

But more on this in the next PR.

jose-elias-alvarez:
I appreciate the effort but am not sure about this approach - I don't love the idea of iterating over every single possible code-severity mapping for each diagnostic when a simple lookup table seems like it'll do the job.

For example, for the `C4` case: could we not capture the full code initially, check if there is a matching entry, and if not, fall back to the entry corresponding to the initial letter?

```diff
diff --git a/lua/null-ls/builtins/diagnostics/ruff.lua b/lua/null-ls/builtins/diagnostics/ruff.lua
index ce36801..2796f1c 100644
--- a/lua/null-ls/builtins/diagnostics/ruff.lua
+++ b/lua/null-ls/builtins/diagnostics/ruff.lua
@@ -1,7 +1,23 @@
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
-local custom_end_col = {
+
+local ruff_rule_severities = {
+    ["C4"] = h.diagnostics.severities["warning"],
+    ["C"] = h.diagnostics.severities["error"],
+}
+
+local ruff_adapters = {
+    severity = function(entries)
+        if ruff_rule_severities[entries.code] then
+            return ruff_rule_severities[entries.code]
+        end
+
+        local first_letters = entries.code:match("%u+") -- there's surely a better name for this
+        if ruff_rule_severities[first_letters] then
+            return ruff_rule_severities[first_letters]
+        end
+    end,
     end_col = function(entries, line)
         if not line then
             return
@@ -67,27 +83,11 @@ return h.make_builtin({
         end,
         to_stdin = true,
         ignore_stderr = true,
-        on_output = h.diagnostics.from_pattern(
-            [[(%d+):(%d+): ((%u)%w+) (.*)]],
-            { "row", "col", "code", "severity", "message" },
-            {
-                adapters = {
-                    custom_end_col,
-                },
-                severities = {
-                    E = h.diagnostics.severities["error"], -- pycodestyle errors
-                    W = h.diagnostics.severities["warning"], -- pycodestyle warnings
-                    F = h.diagnostics.severities["information"], -- pyflakes
-                    A = h.diagnostics.severities["information"], -- flake8-builtins
-                    B = h.diagnostics.severities["warning"], -- flake8-bugbear
-                    C = h.diagnostics.severities["warning"], -- flake8-comprehensions
-                    T = h.diagnostics.severities["information"], -- flake8-print
-                    U = h.diagnostics.severities["information"], -- pyupgrade
-                    D = h.diagnostics.severities["information"], -- pydocstyle
-                    M = h.diagnostics.severities["information"], -- Meta
-                },
-            }
-        ),
+        on_output = h.diagnostics.from_pattern([[(%d+):(%d+): (%w+) (.*)]], { "row", "col", "code", "message" }, {
+            adapters = {
+                ruff_adapters,
+            },
+        }),
     },
     factory = h.generator_factory,
 })
```

As for the user customization aspect: I don't see how `diagnostics_postprocess` doesn't already handle this case:

```lua
local my_custom_severities = {
    ["C4"] = vim.diagnostic.severity["ERROR"],
}

local ruff = null_ls.builtins.diagnostics.ruff.with({
    diagnostics_postprocess = function(diagnostic)
        local custom_severity = my_custom_severities[diagnostic.code]
        if custom_severity then
            diagnostic.severity = custom_severity
        end
    end,
})
```

jose-elias-alvarez:
I also wanted to mention: since there is an LSP solution in active development, I don't want to spend a ton of effort building out a source here, since that should become the preferred option as it matures (which is what's happened with the ESLint language server, for example).

thatlittleboy:
Thanks. Regarding your suggested approach, to be clear, the "code" that is being captured is the full code, like ("F401" or "C400", "C901", not "C4" or "C90"; u can see the code list [here](https://beta.ruff.rs/docs/rules/)). As such,  I was considering a more generic solution in which it would be possible, which means this doesn't quite work

```diff
+        if ruff_rule_severities[entries.code] then
+            return ruff_rule_severities[entries.code]
+        end
```

unless you specify the full code-severity mapping. (Which is why I went for the `startswith` approach). I could alleviate the iteration problem by only iterating those that are ambiguous, but I suspect that code might be even harder to maintain.

Just to hammer the point home, there's also a scenario I'm imagining, that the user might only want to make the "B9xx" rules a different severity from the rest of the "B" rules (NOTE: "B9xx" is still flake8-bugbear == "B" rule group, but such B9 rules are under the experimental flag).

---

> I also wanted to mention: since there is an LSP solution in active development, I don't want to spend a ton of effort building out a source here, since that should become the preferred option as it matures (which is what's happened with the ESLint language server, for example).

Fair enough. Should I close this PR (no hard feelings)? I'll just develop it as a custom source in my own nvim config, then.

jose-elias-alvarez:
I see, thanks for clarifying the issues with my approach. I am hesitant to include this to avoid an ESLint-like situation where we put a lot of effort into a source that becomes obsolete, so I think keeping this as a personal modification might be the best approach.

Issue #1466, opened 2023-03-26
Beautify json before writing to config in cSpell


keaising:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

[In existing code](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/code_actions/cspell.lua#L140), a raw json string is written to cspell.json without beautifying it.

Is there any way to beautify a json string (e.g. sort elements, keys in array and object, add proper indentation) before writing it to cspell.json?

In my research, I found that lua doesn't have a `JSON.stringify` function like js, so I think there are two ways to implement this feature:
1. embed another 3rd party library for beautifying
2. null-ls provide a callback after write to cspell.json, user can call some tools like `jq` to beautify config by theirselves

I have already read https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1412#issuecomment-1454859416 and I just want to know this posibility in null-ls.

### Help

Yes

keaising:
I have a try and it works as my wish. Here is what I have done:

1. Add a `postprocess` function after `vim.fn.writefile`: https://github.com/keaising/null-ls.nvim/commit/ef40cd2a304da245d698b5c09a6b7e4d4ad35b26
2. In my `null-ls` config, call `postprocess` function and format the config file with `jq`: [my null-ls config](https://github.com/keaising/dotfile/commit/c6f4b28ca7febba14fd307725f381a9c482edd14#diff-6859dc28992f2f89ef71e0980ec34d0670eeacbe1e863ffca84579d00eb49dc6R40-R43)

Would you consider merging this `postprocess` function into the main branch? I can make a merge request.

jose-elias-alvarez:
I am not a fan of adding this as a specific option for the `cspell` source, which in my opinion has already grown too large - see my comments in #1412. I'd be fine with adding a general hook for all sources that accepts `params` as an argument, which would make it possible to do this.

keaising:
Yeah, I agree with you, I think `postprocess` is a hook that meet your requirement: it can be called after the code action has been processed, all sources can benefit from this hook.

[...]

PR #1510, opened 2023-04-11
feat(builtins/code_actions/cspell): add `on_success` hook to cspell (#1466)

keaising:
Discussed in https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1466

With this hook, users can do whatever they like with cspell_config_file, such as beautifying the config file.

```lua
    config = {
        find_json = function(_)
            return vim.fn.expand("~/.config/nvim/cspell.json")
        end,
        on_success = function(cspell_config_file)
            os.execute(
                string.format(
                    "cat %s | jq -S '.words |= sort' | tee %s > /dev/null",
                    cspell_config_file,
                    cspell_config_file
                )
            )
        end,
    },
}),

```

https://github.com/keaising/dotfile/blob/553612b8c5b8456e10bfcf588a7f1c4787f2adb8/nvim/.config/nvim/lua/plugins/lsp.lua#L108-L116

### DIFF
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -58,6 +58,35 @@ return h.make_builtin({
             "This source depends on the `cspell` built-in diagnostics source, so make sure to register it, too.",
         },
         usage = "local sources = { null_ls.builtins.diagnostics.cspell, null_ls.builtins.code_actions.cspell }",
+        config = {
+            {
+                key = "find_json",
+                type = "function",
+                description = "Customizing the location of cspell config",
+                usage = [[
+find_json = function(cwd)
+    return vim.fn.expand(cwd .. "/cspell.json")
+end
+]],
+            },
+            {
+                key = "on_success",
+                type = "function",
+                description = "Callback after successful execution of code action.",
+                usage = [[
+function(cspell_config_file, params)
+    -- format the cspell config file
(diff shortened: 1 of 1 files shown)


lua/null-ls/builtins/code_actions/cspell.lua
```diff
@@ -148,6 +148,11 @@ return h.make_builtin({
                             diagnostic.end_col,
                             { word }
                         )
+
+                        local on_success = config.on_success
+                        if on_success then
+                            on_success(cspell_json_file)
```
jose-elias-alvarez, line 154: I think it might also be helpful for this to take `params` as a 2nd argument. That way users have access to the original file's path, `cwd`, and other useful info.

keaising:
Ok, I have add `params` as the second parameter and rebased my branch onto the `main` branch

jose-elias-alvarez:
Thanks - last thing we need before merging this is to [document the new option](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONTRIBUTING.md#configuration) (sorry I didn't mention this earlier).

keaising:
I have documented the two config functions: `find_config` and `on_success`, please review the documentation.

jose-elias-alvarez:
I pushed a commit removing extra blank lines and fixing the syntax for the `find_json` documentation. Other than that, this looks good. Thanks for working on this!

PR #1512, opened 2023-04-12
Add treefmt formatter to builtins

lbodor:
https://numtide.github.io/treefmt/

### DIFF
--- a/lua/null-ls/builtins/formatting/treefmt.lua
+++ b/lua/null-ls/builtins/formatting/treefmt.lua
@@ -1,0 +1,29 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "treefmt",
+    meta = {
+        url = "https://github.com/numtide/treefmt",
+        description = "One CLI to format your repo",
+        usage = [[
+local sources = {
+    null_ls.builtins.formatting.treefmt.with({
+        -- treefmt requires a config file
+        condition = function(utils)
+            return utils.root_has_file("treefmt.toml")
+        end,
+    }),
+}]],
+    },
+    method = FORMATTING,
+    filetypes = {},
+    generator_opts = {
+        command = "treefmt",
+        args = { "--allow-missing-formatter", "--stdin", "$FILENAME" },
+        to_stdin = true,
+    },
+    factory = h.formatter_factory,
+})


lbodor:
Here is what I use to disable formatters that are already handled by treefmt:

```lua
-- Find project root.
local project_root = function()
    -- See airblade/vim-rooter for FindRootDiretory.
    -- vim-rooter updates cwd via autocmd, but not before null-ls receives diagnostics on buffer load.
    return vim.api.nvim_call_function("FindRootDirectory", {}) or vim.loop.cwd()
end

-- Try to find treefmt.toml.
local treefmt_config = function()
    local config = project_root() .. "/treefmt.toml"
    if vim.fn.filereadable(config) == 1 then
        return config
    end
end

-- Return true if formatter is handled by treefmt.
local is_registered_with_treefmt = function(formatter_name)
    local config = treefmt_config()
    if config then
        for line in io.lines(config) do
            if line == "[formatter." .. formatter_name .. "]" then
                return true
            end
        end
    end
end

-- Runtime condition to disable formatters that are handled by treefmt.
local unless_registered_with_treefmt = function(params)
    return not is_registered_with_treefmt(params:get_source().name)
end

local sources = {
    ...
    null_ls.builtins.formatting.treefmt,
    null_ls.builtins.diagnostics.selene.with({
        diagnostics_format = "#{m} (#{s})",
    }),
    null_ls.builtins.formatting.stylua.with({
        extra_args = { "--indent-type", "Spaces" },
    }),
    ...
}

-- Return true if source is a formatter
local is_formatter = function(source)
    return (type(source.method) == "table" and lib.contains(source.method, null_ls.methods.FORMATTING))
        or (source.method == null_ls.methods.FORMATTING)
end

for _, source in ipairs(sources) do
    -- null-ls computes lsp client root once on startup, so we dynamically change the
    -- spawn directory to project root of current file, so that null-ls sources can find
    -- their configuration files (e.g., selene.toml, treefmt.toml).
    source.generator.opts.cwd = project_root

    -- Disable formatters that are handled by treefmt.
    if is_formatter(source) then
        source.generator.opts.runtime_condition = unless_registered_with_treefmt
    end
end

null_ls.setup({ sources = sources })
```

lua/null-ls/builtins/formatting/treefmt.lua
```diff
@@ -0,0 +1,23 @@
...
+return h.make_builtin({
+    name = "treefmt",
+    meta = {
+        url = "https://github.com/numtide/treefmt",
+        description = "One CLI to format your repo",
+    },
+    method = FORMATTING,
+    filetypes = {},
+    condition = function(utils)
+        return utils.root_has_file({ "treefmt.toml" })
```
jose-elias-alvarez, line 15: Does the formatter not run at all if this file is not present? Either way I would leave this out - null-ls root dir detection is not especially robust, and we haven't included anything like this as a default for other sources. If you think it's something users would want, we can mention it in the documentation instead.

jose-elias-alvarez, line 15: Is this necessary for the formatter to run at all? I would leave it out - null-ls root dir detection is flawed, and we haven't included anything like this for other sources. If you think this is something users would want, we can mention it in the documentation instead.

lbodor: No, it fails without a config file.

```
[lbodor@io:~/.config]$ treefmt
[ERR]: treefmt.toml could not be found in /home/lbodor/.config and up. Use the --init option to create one or specify --config-file if it is in a non-standard location.

[lbodor@io:~/.config]$ cd nvim
[lbodor@io:~/.config/nvim]$ treefmt

0 files changed in 13ms (found 31, matched 16, cache misses 0)

[lbodor@io:~/.config/nvim]$ cat treefmt.toml
[formatter.alejandra]
command = "/nix/store/18hrifk50zzl1rr3918ll2c2m8za8n12-alejandra-3.0.0/bin/alejandra"
excludes = []
includes = ["*.nix"]
options = []

[formatter.stylua]
command = "/nix/store/89fzy6znpgxzl3cika4smg8n0w3syywb-stylua-0.17.0/bin/stylua"
excludes = []
includes = ["*.lua"]
options = ["--indent-type", "Spaces"]
[lbodor@io:~/.config/nvim]$
```

I use it with a `runtime_condition` and `cwd` as above, so that `vim.lsp.buf.format` uses `treefmt` if a project has a `treefmt.toml` file and otherwise applies formatters as configured via `null_ls.builtins.formatting`.

jose-elias-alvarez: For consistency I think I'd still prefer to leave `condition` blank and use the documentation to clarify that a config file is required. We've had lots of issues that ultimately come down to "null-ls didn't correctly set my root dir" so I'd prefer to leave it up to users.

lbodor: Fair enough. I moved the condition to `meta.usage`.

jose-elias-alvarez:
Thanks!

PR #1513, opened 2023-04-12
Enable additional filetypes for Terraform files

goetzc:
The formatter [lua/null-ls/builtins/formatting/terraform_fmt.lua](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/f8ffcd7cb8fb3325c711d459152ef132b5b65aed/lua/null-ls/builtins/formatting/terraform_fmt.lua#L13) already had these extra filetypes set.


jose-elias-alvarez:
Thanks!

PR #1517, opened 2023-04-14
feat: add ts-node-action as a code_action builtin

axieax:
This PR adds ts-node-action as a builtin for code actions.


jose-elias-alvarez:
Seems to work well, thanks for the contribution! It might be worth putting up a PR to update their documentation.

Issue #1373, opened 2023-01-28
per-buffer setup


MurdeRM3L0DY:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Provide a function(eg: `start_or_attach`) to start null-ls server on a per-buffer basis such that it can be used in `ftplugin/*`

e.g
`ftplugin/lua.lua`
```lua
local null = require 'null-ls'
null.start_or_attach { sources = { ... } }
```

I haven't looked too deeply into the codebase but I imagine this function could be reused in the `setup` function

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I would need help with the api design

jose-elias-alvarez:
Attaching to the client or starting it if needed can be done using [`try_add`](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/7b2b28e207a1df4ebb13c7dc0bd83f69b5403d71/lua/null-ls/client.lua#L107) (which isn't part of the public API but could be). 

However, null-ls config is either global or source-specific, and sources themselves are filetype-specific, so I think proper per-buffer setup would require a larger rewrite.

Issue #1415, opened 2023-02-23
CSpell ignores custom dictionary `addWord` logic.


dzintars:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

As per [cspell documentation](https://cspell.org/docs/dictionaries-custom/) user can create custom dictionaries in the external files.
Option `addWord` allows to add words to that external file.
Current logic of `cspell` Builtin code action ignores that and writes all words to the cspell configuration file.
I am not aware what happens if some user (by some reason) creates several external dictionaries and each has `addWord` option.

### Help

No

jose-elias-alvarez:
I mentioned this in #1412, but the same comment applies here:

> The `cspell` built-in sources are some of the most complex we have, and there's still more features that are missing - see #1415 for an example. Honestly I think the best approach would be to create a companion plugin that uses the null-ls API, similar to [prettier.nvim](https://github.com/MunifTanjim/prettier.nvim).

davidmh:
Hey @dzintars I had the same request, and after chatting with @jose-elias-alvarez we decided to split off the CSpell sources into a companion plugin, checkout [cspell.nvim](https://github.com/davidmh/cspell.nvim) which adds support custom dictionaries.

PR #1518, opened 2023-04-14
feat(builtins/code_actions): cspell dictionaries

davidmh:
Hi @jose-elias-alvarez,

I've been playing with the configuration of CSpell for one of the projects at my job, and I noticed that the code actions for CSpell could use a custom dictionary integration.

Custom dictionaries are local files referenced in the CSpell config file, see: https://cspell.org/docs/dictionaries-custom/

There's one things I would like to add, but I'm not sure what's the best approach:

- [ ] It would be great if we could display the list of custom dictionaries without having to drill drown into a second select.

  But reading and parsing the config file in the generator's callback function seems to slow down the warm up significantly.

  One idea I have is to defer the read/parse of the config, keep the results in a local cache, and invalidate that cache on the refresh mentioned in the previous point.
 
  Another idea is to expose a 'on_attach' callback at the source level, so we can run initializations there, which sounds close to #1373
  
What do you think?

### DIFF
--- a/lua/null-ls/builtins/code_actions/cspell.lua
+++ b/lua/null-ls/builtins/code_actions/cspell.lua
@@ -60,6 +60,17 @@ local get_word = function(diagnostic)
     )[1]
 end
 
+local set_word = function(diagnostic, word)
+    vim.api.nvim_buf_set_text(
+        diagnostic.bufnr,
+        diagnostic.lnum,
+        diagnostic.col,
+        diagnostic.end_lnum,
+        diagnostic.end_col,
+        { word }
+    )
+end
+
 return h.make_builtin({
     name = "cspell",
     meta = {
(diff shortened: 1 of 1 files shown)


davidmh:
cc: @yoo, @marianozunino and @PumpedSardines since you've all worked on this code action

marianozunino:
Funny, I was thinking about doing the small change of `Add <word> to cspell`!

PumpedSardines:
@davidmh I made a very tiny change to the cspell code action, so i don't have a lot of input here!

davidmh:
Updated the PR to a functional state, ready for review 😃

marianozunino:
May I suggest a small change? Can you alter the title in order to also display the word?

> title = "Add '" .. get_word(diagnostic) .. "' to cspell json file"

I know is not the scope of the PR, but I find that in some cases I have more than 1 misspelling in the same line and I'm not sure which one I'm adding to the dic.

davidmh:
@marianozunino done

jose-elias-alvarez:
Apologies - I'm hesitant to continue adding to the `cspell` source for reasons I've mentioned in other PRs:

> The `cspell` built-in sources are some of the most complex we have, and there's still more features that are missing - see #1415 for an example. Honestly I think the best approach would be to create a companion plugin that uses the null-ls API, similar to [prettier.nvim](https://github.com/MunifTanjim/prettier.nvim).

I'm happy to provide support in creating a companion plugin, but the `cspell` source is already difficult to maintain (especially since I don't use it).

davidmh:
That's fair, I can take a look at the companion plugin approach.

davidmh:
Continuing this conversation in #1534

PR #1524, opened 2023-04-18
help selene find its config

bukzor-sentryio:
selene does no search for selene.toml and only finds it if it's in the $PWD

I tested interactively, and with this change selene is always properly configured.


bukzor-sentryio:
Please review.
<3

jose-elias-alvarez:
LGTM, thanks!

PR #1525, opened 2023-04-19
feat: run tests without the pack system

davidmh:
When a Neovim configuration doesn't use packer or other pack-related ways of managing plugins, like lazy.nvim, the tests fail to run.

To work around that, this commit updates the minimal config to ensure plenary is always available, regardless of the chosen plugin manager.

I adapted the configuration from https://github.com/akinsho/toggleterm.nvim/commit/50f58a0624c92d55201398ae50c85857bf53e744


jose-elias-alvarez:
This seems fine, especially if toggleterm is already using it. My only concern is that it seems easy for users to end up with stale versions of plenary.nvim. Maybe we could add a new make target that deletes `.tests` and add instructions to the test documentation?

davidmh:
Done, feel free to change anything.

jose-elias-alvarez:
Thanks!

PR #1526, opened 2023-04-19
fix(builtins/diagnostics/clazy): Handle fatal errors

syyyr:
clazy can sometimes output fatal errors, for example when a file doesn't exist:
```
main.cpp:1:10: fatal error: 'non-existent' file not found
#include <non-existent>
         ^~~~~~~~~~~~~~
1 error generated.
```

These weren't caught by the regex and that resulted in 'msg' being nil.

Fix this by allowing a space when parsing the severity.


jose-elias-alvarez:
Thanks!

PR #1528, opened 2023-04-20
update readme to include conditional function

cloudre01:
Would like to add to **MAIN.md** readmefile, a new function in that was added in Dec 2022 based on the pull request to help others, so that they don't have to dig through past issues to find out.


jose-elias-alvarez:
Thanks!

Issue #1527, opened 2023-04-19
null_ls.builtins.diagnostics.todo_comments breaks typescriptreact syntax highlighting in new buffers


thallada:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

v0.9.0

### Dev Version?

- [X] I am using a stable Neovim release version, or if I am using a dev version of Neovim I have confirmed that my issue is reproducible on a stable version.

### Operating System

MacOS (but also reproducible on WSL Linux)

### Minimal Config

```Lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
    local null_ls = require("null-ls")
    -- add only what you need to reproduce your issue
    null_ls.setup({
        sources = {
            null_ls.builtins.diagnostics.todo_comments,
        },
    })
end

local function load_plugins()
    -- only add other plugins if they are necessary to reproduce the issue
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {
                'nvim-treesitter/nvim-treesitter',
                run = function() require('nvim-treesitter.install').update({ with_sync = true }) end,
                config = function()
                    require('nvim-treesitter.configs').setup {
                        ensure_installed = { "typescript", "tsx" },
                        auto_install = true,
                        highlight = {
                            enable = true,
                        }
                    }
                end,
            },
            {
                "jose-elias-alvarez/null-ls.nvim",
                requires = { "nvim-lua/plenary.nvim" },
[...]

PR #1533, opened 2023-04-22
fix(builtins/diagnostics/todo_comments): convert filetype to lang before calling vim.treesitter.get_parser

thallada:
Fixes issue #1527. Same reproduction steps with this patch does not produce the syntax bug.

I also created a separate bug with nvim-treesitter about the syntax breaking when `get_parser` is called with an invalid lang: https://github.com/nvim-treesitter/nvim-treesitter/issues/4693

### DIFF
--- a/lua/null-ls/builtins/diagnostics/todo_comments.lua
+++ b/lua/null-ls/builtins/diagnostics/todo_comments.lua
@@ -1,4 +1,5 @@
 local h = require("null-ls.helpers")
+local log = require("null-ls.logger")
 local methods = require("null-ls.methods")
 
 local DIAGNOSTICS = methods.internal.DIAGNOSTICS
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Thanks! Re: logging the error: the source is enabled for all filetypes by default, but most filetypes don't have Treesitter parsers (nor is the user guaranteed to have them all installed), so I'd support logging it at the `debug` or `trace` level so that it only shows up in logs. That would make it possible to debug issues like this without overloading users with notifications.

Also, I think we can get rid of [this line](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b862270965aa616df4e78b01477e6007a6be4b72/lua/null-ls/builtins/diagnostics/todo_comments.lua#L93), since it should be handled by the correspondence table.

thallada:
@jose-elias-alvarez I added a debug log to the `get_parser` error case and removed the latex ft conversion line you mentioned (I also confirmed that `vim.treesitter.language.get_lang('tex')` does return 'latex').

jose-elias-alvarez:
LGTM, thanks for putting this together!

PR #1542, opened 2023-04-28
docs: fix from_temp_file option for generator_factory helper

hituzi-no-sippo:
When reading the following source code, it assigns to `params.output`.

https://github.com/jose-elias-alvarez/null-ls.nvim/blob/33b853a3933eed3137cd055aac4e539e69832ad0/lua/null-ls/helpers/generator_factory.lua#L304-L306


jose-elias-alvarez:
Thanks for catching this!

PR #1557, opened 2023-05-12
fix(builtins): add golangci_lint support for monorepos

arnevm123:
Hi,
I had issues trying to get golangci-lint to work in a monorepo.
I believe this shoud fix the issue. While trying to figure this out myself I also found #1206 which helped me fix the issue, but seems stale.
Please let me know if anything needs to change.
I've also added a space to the string formatting as this looks more in line with for example eslint_d output


jose-elias-alvarez:
LGTM, thanks for picking this up!

PR #1560, opened 2023-05-14
fix(builtins):Goimports-reviser can not accept extra_args at the end of the command

arnevm123:
Hi,

I have found out that goimports-reviser does not accept arguments at the end of the command.
Example:
goimports-reviser main.go -company-prefixes k8s.io --> Just the normal import sorting
goimports-reviser -company-prefixes k8s.io main.go --> Takes into account the -company-prefixes

This is the sollution I came up with, while not impacting other builtins, but it should be usable for others if the same problem arrises.

All feedback is welcome 😄 I am not too familiar with the project so I might have missed stuff.


jose-elias-alvarez:
Thanks, this seems like a good solution. A few requests before merging this:

1. Can we rename the new option to something like `prepend_extra_args`?
2. Can we add documentation [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/HELPERS.md)?
3. Can we add test coverage [here](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/test/spec/helpers/make_builtin_spec.lua)? It should be pretty straightforward to copy-paste and modify one of the existing tests.

arnevm123:
Hi,
I've fixed your comments I think. I also noticed it would break the "-" arg.
Let me know if anything else needs to change!
Thanks for this project and the feedback 😄

lua/null-ls/helpers/make_builtin.lua
```diff
@@ -65,17 +66,24 @@ local function make_builtin(opts)
 
     if opts.extra_args then
         local original_args, original_extra_args = generator_opts.args, opts.extra_args
+        local original_prepend_extra = generator_opts.prepend_extra_args
```
jose-elias-alvarez, line 69: I don't think there's a need for this naming, since this will always be a boolean. We also don't need to make a copy.

jose-elias-alvarez (commented):
Sorry for the delayed review, one small change and I think this is good to go.

lua/null-ls/helpers/make_builtin.lua
arnevm123: I think everything should be fine now :smile:

jose-elias-alvarez:
Thanks for working on this!

PR #1561, opened 2023-05-14
Switch sqlfmt to using stdin instead of file

JulesNP:
When using vim-dadbod-ui to write SQL queries, it creates the query files with no extension, but sets the SQL filetype for the buffer. These files currently will not format using sqlfmt, as sqlfmt ignores non-SQL file extensions when given a filename as input.

This patch switches the sqlfmt built-in formatter to using stdin instead of passing the file, which allows these files to be formatted correctly.


jose-elias-alvarez:
Thanks!

PR #1568, opened 2023-05-20
feat: add pyink range formatting

kuator:
`pyink` supports range formatting through the `--pyink-lines` option. It works if one decides to format a piece of code in visual mode with `vim.lsp.format()`. But for some reason `gqq` doesn't work. I checked and null-ls does set `formatexpr` as `vim.lsp.formatexpr()`. Did I miss something?

### DIFF
--- a/lua/null-ls/builtins/formatting/pyink.lua
+++ b/lua/null-ls/builtins/formatting/pyink.lua
@@ -2,6 +2,7 @@ local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
 
 local FORMATTING = methods.internal.FORMATTING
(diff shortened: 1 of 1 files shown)


jose-elias-alvarez:
Not sure about the `gqq` issue, do other range formatting sources work?

kuator:
> Not sure about the `gqq` issue, do other range formatting sources work?
Stylua works
Peek 2023-05-29 22-21

jose-elias-alvarez:
Pretty weird, might be something in how the CLI works since range formatting isn't really standardized. I'm happy to merge this as-is once CI passes if you're good with it.

jose-elias-alvarez:
Thanks!

PR #1571, opened 2023-05-21
WIP: feat: add forge fmt formatter for solidity files

ashishbinu:


ashishbinu:
This works well for files without syntax error. For files with syntax error `cat main.sol | forge fmt --raw -` outputs the syntax error, but for null_ls to work perfectly it shouldn't output the file. I am not sure how to fix that issue.

PR #1575, opened 2023-05-23
fix(builtins)!: rename swift-format to swift_format

Almaz5200:
Swift has both "swiftformat" and "swift-format" formatter's which got mixed up in builtins and got registered under the same name. This PR adresses that issue


jose-elias-alvarez:
Thanks, I think this closes #1551 too.

Almaz5200:
Sure! Done

jose-elias-alvarez:
Thanks!

PR #1578, opened 2023-05-28
feat: add forge_fmt formatter for solidity

ashishbinu:


jose-elias-alvarez:
Were you able to solve the issue in #1571?

ashishbinu:
Yes, it works now.

Issue #1577, opened 2023-05-28
Incorrect diagnostics positions


name-snrl:
### FAQ

- [X] I have checked the [FAQ](https://github.com/jose-elias-alvarez/null-ls.nvim#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues?q=is%3Aissue) and there are no issues with the same problem.

### Neovim Version

NVIM v0.9.0

### Dev Version?

- [X] I am using a stable Neovim release version, or if I am using a dev version of Neovim I have confirmed that my issue is reproducible on a stable version.

### Operating System

NixOS rev: 963006aab35e3e8ebbf6052b6bf4ea712fdd3c28

### Minimal Config

```Lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
  local path_sep = on_windows and "\\" or "/"
  local result = table.concat({ ... }, path_sep)
  return result
end

vim.g.loaded_remote_plugins = ""
vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir = vim.loop.os_getenv("TEMP") or "/tmp"

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local null_ls_config = function()
  local null_ls = require("null-ls")
  -- add only what you need to reproduce your issue
  null_ls.setup({
    sources = {
      null_ls.builtins.diagnostics.shellcheck,
    },
    debug = true,
  })
end

local function load_plugins()
  -- only add other plugins if they are necessary to reproduce the issue
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      {
        "jose-elias-alvarez/null-ls.nvim",
        requires = { "nvim-lua/plenary.nvim" },
        config = null_ls_config,
      },
      --{
      --  'neovim/nvim-lspconfig',
      --  config = function() require 'lspconfig'.bashls.setup { } end,
      --},
    },
    config = {
      package_root = package_root,
      compile_path = compile_path,
    },
  })
end

if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #1579, opened 2023-05-28
diagnostics: convert columns to bytes

name-snrl:
Explanation https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1577.

Works in my case, but it was done in the morning hastily, but I think it is a good start


jose-elias-alvarez:
Seems like a reasonable change, but the failing tests indicate that this might need more attention. Can you check them out?

name-snrl:
> Can you check them out?

I'm sorry, I don't know when I'll have time for this. Also, I think the code needs to be refined. In general, my code is a highly stripped-down function, which I mentioned in the issue, but I may have cut out important things.

name-snrl:
@jose-elias-alvarez This is finally solution. I'm not familiar with the tests, could you please resolve this part?

PR #1589, opened 2023-06-07
feat(builtins): added nelua

LucasTavaresA:
im currently selecting only `error` cause nelua -a prints AST nodes as well which makes no sense in this context
nelua --lint does not get "normal" errors, eg: array size inference stuff

i need a way to remove `from` from the capture or to capture only the necessary words, skimming through the source code `error`, `warning` and `info` seem to be the ones 

but since im not sure @edubart can you help here?

also nelua -a is a thing that get executed on the file one time so does not look like an lsp so i figured this is the place this goes and not lspconfig

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -1351,6 +1351,23 @@ local sources = { null_ls.builtins.diagnostics.mypy }
 - Command: `mypy`
 - Args: dynamically resolved (see [source](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/mypy.lua))
 
+### [nelua](https://github.com/edubart/nelua-lang)
+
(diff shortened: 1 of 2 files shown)


LucasTavaresA:
there is information on the source code about errors and they are error|warning|info so i matched them and removed AST nodes information

LucasTavaresA:
i just found out about -L flag, im using it with `$DIRPATH` to find nelua modules

LucasTavaresA:
ok sry for the pushfest, nelua is just not good at all at finding lib paths, tried everything but in any project that has a different structure or when you are not on its path i keep finding edge cases, its pattern matching with '?/?.nelua' does not help either

so $DIRNAME is the best i can do at least to catch when you are not in the project folder

LucasTavaresA:
im using this for a while and looks like nelua-lsp does more without ast information on the way, so im closing this for now

PR #1600, opened 2023-06-23
feat: add swipl

tani:
### [swipl](https://www.swi-prolog.org/)

SWI-Prolog is a versatile implementation of the Prolog language.

#### Usage

```lua
local sources = { null_ls.builtins.diagnostics.swipl }
```

#### Defaults

- Filetypes: `{ "prolog" }`
- Method: `diagnostics`
- Command: `swipl`
- Args: `{ "-q", "-t", "halt(1)", "-s", "$FILENAME" }`

### DIFF
--- a/lua/null-ls/builtins/diagnostics/swipl.lua
+++ b/lua/null-ls/builtins/diagnostics/swipl.lua
@@ -3,87 +3,87 @@ local methods = require("null-ls.methods")
 local DIAGNOSTICS_ON_SAVE = methods.internal.DIAGNOSTICS_ON_SAVE
 
 return h.make_builtin({
-  method = DIAGNOSTICS_ON_SAVE,
-  filetypes = { "prolog" },
-  generator_opts = {
-    command = "swipl",
-    args = { "-q", "-t", "halt(1)", "-s", "$FILENAME" },
-    format = "raw",
-    check_exit_code = function(code)
-      return code <= 1
-    end,
-    on_output = function(params, done)
-      local diagnostics = {}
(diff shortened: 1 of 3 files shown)


lua/null-ls/builtins/diagnostics/swipl.lua
```diff
@@ -0,0 +1,82 @@
...
+    generator_opts = {
+        command = "swipl",
+        args = { "-q", "-t", "halt(1)", "-s", "$FILENAME" },
+        format = "raw",
+        check_exit_code = function(code)
+            return code <= 1
+        end,
+        on_output = function(params, done)
+            local diagnostics = {}
+            local lines = vim.fn.split(params.err, "\r\n\\|\r\\|\n")
```
jose-elias-alvarez, line 17: Small improvement: we could consider using [vim.split](https://github.com/neovim/neovim/blob/7968322e7a20b557631f4b496751658e80f6e7b0/runtime/lua/vim/shared.lua#L176) here instead, which would simplify your check for emptiness a few lines down.

jose-elias-alvarez (commented):
Looks fine to me, but based on what I'm seeing, you are basically iterating over each line and matching against patterns, right? If so you may want to consider using the [diagnostics.from_patterns helper](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/aac27a1fa550de3d0b2c651168167cc0d5366a9a/lua/null-ls/builtins/diagnostics/mypy.lua#L48), which could simplify your code a bunch and make future maintenance easier.

PR #1603, opened 2023-06-24
fix(builtins/formatting/black): set the cwd according to the config files

Nelyah:
This is exactly the same change as in https://github.com/jose-elias-alvarez/null-ls.nvim/pull/1449

The black builtin formatter would not find the `pyproject.toml` in my case if the `pyproject.toml` file was not at the repo root. This fixes the issue. 

In this folder architecture, the `black` formatter would correctly respect the settings in the TOML file.
```
my_repo/
    |_ pyproject.toml
    |_ python_package/
        |_ README.md
        |_ src/
        |_ setup.py
```

---

In this folder architecture, the `black` formatter would **NOT** correctly respect the settings in the TOML file.
```
my_repo/
    |_ python_package/
        |_ README.md
        |_ src/
        |_ setup.py
        |_ pyproject.toml
```

Both cases work with my changes.

I have run `FILE=test/spec/builtins/formatting_spec.lua make test-file` successfully.


test/spec/builtins/formatting_spec.lua
```diff
@@ -0,0 +1,33 @@
+local mock = require("luassert.mock")
```
jose-elias-alvarez, line 1: Thanks for adding this, but I don't think there's a need to specifically test this type of behavior for built-ins, since the helpers are separately tested. Can you go ahead and remove this?

Nelyah: Sounds good, I've removed the test file. 

Although, I feel that since I somewhat changed _a_ behaviour, there should be a way to test it.

Nelyah:
I've pushed a `fixup!` commit, but if you intend to merge, LMK and I'll squash the two of them.

jose-elias-alvarez:
Thanks!

Issue #1239, opened 2022-11-14
add blackd support


miki725:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

just realized `black` comes with server variant:

https://black.readthedocs.io/en/stable/usage_and_configuration/black_as_a_server.html

not sure what is exact performance benefit but might be useful to use that instead of black in some-cases

### Help

Yes

### Implementation help

will attempt to send PR although not Im not very well versed how to test things locally if I have custom branch of null-ls but presume packer documents it somewhere. filing issue in case other people would be interested for tracking

jose-elias-alvarez:
Based on a quick read of the documentation, Plenary's [curl module](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/curl.lua) may come in handy here, since we already depend on it. To run a local branch, you can just clone the repo and point Plenary at a path on your local filesystem. Happy to review a PR whenever!

PR #1606, opened 2023-06-26
feat: add blackd support

tiagovla:
#1239

### DIFF
--- a/lua/null-ls/builtins/formatting/blackd.lua
+++ b/lua/null-ls/builtins/formatting/blackd.lua
@@ -1,6 +1,7 @@
+local curl = require("plenary.curl")
 local h = require("null-ls.helpers")
 local methods = require("null-ls.methods")
-local curl = require("plenary.curl")
+local u = require("null-ls.utils")
 
 local FORMATTING = methods.internal.FORMATTING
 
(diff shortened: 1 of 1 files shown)


tiagovla:
I'm not sure how to change the parameters, something like this
```lua
local null = require("null-ls")
null.setup {
    sources = {
        null.builtins.formatting.blackd.with {
            extra_args = function(params)
                params.options.line_length = 42 
            end,
        },
    }
}
```
does not work.

tiagovla:
```lua
local null = require "null-ls"
null.setup {
    sources = {
        null.builtins.formatting.blackd.with {
            args = {
                line_length = 42,
                fast = true,
            },
        },
    },
}
```
This works.

lua/null-ls/builtins/formatting/blackd.lua
```diff
@@ -0,0 +1,38 @@
...
+    },
+    method = FORMATTING,
+    filetypes = { "python" },
+    factory = function(opts)
+        return {
+            fn = function(params)
+                local options = u.handle_function_opt(opts.args)
+                local hostname = options.hostname or "localhost"
+                local port = options.port or 45484
+                local response = curl.post(hostname .. ":" .. port, {
```
jose-elias-alvarez, line 22: I haven't worked with it recently, but does Plenary's `curl` module have an async option? If so we could try using that and convert this into an [async generator](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/b919452c84e461c21a79185bef90c96e1cfecff9/doc/MAIN.md?plain=1#L262).

lua/null-ls/builtins/formatting/blackd.lua
```diff
@@ -0,0 +1,38 @@
...
+    meta = {
+        url = "https://github.com/psf/black",
+        description = "The uncompromising Python code formatter",
+    },
+    method = FORMATTING,
+    filetypes = { "python" },
+    factory = function(opts)
+        return {
+            fn = function(params)
+                local options = u.handle_function_opt(opts.args)
```
jose-elias-alvarez, line 19: Strictly speaking, `args` should be used for the arguments actually passed to a command line process. In this case I think it would be more idiomatic to use [the `config` key](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONTRIBUTING.md#configuration).

lua/null-ls/builtins/formatting/blackd.lua
```diff
@@ -0,0 +1,38 @@
...
+local methods = require("null-ls.methods")
+local u = require("null-ls.utils")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "blackd",
+    meta = {
+        url = "https://github.com/psf/black",
+        description = "The uncompromising Python code formatter",
```
jose-elias-alvarez, line 12: Does this need to be updated? Ideally this would clarify the difference between `black` and `blackd` and also mention how to configure the source.

Also, will `blackd` automatically pick up config files?

lua/null-ls/builtins/formatting/blackd.lua
tiagovla: I just found the callback option.

lua/null-ls/builtins/formatting/blackd.lua
tiagovla: Oh, that's a thing!

lua/null-ls/builtins/formatting/blackd.lua
tiagovla: I added the meta info, not sure if the lines can be that long.

It can't pick up config files at all, the only possible configuration are the options I listed. It is quite limited.

lua/null-ls/builtins/formatting/blackd.lua
```diff
@@ -0,0 +1,84 @@
+local curl = require("plenary.curl")
```
jose-elias-alvarez, line 1: Can you move this require statement into the generator callback? At the moment it's breaking the documentation autogen script, since the script doesn't have access to Plenary.

lua/null-ls/builtins/formatting/blackd.lua
```diff
@@ -0,0 +1,84 @@
...
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
+    name = "blackd",
+    meta = {
+        url = "https://github.com/psf/black",
+        description = "blackd is a small HTTP server that exposes Black’s functionality over a simple protocol. The main benefit of using it is to avoid the cost of starting up a new Black process every time you want to blacken a file.",
```
jose-elias-alvarez, line 11: Can we also mention that the *only* way to configure the formatter is by using the provided config options, i.e. that it won't pick up on config files at all? If you'd like to split info across lines, you can use the [`meta.notes` key](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/CONTRIBUTING.md#contributing-built-ins).

tiagovla: Done.

lua/null-ls/builtins/formatting/blackd.lua
tiagovla: Done.

jose-elias-alvarez:
Thank you!

PR #1509, opened 2023-04-11
 feat(builtins): add textidote

igorlfs:
Hello there,

I'm trying to close #1312, but I'm having some issues. 

First things first: let's define a sample file to understand textidote's output:

```tex
\documentclass{article}

\begin{document}

This documnet contians erors.

\end{document}
```

If I run `textidote --quiet --no-color --check en --output singleline sample.tex`, this gets printed to **stdout**:

```
sample.tex(L5C6-L5C14): Possible spelling mistake found.. Suggestions: [document] (6) "This documne
t contians erors."
sample.tex(L5C15-L5C23): Possible spelling mistake found.. Suggestions: [contains] (15) "This docum
net contians erors."
sample.tex(L5C24-L5C29): Possible spelling mistake found.. Suggestions: [errors, Eros, errs] (24) "
This documnet contians erors."
```

I then defined the following pattern: `[[(.*)(L(%d+)C(%d+)-L(%d+)C(%d+)): (.*)]]`. However, I'm getting the following warning:

```
failed to run generator: null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:219: error in generator output: 
/path/to/sample.tex(L5C6-L5C14): Possible spelling mistake found.. Suggestions: [document] (6) "This documnet contians erors."
/path/to/sample.tex(L5C15-L5C23): Possible spelling mistake found.. Suggestions: [contains] (15) "This documnet contians erors."
/path/to/sample.tex(L5C24-L5C29): Possible spelling mistake found.. Suggestions: [errors, Eros, errs] (24) "This documnet contians erors."
```

I tried enabling debug mode (here's the [log](https://gist.github.com/igorlfs/7abedb32ec3b54a3051aa6102fb8b080)), which suggests the output is going to _stderr_ instead:

```
[TRACE Tue 11 Apr 2023 11:15:43 AM -03] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:204: **error output**: /path/to/sample.tex(L5C6-L5C14): Possible spelling mistake found.. Suggestions: [document] (6) "This documnet contians erors."
```

However, if I try to enable `from_stderr` (currently not pushed), while I don't get any warnings, I don't get any diagnostics either. Here's a [log](https://gist.github.com/igorlfs/77dc7388bc7993238b20fe094f97e0fb) with `from_stderr` enabled. What's going on? Do I just have a faulty pattern, or am I missing something else?

[...]

Issue #1312, opened 2022-12-27
add textidote builtin 


igorlfs:
### Issues

- [X] I have checked [existing issues](https://github.com/jose-elias-alvarez/null-ls.nvim/issues) and there are no existing ones with the same request.

### Feature description

Textidote is a grammar checker that works with latex and markdown. It uses LanguageTool under the hood, which means it supports a wide variety of languages.

### Help

No

PR #1616, opened 2023-07-05
feat: add textidote support

tiagovla:
Closes #1509 and #1312.


tiagovla:
@jose-elias-alvarez this might be really slow depending on the tex file. Should I increase the timeout?

jose-elias-alvarez:
> @jose-elias-alvarez this might be really slow depending on the tex file. Should I increase the timeout?

Sure, we can increase the timeout by default since the source only runs on save.

jose-elias-alvarez:
Thanks for picking this up!

PR #1619, opened 2023-07-08
feat(builtin-formatter): python-docformatter

0xblur:
`docformatter` is a PEP-257 complaint Python autoformatter. It is compatible with autopep8 and black. It works on docstrings and no other formatter has this capability. It is also highly customizable (see https://docformatter.readthedocs.io/en/latest/usage.html#use-from-the-command-line for additional arguments).

### DIFF
--- a/lua/null-ls/builtins/formatting/docformatter.lua
+++ b/lua/null-ls/builtins/formatting/docformatter.lua
@@ -1,0 +1,22 @@
+local h = require("null-ls.helpers")
+local methods = require("null-ls.methods")
+
+local FORMATTING = methods.internal.FORMATTING
+
+return h.make_builtin({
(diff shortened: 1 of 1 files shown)


mskelton:
I'm taking a stab at forking null-ls given the recent announcement that null-ls is being archived, and I've implemented this in https://github.com/mskelton/null-ls.nvim/commit/835ea4d5

0xblur:
> I'm taking a stab at forking null-ls given the recent announcement that null-ls is being archived, and I've implemented this in [mskelton@835ea4d5](https://github.com/mskelton/null-ls.nvim/commit/835ea4d5)

Damn... That's sad. Thanks for updating this on your fork, nonetheless.

PR #1622, opened 2023-07-11
feat: add builtin isortd formatter

disrupted:
based on blackd #1606


mskelton:
I'm taking a stab at forking null-ls given the recent announcement that null-ls is being archived, and I've implemented this in https://github.com/mskelton/null-ls.nvim/commit/37a37b4f.

disrupted:
@mskelton that's great to hear. will be switching to your fork then.

PR #1623, opened 2023-07-11
feat(formatting):adding range_formatting for google-java-format builtin

deanbear:
Hi there,

This PR is about adding the range_formatting method to the built-in source of google-java-format.
In fact, google-java-format already supports the --lines option for range formatting.
However, there's a trick: you need to add four additional options to properly perform the range formatting operation; otherwise, the entire file will be formatted.
Please see the issue at https://github.com/google/google-java-format/issues/942 for more details.

I chose not to use the helpers because there are extra options involved, and I didn't want to disrupt the range_formatting_args_factory.

Have a nice day.

### DIFF
--- a/doc/BUILTINS.md
+++ b/doc/BUILTINS.md
@@ -3483,15 +3483,19 @@ Reformats Java source code according to Google Java Style.
 #### Usage
 
 ```lua
-local sources = { null_ls.builtins.formatting.google_java_format }
+local sources = {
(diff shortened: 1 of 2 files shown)


mskelton:
I'm taking a stab at forking null-ls given the recent announcement that null-ls is being archived, and I've implemented this in https://github.com/mskelton/null-ls.nvim/commit/ea1a4539

